PR #6, opened 2023-05-22
Generate `protocolUpdateUTxOCostPerByte` in `genProtocolParametersUpdate` era dependently (typesafe version)

newhoggy:
# Description

This version is more type safe, but being type safe is not free.  In the type-safe version any type used in the `a` position of `Feature` needs to be in scope which restricts where `Feature` can be defined.  This means `Feature` must import other modules.

Additionally, we require an untyped version of the function to implement the type safe versions.

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [ ] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/cardano-api/internal/Cardano/Api/Features.hs
+++ b/cardano-api/internal/Cardano/Api/Features.hs
@@ -1,0 +1,54 @@
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE StandaloneDeriving #-}
+
+module Cardano.Api.Features
+  ( Feature(..)
+  , SupportedInEra(..)
+  , supportedInEra
+  ) where
+
+import           Cardano.Api.Eras
+import           Data.Aeson (ToJSON (..))
+
+data ProtocolParameterUTxOCostPerWord
+
+data ProtocolParameterUTxOCostPerByte
+
+-- | A representation of a feature that is supported in a given era.
+data Feature f where
+  ProtocolParameterUTxOCostPerWord :: Feature ProtocolParameterUTxOCostPerWord
+  ProtocolParameterUTxOCostPerByte :: Feature ProtocolParameterUTxOCostPerByte
+
+deriving instance Eq (Feature f)
+deriving instance Show (Feature f)
+
+instance ToJSON (Feature f) where
+  toJSON = toJSON . show
+
+-- | A representation of a feature whether a feature is supported in a given era.
+data SupportedInEra f era where
+  ProtocolParameterUTxOCostPerWordSupportedInAlonzoEra  :: SupportedInEra ProtocolParameterUTxOCostPerWord AlonzoEra
+
+  ProtocolParameterUTxOCostPerByteSupportedInBabbageEra :: SupportedInEra ProtocolParameterUTxOCostPerByte BabbageEra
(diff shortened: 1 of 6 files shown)


cardano-api/internal/Cardano/Api/Features.hs
```diff
@@ -0,0 +1,103 @@
...
+deriving instance Eq (Feature f a)
+deriving instance Show (Feature f a)
+
+instance ToJSON (Feature f a) where
+  toJSON = toJSON . show
+
+-- | Determine whether a feature is supported in a given era.
+--
+-- If the feature is not supported in the given era, 'no' is returned, otherwise 'yes'.
+untypedSupportedInEra
```
carbolymer, line 35: The order of arguments is a bit unintuitive for me. I think it would be better to follow the order of `if` or `if`-like functions with the checks at the beginning of the arguments:
````haskell
if precondition then yes else no
````
so in this case:
```haskell
  :: Feature feature b
  -- ^ The feature to check.
  -> CardanoEra era
  -- ^ The era to check
  -> a
  -- ^ The value to return if the feature is not supported in the given era.
  -> a
  -- ^ The value to return if the feature is supported in the given era.
  -> a
````

File: cardano-api/internal/Cardano/Api/Features.hs [hunk 2]
```diff
@@ -0,0 +1,103 @@
...
+import           Cardano.Api.Eras
+import           Cardano.Api.Value
+import           Data.Aeson (ToJSON (..))
+
+data ProtocolParameterUTxOCostPerWord
+
+data ProtocolParameterUTxOCostPerByte
+
+-- | A representation of a feature that is supported in a given era.
+data Feature feature a where
```
carbolymer, line 22: Why do we need GADT here? Phantom type parameter would be enough in my opinion.

cardano-api/internal/Cardano/Api/Features.hs
```diff
@@ -0,0 +1,103 @@
...
+  , whenMaybeSupportedInEra
+  ) where
+
+import           Cardano.Api.Eras
+import           Cardano.Api.Value
+import           Data.Aeson (ToJSON (..))
+
+data ProtocolParameterUTxOCostPerWord
+
+data ProtocolParameterUTxOCostPerByte
```
carbolymer, line 19: Maybe `DataKinds`  would be useful to have a single type for features?

cardano-api/internal/Cardano/Api/Features.hs
```diff
@@ -0,0 +1,103 @@
...
+
+-- | A representation of a feature that is supported in a given era.
+data Feature feature a where
+  ProtocolParameterUTxOCostPerWord :: Feature ProtocolParameterUTxOCostPerWord Lovelace
```
carbolymer, line 23: Why do we need type-level and value level features? This seems redundant for a simple check.

carbolymer (changes requested):
https://github.com/input-output-hk/cardano-api/pull/6/files#r1200353959
https://github.com/input-output-hk/cardano-api/pull/6/files#r1200344326

PR #7, opened 2023-05-22
Import groups in stylish haskell

newhoggy:
# Description

This means we no longer need to manually enforce import groups manually and the relevant sections can be removed from the style guide.

Also split out OS specific code into separate modules

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [ ] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/.stylish-haskell.yaml
+++ b/.stylish-haskell.yaml
@@ -9,8 +9,9 @@
 # 5. Redunant imports and pragmas should be removed
 # 6. Consistent syntax
 # 7. No trailing whitespaces
-# 8. Slightly generous screen with assumed
+# 8. Slightly generous screen width assumed
 # 9. All Haskell files in the project are subject to code formatting style
+# 10. Import grouping is handled by stylish-haskell
 
 steps:
   # Principle 4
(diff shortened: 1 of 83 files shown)


cardano-api/cardano-api.cabal
```diff
@@ -59,6 +59,9 @@ library internal
                         Cardano.Api.HasTypeProxy
                         Cardano.Api.InMode
                         Cardano.Api.IO
+                        Cardano.Api.IO.Compat
+                        Cardano.Api.IO.Compat.Posix
+                        Cardano.Api.IO.Compat.Win32
```
newhoggy, line 64: `stylish-haskell` import grouping does not work well in modules that have `CPP`.  New modules have been added to accomodate Win32-only and Posix-only code respectively.  This minimises the amount of code where CPP is used with lots of imports to avoid triggering `stylish-haskell`'s deficiency in this area.

erikd (approved):

.stylish-haskell.yaml
```diff
@@ -41,6 +42,18 @@ steps:
...
 
+      # Principle 10
+      group_imports: true
+
+      group_rules:
+        - match: ^Cardano.Api\>
+        - match: ^(Cardano|Ouroboros|PlutusCore|PlutusLedgerApi)\>
+        - match: ^Prelude\>
+        - match: ^(Control|Codec|Data|Formatting|GHC|Lens|Network|Numeric|Options|Prettyprinter|System|Text)\>
+        - match: ^(Test.Gen.Cardano.Api)\>
```
carbolymer, line 53: Why do we separate `Test.Gen.Cardano.Api` from `Test.Gen.Cardano`?

carbolymer: Nevermind, it's a local package. Makes sense.

carbolymer (approved):
Nice! I think we should add those changes to `cardano-node` as well.  It would be nice if we could sync those files between repos somehow (submodules?, additional CI checks?)

Jimbo4350 (approved):

PR #23, opened 2023-05-25
update chaps index

JaredCorduan:
# Description

update chaps index:

* increase bounds on ledger core
* increase bounds on plutus
* support Plutus V3 in conway ledger era

# TODO

- [ ] There is a broken golden test, `Bech32DecodingError`
- [ ] There is a TODO to decide if we should change the signature of `mkVersionedProtocols` pass the new query parameter to `NodeToClientVersionData`.

# Checklist

- [ ] Commit sequence broadly makes sense and commits have useful messages
- [ ] New tests are added if needed and existing tests are updated.  These may include:
  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
  See [Runnings tests](https://github.com/input-output-hk/cardano-node-wiki/wiki/Running-tests) for more details
- [ ] Any changes are noted in the `CHANGELOG.md` for affected package
- [ ] The version bounds in `.cabal` files are updated
- [ ] CI passes. See note on CI.  The following CI checks are required:
  - [ ] Code is linted with `hlint`.  See `.github/workflows/check-hlint.yml` to get the `hlint` version
  - [ ] Code is formatted with `stylish-haskell`.  See `.github/workflows/stylish-haskell.yml` to get the `stylish-haskell` version
  - [ ] Code builds on Linux, MacOS and Windows for `ghc-8.10.7` and `ghc-9.2.7`
- [ ] Self-reviewed the diff

# Note on CI
If your PR is from a fork, the necessary CI jobs won't trigger automatically for security reasons.
You will need to get someone with write privileges.  Please contact IOG node developers to do this
for you.

PR #24, opened 2023-05-25
update chaps index

JaredCorduan:
# Description

update chaps index:

* increase bounds on ledger core
* increase bounds on plutus
* support Plutus V3 in conway ledger era
* add `query` field to `LocalNodeConnectInfo` record. (However, I was reluctant to break the API functions `foldBlocks` and `queryStateForBalancedTx` without knowing what I am doing, so I hardcoded the value `True` for them).

(replaces #23, which was from a fork)

# Changelog

```yaml
- description: |
    - Updated plutus, ledger, and consensus dependency bounds
    - added support for Plutus V3 in conway
    - added a `query` field to `LocalNodeConnectInfo`
  # no-changes: the API has not changed
  # compatible: the API has changed but is non-breaking
  # breaking: the API has changed in a breaking way
  compatibility: breaking (field added to `LocalNodeConnectInfo`)
  # feature: the change implements a new feature in the API
  # bugfix: the change fixes a bug in the API
  # test: the change fixes modifies tests
  # maintenance: the change involves something other than the API
  type: maintenance
```
# TODO

- [ ] Is it okay that I have hard-coded a value of `True` for the query field used by the two public API functions `queryStateForBalancedTx` and `foldBlocks`?

# Checklist

- [x] New tests are added if needed and existing tests are updated.  These may include:
  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [x] Any changes are noted in the `CHANGELOG.md` for affected package
- [x] The version bounds in `.cabal` files are updated
- [x] CI passes. See note on CI.  The following CI checks are required:
  - [x] Code is linted with `hlint`.  See `.github/workflows/check-hlint.yml` to get the `hlint` version
  - [x] Code is formatted with `stylish-haskell`.  See `.github/workflows/stylish-haskell.yml` to get the `stylish-haskell` version
  - [x] Code builds on Linux, MacOS and Windows for `ghc-8.10.7` and `ghc-9.2.7`
- [x] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -14,7 +14,7 @@ repository cardano-haskell-packages
 -- you need to run if you change them
 index-state:
   , hackage.haskell.org 2023-05-10T10:34:57Z
-  , cardano-haskell-packages 2023-05-13T07:08:44Z
+  , cardano-haskell-packages 2023-05-24T10:41:02Z
 
 packages:
     cardano-api
--- a/cardano-api/internal/Cardano/Api/IPC.hs
+++ b/cardano-api/internal/Cardano/Api/IPC.hs
@@ -265,7 +265,8 @@ mkVersionedProtocols networkid ptcl unversionedClients =
           Net.versionedNodeToClientProtocols
             ptclVersion
             NodeToClientVersionData {
-              networkMagic = toNetworkMagic networkid
+              networkMagic = toNetworkMagic networkid,
+              query = False
             }
             (\_connid _ctl -> protocols (unversionedClients ptclVersion) ptclBlockVersion ptclVersion))
       (Map.toList (Consensus.supportedNodeToClientVersions proxy))
--- a/cardano-api/internal/Cardano/Api/Script.hs
+++ b/cardano-api/internal/Cardano/Api/Script.hs
@@ -100,6 +100,7 @@ module Cardano.Api.Script (
     toAlonzoLanguage,
     fromAlonzoLanguage,
     fromShelleyScriptToReferenceScript,
+    scriptInEraToRefScript,
 
     -- * Data family instances
     AsType(..),
@@ -192,6 +193,7 @@ data SimpleScript'
 --
 data PlutusScriptV1
 data PlutusScriptV2
+data PlutusScriptV3
 
 instance HasTypeProxy SimpleScript' where
     data AsType SimpleScript' = AsSimpleScript
@@ -205,6 +207,10 @@ instance HasTypeProxy PlutusScriptV1 where
 instance HasTypeProxy PlutusScriptV2 where
     data AsType PlutusScriptV2 = AsPlutusScriptV2
     proxyToAsType _ = AsPlutusScriptV2
+
+instance HasTypeProxy PlutusScriptV3 where
+    data AsType PlutusScriptV3 = AsPlutusScriptV3
+    proxyToAsType _ = AsPlutusScriptV3
 
 -- ----------------------------------------------------------------------------
 -- Value level representation for script languages
@@ -230,6 +236,7 @@ instance TestEquality ScriptLanguage where
 data PlutusScriptVersion lang where
     PlutusScriptV1 :: PlutusScriptVersion PlutusScriptV1
     PlutusScriptV2 :: PlutusScriptVersion PlutusScriptV2
+    PlutusScriptV3 :: PlutusScriptVersion PlutusScriptV3
 
 deriving instance (Eq   (PlutusScriptVersion lang))
 deriving instance (Show (PlutusScriptVersion lang))
@@ -237,6 +244,7 @@ deriving instance (Show (PlutusScriptVersion lang))
 instance TestEquality PlutusScriptVersion where
     testEquality PlutusScriptV1 PlutusScriptV1 = Just Refl
     testEquality PlutusScriptV2 PlutusScriptV2 = Just Refl
+    testEquality PlutusScriptV3 PlutusScriptV3 = Just Refl
     testEquality _ _ = Nothing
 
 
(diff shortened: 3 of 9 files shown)


cardano-api/internal/Cardano/Api/Script.hs
```diff
@@ -100,6 +100,7 @@ module Cardano.Api.Script (
     toAlonzoLanguage,
     fromAlonzoLanguage,
     fromShelleyScriptToReferenceScript,
+    scriptInEraToRefScript,
```
JaredCorduan, line 103: Is it okay that I have exported this? I needed it for the tests.

cardano-api/internal/Cardano/Api/IPC.hs
```diff
@@ -265,7 +265,8 @@ mkVersionedProtocols networkid ptcl unversionedClients =
           Net.versionedNodeToClientProtocols
             ptclVersion
             NodeToClientVersionData {
-              networkMagic = toNetworkMagic networkid
+              networkMagic = toNetworkMagic networkid,
+              query = True --TODO add this as a parameter to mkVersionedProtocols?
```
JaredCorduan, line 269: @jprider63 - I've just hard-coded the query flag to True, which I assume is a bad idea. :) Do we just want to pass the flag to `mkVersionedProtocols`?

cardano-api/src/Cardano/Api/Shelley.hs
```diff
@@ -145,6 +145,7 @@ module Cardano.Api.Shelley
     ReferenceTxInsScriptsInlineDatumsSupportedInEra(..),
     refInsScriptsAndInlineDatsSupportedInEra,
     refScriptToShelleyScript,
+    scriptInEraToRefScript,
```
newhoggy, line 148: This is an addition to the API right?

cardano-api/internal/Cardano/Api/Script.hs
newhoggy: This is an internal module.  I think it's fine.  You can use internal modules from tests.

cardano-api/src/Cardano/Api/Shelley.hs
newhoggy: As you can use internal modules from tests this might not be necessary.

JaredCorduan: yes. I asked [above](https://github.com/input-output-hk/cardano-api/pull/24#discussion_r1204900322) if this was okay. I think I needed it...

JaredCorduan: ah! sorry for the misunderstanding!

cardano-api/internal/Cardano/Api/IPC.hs
JaredCorduan: I've now instead added the `query` field to `LocalNodeConnectInfo` record. However, I was reluctant to break the API functions `foldBlocks` and `queryStateForBalancedTx` without knowing what I am doing, so I hardcoded the value True for them. Please let me know if there is something better to do here.

cabal.project
```diff
@@ -14,7 +14,7 @@ repository cardano-haskell-packages
 -- you need to run if you change them
 index-state:
   , hackage.haskell.org 2023-05-10T10:34:57Z
-  , cardano-haskell-packages 2023-05-13T07:08:44Z
+  , cardano-haskell-packages 2023-05-24T10:41:02Z
```
carbolymer, line 17: I think `flake.lock` needs refreshing as well.
````
nix flake lock --update-input CHaP
````

File: cardano-api/internal/Cardano/Api/Convenience/Query.hs [hunk 5]
```diff
@@ -143,6 +144,7 @@ executeQueryCardanoMode socketPath era nid q = runExceptT $ do
           { localConsensusModeParams = CardanoModeParams (EpochSlots 21600)
           , localNodeNetworkId = nid
           , localNodeSocketPath = socketPath
+          , localNodeQuery = True -- make configurable?
```
Jimbo4350: This is fine because the connection info will only be used with queries.

cardano-api/internal/Cardano/Api/LedgerState.hs
```diff
@@ -411,7 +411,8 @@ foldBlocks nodeConfigFilePath socketPath validationMode state0 accumulate = do
           LocalNodeConnectInfo {
             localConsensusModeParams = cardanoModeParams,
             localNodeNetworkId       = networkId',
-            localNodeSocketPath      = socketPath
+            localNodeSocketPath      = socketPath,
+            localNodeQuery           = True -- TODO should this be configurable?
```
Jimbo4350: This should be false because the connection info will be used with the chain sync protocol: https://github.com/input-output-hk/cardano-api/blob/bb44f10e84b30991ba184d0e65b400a03363c614/cardano-api/internal/Cardano/Api/LedgerState.hs#L429

cardano-api/internal/Cardano/Api/Script.hs
```diff
@@ -306,13 +318,16 @@ instance ToJSON AnyPlutusScriptVersion where
...
+      Aeson.String "PlutusScriptV3"
 
 parsePlutusScriptVersion :: Text -> Aeson.Parser AnyPlutusScriptVersion
 parsePlutusScriptVersion t =
   case t of
     "PlutusScriptV1" -> return (AnyPlutusScriptVersion PlutusScriptV1)
     "PlutusScriptV2" -> return (AnyPlutusScriptVersion PlutusScriptV2)
-    _                -> fail "Expected PlutusScriptV1 or PlutusScriptV2"
+    "PlutusScriptV3" -> return (AnyPlutusScriptVersion PlutusScriptV3)
+    _                -> fail "Expected PlutusScriptVX, for X = 1, 2, or 3"
```
Jimbo4350, line 330: Such a mathematician :laughing:

Jimbo4350 (commented):
One small change

cardano-api/internal/Cardano/Api/LedgerState.hs
JaredCorduan: you are absolutely right! In fact, @jprider63 just told me the same thing.

Jimbo4350 (approved):

cardano-api/internal/Cardano/Api/Script.hs
JaredCorduan: at least I didn't leave it to the reader to determine the correct values of X :laughing: .

disassembler (approved):

cardano-api/internal/Cardano/Api/IPC.hs
JaredCorduan: I've now just hard coded `False` in `mkVersionedProtocols`.

PR #34, opened 2023-06-02
Use tag script from `cardano-dev` repository instead

newhoggy:
# Description

The `tag.sh` script is re-usable between projects so it has been moved to a separate repository.

# Changelog

```yaml
- description: |
    Use tag script from `cardano-dev` repository instead
  compatibility: no-api-changes
  type: maintenance
```

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [ ] Self-reviewed the diff

# Note on CI
for you.


cardano-dev.yaml
```diff
@@ -0,0 +1,5 @@
+version: 
+main-branch: main
+release-branch:
+  prefix: release/
+  suffix: -branch
```
carbolymer, line 5: why do we need this suffix?

carbolymer (approved):

cardano-dev.yaml
newhoggy: It's to indicate that the version number is incomplete.

For example if the tag's name is `cardano-api-1.2.3.4`, we may make any of the following tags depending which number in the version we wish maintain:

* `cardano-api-1.2.3-branch`: Next version will be `cardano-api-1.2.3.5`
* `cardano-api-1.2-branch`: Next version will be `cardano-api-1.2.4`
* `cardano-api-1-branch`: Next version will be `cardano-api-1.3`

We can use some other convention.  For example `.x` in which case it becomes:

* `cardano-api-1.2.3.x`: Next version will be `cardano-api-1.2.3.5`
* `cardano-api-1.2.x`: Next version will be `cardano-api-1.2.4`
* `cardano-api-1.x`: Next version will be `cardano-api-1.3`

The point of the PR is to make it configurable.

PR #40, opened 2023-06-05
Generate fields only if they make sense for the given era

newhoggy:
# Description

Introduce the `FeatureInEra` type class.  This type class will be used by "feature" types to define a way to determine if that feature is supported in an era.

Introduce the `FeatureValue` type.  This type is meant to be used to define data types with era-sensitive type-safe fields.

This PR does not _yet_ use `FeatureValue` to implement type-safe fields for the `ProtocolParameter` type because doing so requires that `ProtocolParameter` take an `era` type argument, and there are challenges to doing that as this time.

The changes to introduce type-safe fields will be introduced in https://github.com/input-output-hk/cardano-api/pull/39 when some other issues have been resolved.

More explanation is made inline.

# Changelog

```yaml
- description: |
    Generate fields only if they make sense for the given era.
    Changes:
    - New `Cardano.Api.Feature` module
    - New`FeatureInEra` type class
    - New `FeatureValue` type
    - New functions:
      - `genFeatureValueInEra`
      - `featureInShelleyBasedEra`
      - `isFeatureValue`
      - `valueOrDefault`
      - `asFeatureValue`
      - `asFeatureValueInShelleyBasedEra`
    - `genProtocolParameters` and `genValidProtocolParameters` functions take additional `era` argument
  compatibility: breaking
  type: feature
```

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [ ] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/cardano-api/gen/Test/Gen/Cardano/Api/Typed.hs
+++ b/cardano-api/gen/Test/Gen/Cardano/Api/Typed.hs
@@ -7,7 +7,9 @@
 {-# LANGUAGE TypeApplications #-}
 
 module Test.Gen.Cardano.Api.Typed
-  ( genAddressByron
+  ( genFeatureValueInEra
+
+  , genAddressByron
   , genAddressInEra
   , genAddressShelley
   , genCertificate
@@ -119,6 +121,7 @@ import           Cardano.Api hiding (txIns)
 import qualified Cardano.Api as Api
 import           Cardano.Api.Byron (KeyWitness (ByronKeyWitness),
                    WitnessNetworkIdOrByronAddress (..))
+import           Cardano.Api.Feature
 import           Cardano.Api.Script (scriptInEraToRefScript)
 import           Cardano.Api.Shelley (GovernancePoll (..), GovernancePollAnswer (..), Hash (..),
                    KESPeriod (KESPeriod),
@@ -137,7 +140,7 @@ import qualified Cardano.Ledger.Alonzo.Scripts as Alonzo
 import           Cardano.Ledger.SafeHash (unsafeMakeSafeHash)
 import qualified Cardano.Ledger.Shelley.TxBody as Ledger (EraIndependentTxBody)
 
-import           Control.Applicative (optional)
+import           Control.Applicative (Alternative (..), optional)
 import           Data.ByteString (ByteString)
 import qualified Data.ByteString as BS
 import qualified Data.ByteString.Short as SBS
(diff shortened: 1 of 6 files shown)


cardano-api/internal/Cardano/Api/ProtocolParameters.hs
```diff
@@ -699,6 +700,27 @@ instance FeatureInEra ProtocolUTxOCostPerByteFeature where
     BabbageEra  -> yes ProtocolUTxOCostPerByteInBabbageEra
     ConwayEra   -> yes ProtocolUTxOCostPerByteInConwayEra
 
+-- | A representation of whether the era supports the 'UTxO Cost Per Word'
+-- protocol parameter.
+--
+-- The Babbage and subsequent eras support such a protocol parameter.
+--
+data ProtocolUTxOCostPerWordFeature era where
+  ProtocolUpdateUTxOCostPerWordInAlonzoEra :: ProtocolUTxOCostPerWordFeature AlonzoEra
```
newhoggy, line 709: This is how to define a "feature".  This is exactly the way we currently define features except the naming convention for them is "<FeatureName>SupportedInEra".  This PR proposes to use the word `Feature` in place of `SupportedInEra` because it is shorter.

cardano-api/gen/Test/Gen/Cardano/Api/Typed.hs
```diff
@@ -917,7 +933,7 @@ genProtocolParametersUpdate = do
   protocolUpdatePoolPledgeInfluence <- Gen.maybe genRationalInt64
   protocolUpdateMonetaryExpansion   <- Gen.maybe genRational
   protocolUpdateTreasuryCut         <- Gen.maybe genRational
-  protocolUpdateUTxOCostPerWord     <- Gen.maybe genLovelace
+  protocolUpdateUTxOCostPerWord     <- featureInEra @ProtocolUTxOCostPerWordFeature (pure Nothing) (const (Just <$> genLovelace)) era
```
newhoggy, line 936: The main part of the PR is implemented here and similar.

We only generate values for fields if they make sense for the given era.

Note, this code is _not yet_ type-safe because the type of `protocolUpdateUTxOCostPerWord` and `protocolUpdateUTxOCostPerByte` are both `Maybe Lovelace` and there is no defence against accidentally confusing the two in the code.

The intent is to eventually define these fields like this:

```
protocolUpdateUTxOCostPerWord :: FeatureValue ProtocolUTxOCostPerWordFeature era Lovelace
protocolUpdateUTxOCostPerByte :: FeatureValue ProtocolUTxOCostPerByteFeature era Lovelace
```

This would make the code type-safe because the code that checks the relevant feature for the field would not compile if they are accidentally mismatched.

Unfortunately this is not implemented here because doing so requires `ProtocolParameters` type to take an `era` type argument, but there are unresolved issues with making that change for reasons unrelated to this PR, but hopefully they can be resolved in consultation with the relevant people.

cardano-api/gen/Test/Gen/Cardano/Api/Typed.hs
```diff
@@ -703,6 +707,17 @@ genTxBody era = do
...
+-- | Generate a 'FeatureValue' for the given 'CardanoEra' with the provided generator.
+genFeatureValueInEra :: ()
+  => FeatureInEra feature
+  => Alternative f
+  => f a
+  -> CardanoEra era
+  -> f (FeatureValue feature era a)
+genFeatureValueInEra gen =
+  featureInEra (pure NoFeatureValue) $ \witness ->
+    pure NoFeatureValue <|> fmap (FeatureValue witness) gen
```
newhoggy, line 719: This function is not yet used because the relevant fields are not yet using the `FeatureValue` type, but it will be once that change is made (possibly in a future PR).

The advantage of having a `genFeatureValueInEra` is that using it is less error prone than exist practise which is to define per-feature generators manually.  It is very easy to accidentally not return `NoFeatureValue` in the case where the feature is supported in the given era in implementation of those manually written functions and that bug currently exists in the code.

carbolymer (approved):

PR #44, opened 2023-06-08
Type safe protocol parameters

newhoggy:
# Description

Add your description here, if it fixes a particular issue please provide a
[link](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword=)
to the issue.

# Changelog

```yaml
- description: |
    <insert-changelog-description-here>
  # no-changes: the API has not changed
  # compatible: the API has changed but is non-breaking
  # breaking: the API has changed in a breaking way
  compatibility: <no-api-changes|compatible|breaking>
  # feature: the change implements a new feature in the API
  # bugfix: the change fixes a bug in the API
  # test: the change fixes modifies tests
  # maintenance: the change involves something other than the API
  # If more than one is applicable, it may be put into a list.
  type: <feature|bugfix|test|maintenance>
```

# Checklist

  - golden tests
  - property tests
  - roundtrip tests
  - integration tests
- [ ] Self-reviewed the diff

# Note on CI
for you.


cardano-api/internal/Cardano/Api/EraCast.hs
```diff
@@ -26,3 +27,11 @@ class EraCast (f :: Type -> Type) where
           => CardanoEra toEra
           -> f fromEra
           -> Either EraCastError (f toEra)
+
+-- | Cast a value from one era to another.  Any data contained within
+-- that is not applicable in the target era is dropped.
+class EraCastLossy (f :: Type -> Type) where
+  eraCastLossy :: ()
+    => CardanoEra toEra -- ^ The era to cat=st to
```
carbolymer, line 35: typo in comment

carbolymer (approved):

cardano-api/internal/Cardano/Api/EraCast.hs
newhoggy: Fixed.  Thanks!

cardano-api/gen/Test/Gen/Cardano/Api/Typed.hs
```diff
@@ -899,7 +902,29 @@ genPraosNonce = makePraosNonce <$> Gen.bytes (Range.linear 0 32)
...
+genValidFeatured :: ()
+  => FeatureInEra feature
+  => CardanoEra era
+  -> Gen a
+  -> Gen (Maybe (Featured feature era a))
+genValidFeatured era g =
+  featureInEra
+    (pure Nothing)
+    (\w -> Just . Featured w <$> g)
+    era
```
newhoggy, line 914: This is the same as `genMaybeFeatured` except when the feature is valid, `Nothing` is not returned.

cardano-api/internal/Cardano/Api/Eras/Core.hs
```diff
@@ -179,6 +183,14 @@ inShelleyBasedEraFeature :: ()
 inShelleyBasedEraFeature era no yes =
   featureInShelleyBasedEra no yes era
 
+-- ----------------------------------------------------------------------------
+-- ToCardanoEra
+
+class ToCardanoEra (feature :: Type -> Type) where
+  toCardanoEra :: ()
+    => feature era
+    -> CardanoEra era
```
newhoggy, line 192: This generalises over `conwayEraOnwardsToCardanoEra`, `shelleyToBabbageEraToCardanoEra` etc.

Those other functions still exist.  This type-class enables `eraCastFeatured` function to be implemented.

cardano-api/internal/Cardano/Api/Eras/Core.hs
```diff
@@ -378,6 +390,10 @@ data InAnyCardanoEra thing where
                      -> thing era
                      -> InAnyCardanoEra thing
 
+-- | This pairs up some era-dependent type with a 'feature' value that tells
+-- us what feature involves, but hides the era type.
+data InAnyEra (feature :: Type -> Type) (t :: Type -> Type) where
+  InAnyEra :: feature era -> t era -> InAnyEra feature t
```
newhoggy, line 396: This allows the `era` type to be smuggled out of contexts where it otherwise wouldn't be possible due to the behaviour of GADTs.

newhoggy: See https://github.com/input-output-hk/cardano-cli/pull/162/files#r1290242532

cardano-api/internal/Cardano/Api/ProtocolParameters.hs
```diff
@@ -325,17 +328,17 @@ instance FromJSON ProtocolParameters where
...
+        <*> o .:?^ "utxoCostPerWord"
         <*> (fmap unCostModels <$> o .:? "costModels") .!= Map.empty
         <*> o .:? "executionUnitPrices"
         <*> o .:? "maxTxExecutionUnits"
         <*> o .:? "maxBlockExecutionUnits"
         <*> o .:? "maxValueSize"
         <*> o .:? "collateralPercentage"
         <*> o .:? "maxCollateralInputs"
-        <*> o .:? "utxoCostPerByte"
+        <*> o .:?^ "utxoCostPerByte"
```
newhoggy, line 339: This field has `Featured` in its type so we need to use a different operator to render it.

cardano-api/internal/Cardano/Api/Feature.hs
```diff
@@ -37,10 +59,62 @@ asFeaturedInEra :: ()
...
+    parseSupported v fe = Just . Featured fe <$> p v
+
+(.:?^) :: ()
+  => IsCardanoEra era
+  => FromJSON a
+  => FeatureInEra feature
+  => Object
+  -> Key
+  -> Parser (Maybe (Featured feature era a))
+(.:?^) = explicitParseFieldFeatureValue parseJSON
```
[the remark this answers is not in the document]
newhoggy: See https://github.com/input-output-hk/cardano-api/pull/44/files#r1290247409

cardano-api/internal/Cardano/Api/Feature.hs
```diff
@@ -37,10 +59,62 @@ asFeaturedInEra :: ()
...
+  featureInEra
+    (Left e)
+    (\w -> Right (Featured w a))
+    targetEra
+  where
+    e = EraCastError
+      { originalValue = a
+      , fromEra = toCardanoEra feature
+      , toEra = targetEra
+      }
```
newhoggy, line 120: `eraCastFeatured` is a function and not an `EraCast` type class instance.

The reason is that defining an `EraCast` instance is not possible to due the order of arguments of the `Featured` type:

```
Featured feature targetEra a
```

The type is defined such that `a` is the last argument, not `targetEra`.  If `targetEra` was the last argument then it would be possible to define a type class instance of `EraCast`.  However, in doing so we would lose some useful capabilities.  For example `Featured` would not be able to be a `Functor` anymore.

I think this is a good tradeoff.

cardano-api/internal/Cardano/Api/GenesisParameters.hs
```diff
@@ -33,7 +33,7 @@ import           Data.Time (NominalDiffTime, UTCTime)
 -- Genesis parameters
 --
 
-data GenesisParameters =
+data GenesisParameters era =
```
newhoggy, line 36: This is now parameterised by `era`.

cardano-api/internal/Cardano/Api/ProtocolParameters.hs
```diff
@@ -140,7 +143,7 @@ import           Text.PrettyBy.Default (display)
 --
 -- There are also parameters fixed in the Genesis file. See 'GenesisParameters'.
 --
-data ProtocolParameters =
+data ProtocolParameters era =
```
newhoggy, line 146: This is now parameterised by `era`.

cardano-api/internal/Cardano/Api/ProtocolParameters.hs
```diff
@@ -298,12 +301,12 @@ data ProtocolParameters =
        -- | Cost in ada per byte of UTxO storage.
        --
        -- /Introduced in Babbage/
-       protocolParamUTxOCostPerByte :: Maybe Lovelace
+       protocolParamUTxOCostPerByte :: Maybe (Featured ProtocolUTxOCostPerByteFeature era Lovelace)
```
newhoggy, line 304: This field exists or not depending on which `era` we are in as prescribed by the feature `ProtocolUTxOCostPerByteFeature`

cardano-api/internal/Cardano/Api/ProtocolParameters.hs
```diff
@@ -369,6 +372,67 @@ instance ToJSON ProtocolParameters where
...
+    newProtocolParamTreasuryCut         <- pp & protocolParamTreasuryCut          & pure
+    newProtocolParamUTxOCostPerWord     <- pp & protocolParamUTxOCostPerWord      & fmap (eraCastFeatured era) & sequence
+    newProtocolParamCostModels          <- pp & protocolParamCostModels           & pure
+    newProtocolParamPrices              <- pp & protocolParamPrices               & pure
+    newProtocolParamMaxTxExUnits        <- pp & protocolParamMaxTxExUnits         & pure
+    newProtocolParamMaxBlockExUnits     <- pp & protocolParamMaxBlockExUnits      & pure
+    newProtocolParamMaxValueSize        <- pp & protocolParamMaxValueSize         & pure
+    newProtocolParamCollateralPercent   <- pp & protocolParamCollateralPercent    & pure
+    newProtocolParamMaxCollateralInputs <- pp & protocolParamMaxCollateralInputs  & pure
+    newProtocolParamUTxOCostPerByte     <- pp & protocolParamUTxOCostPerByte      & fmap (eraCastFeatured era) & sequence
```
newhoggy, line 405: We refuse to cast `ProtocolParameters` to another `era` if doing so will discard information.

cardano-api/internal/Cardano/Api/ProtocolParameters.hs
```diff
@@ -257,7 +260,7 @@ data ProtocolParameters =
        -- | Cost in ada per word of UTxO storage.
        --
        -- /Introduced in Alonzo/
-       protocolParamUTxOCostPerWord :: Maybe Lovelace,
+       protocolParamUTxOCostPerWord :: Maybe (Featured ProtocolUTxOCostPerWordFeature era Lovelace),
```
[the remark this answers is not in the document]
ch1bo: Why do we have `Maybe (Featured ..)` things?

Are they optional in `ProtocolParameters` if they are enabled in an era? For example, is `protocolParamUTxOCostPerWord` ever allowed to be `Nothing` in `AlonzoEra`?

Inversely, if we want to construct a `ProtocolParameters BabbageEra`, we should not be allowed to set `protocolParamUTxOCostPerWord`, but should be forced to set `protocolParamUTxOCostPerByte` to construct a protocol parameters value.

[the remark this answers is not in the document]
newhoggy: Yeah, I'm punting on required-ness for the fields of this type.  `Featured` has `0` inhabitants if the feature doesn't support the era so it doesn't actually work if you remove the `Maybe`.

It may be the case that we have to use type families somehow in which case it might make sense to use the ledger types directly.

Jimbo4350 (changes requested):
I don't want to merge these changes as I'm working on getting rid of `ProtocolParameters`. After that is done we can revisit this PR and see what changes are relevant.

PR #182, opened 2023-08-11
Parameterising `ProtocolParameters` and `ProtocolParametersUpdate`

newhoggy:
# Changelog

```yaml
- description: |
    Features for protocol parameters and protocol parameter updates
  type:
```

# Context

The substantive commits in this PR are the last five commits.   The earlier commits are quality of life commits that are included in https://github.com/input-output-hk/cardano-api/pull/184.  For those earlier commits, please review them in that PR.

This PR introduces a brand new parameterised `ProtocolParameters era` data type that lives in `Cardano.Api.Domain.ProtocolParameters`.

The switch over to use this type has not happened yet.  Rather this PR is meant to demonstrate are particular approach to the parameterisation of `ProtocolParameters` for consideration as it provides a number of advantages.

## Quality of Life changes
It uses the quality of life changes from the earlier PR which provides new features that are useful but not specific to the parameterisation of `ProtocolParameters`.  This encourages code re-use.

## Newtype
It is a `newtype` wrapper around the `Ledger` type.  This means the type as simple as it can be whilst still being a distinct API type.  But offering no new structure, it relies completely on the `Ledger` for type cues.

The main motivation for the `newtype` is to allow for the `cardano-api` to control JSON and perhaps CBOR serialisation, but the `newtype` is not a necessary design aspect of the PR.

## Lenses
It uses lenses.  If the intent is to minimise and eventually get rid of `cardano-api`, the use of lenses in unavoidable going forward.  The introduction of lens provides a nice succinct way to both get and set fields in protocol parameters.  

Furthermore instead of have just plain lenses, we have functions that return a lens in return for a feature witness.  This is what bridges the GADT based `cardano-api` way of era type-check with Ledger style type family type class constraints.

## `ToJSON` and `FromJSON` intances
There is an incomplete, `ToJSON` and `FromJSON` instance.  These instances are currently for illustration only.

It shows that it is possible with a combination of `featurInEra`, feature witnesses and lenses that is possible to write a `ToJSON` and `FromJSON` instance.

Moreover, the implementation of those instances are highly regular.  The type class instance can be extended with copy-past code.

Finally, the implementation of the `FromJSON` instance use the applicative style which demonstrates that it is possible to use this approach with `optparse-applicative`.

I can provide an in-person meeting to walk through this code.

## No breaking changes
This PR offers no breaking changes only because the rest of `cardano-api` does not yet integrate with it.  This will be in a future PR.  This shows the PR can demonstrate its advantages whilst being self contained.

Depends on: https://github.com/input-output-hk/cardano-api/pull/184

# Checklist

  - golden tests
  - property tests
  - round trip tests
  - integration tests
- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/Domain/ExecutionUnits.hs
+++ b/cardano-api/internal/Cardano/Api/Domain/ExecutionUnits.hs
@@ -1,0 +1,79 @@
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies #-}
+
+module Cardano.Api.Domain.ExecutionUnits
+  ( ExecutionUnits (..)
+  , toAlonzoExUnits
+  , fromAlonzoExUnits
+  ) where
+
+import           Cardano.Api.SerialiseCBOR
+import           Cardano.Api.SerialiseJSON
+
+import qualified Cardano.Binary as CBOR
+import qualified Cardano.Ledger.Alonzo.Scripts as Alonzo
+
+import           Data.Aeson (object, (.:), (.=))
+import qualified Data.Aeson as Aeson
+import           Numeric.Natural (Natural)
+
+-- | Script execution units.
+--
+-- The units for how long a script executes for and how much memory it uses.
+-- This is used to declare the resources used by a particular use of a script.
+--
+-- This type is also used to describe the limits for the maximum overall
+-- execution units per transaction or per block.
+--
+data ExecutionUnits = ExecutionUnits
+  { -- | This corresponds roughly to the time to execute a script.
+    executionSteps  :: Natural
+  , -- | This corresponds roughly to the peak memory used during script
+    -- execution.
+    executionMemory :: Natural
+  }
+  deriving (Eq, Show)
+
+instance ToCBOR ExecutionUnits where
+  toCBOR ExecutionUnits {executionSteps, executionMemory} =
+      CBOR.encodeListLen 2
+   <> toCBOR executionSteps
+   <> toCBOR executionMemory
+
+instance FromCBOR ExecutionUnits where
+  fromCBOR = do
+    CBOR.enforceSize "ExecutionUnits" 2
+    ExecutionUnits
+      <$> fromCBOR
+      <*> fromCBOR
+
+instance ToJSON ExecutionUnits where
+  toJSON ExecutionUnits{executionSteps, executionMemory} =
+    object [ "steps"  .= executionSteps
+           , "memory" .= executionMemory ]
+
+instance FromJSON ExecutionUnits where
+  parseJSON =
+    Aeson.withObject "ExecutionUnits" $ \o ->
+      ExecutionUnits
+        <$> o .: "steps"
+        <*> o .: "memory"
+
+toAlonzoExUnits :: ExecutionUnits -> Alonzo.ExUnits
+toAlonzoExUnits ExecutionUnits{executionSteps, executionMemory} =
+  Alonzo.ExUnits
+  { Alonzo.exUnitsSteps = executionSteps
+  , Alonzo.exUnitsMem   = executionMemory
+  }
+
+fromAlonzoExUnits :: Alonzo.ExUnits -> ExecutionUnits
+fromAlonzoExUnits Alonzo.ExUnits{Alonzo.exUnitsSteps, Alonzo.exUnitsMem} =
+  ExecutionUnits
+  { executionSteps  = exUnitsSteps
+  , executionMemory = exUnitsMem
+  }
--- a/cardano-api/internal/Cardano/Api/Domain/ProtVer.hs
+++ b/cardano-api/internal/Cardano/Api/Domain/ProtVer.hs
@@ -1,0 +1,41 @@
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module Cardano.Api.Domain.ProtVer
+  ( ProtVer(..)
+  , reLedgerProtVerL
+  , unLedgerProtVerL
+  ) where
+
+import qualified Cardano.Ledger.BaseTypes as Ledger
+
+import           Control.Monad.Trans.Fail.String (runFail)
+import           Data.Aeson (FromJSON (..), ToJSON (..), object, withObject, (.:), (.=))
+import           Lens.Micro (Lens', lens)
+import           Numeric.Natural (Natural)
+
+newtype ProtVer = ProtVer
+  { unProtVer :: Ledger.ProtVer
+  }
+
+instance FromJSON ProtVer where
+  parseJSON v =
+    flip (withObject "ProtocolParameters") v $ \o -> do
+      major :: Natural <- o .: "major"
+      minor :: Natural <- o .: "minor"
+      let result = runFail $ (`Ledger.ProtVer` minor) <$> Ledger.mkVersion major
+      case result of
+        Right protVer -> pure $ ProtVer protVer
+        Left msg -> fail $ "Invalid protocol version: " <> show v <> " (" <> msg <> ")"
+
+instance ToJSON ProtVer where
+  toJSON (ProtVer (Ledger.ProtVer a b)) =
+    object
+      [ "major" .= (Ledger.getVersion a :: Natural)
+      , "minor" .= (b :: Natural)
+      ]
+
+reLedgerProtVerL :: Lens' ProtVer Ledger.ProtVer
+reLedgerProtVerL = lens unProtVer (const ProtVer)
+
+unLedgerProtVerL :: Lens' Ledger.ProtVer ProtVer
+unLedgerProtVerL = lens ProtVer (const unProtVer)
--- a/cardano-api/internal/Cardano/Api/Domain/ProtocolParameters.hs
+++ b/cardano-api/internal/Cardano/Api/Domain/ProtocolParameters.hs
@@ -1,0 +1,204 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE RankNTypes #-}
+
+module Cardano.Api.Domain.ProtocolParameters
+  ( ProtocolParameters(..)
+  , emptyProtocolParameters
+  , protocolParametersL
+  , protocolParamProtocolVersionL
+  , protocolParamDecentralizationL
+  , protocolParamExtraPraosEntropyL
+  , protocolParamMaxBlockHeaderSizeL
+  , protocolParamMaxBlockBodySizeL
+  , protocolParamMaxTxSizeL
+  , protocolParamTxFeeFixedL
+  , protocolParamTxFeePerByteL
+  , protocolParamMinUTxOValueL
+  , protocolParamStakeAddressDepositL
+  , protocolParamStakePoolDepositL
+  , protocolParamMinPoolCostL
+  , protocolParamPoolRetireMaxEpochL
+  , protocolParamStakePoolTargetNumL
+  , protocolParamPoolPledgeInfluenceL
+  , protocolParamMonetaryExpansionL
+  , protocolParamTreasuryCutL
+  , protocolParamUTxOCostPerWordL
+  , protocolParamCostModelsL
+  , protocolParamPricesL
+  , protocolParamMaxTxExUnitsL
+  , protocolParamMaxBlockExUnitsL
+  , protocolParamMaxValueSizeL
+  , protocolParamCollateralPercentL
+  , protocolParamMaxCollateralInputsL
+  , protocolParamUTxOCostPerByteL
+  ) where
+
+import           Cardano.Api.Eras.Constraints
+import           Cardano.Api.Eras.Core
+import           Cardano.Api.Feature.AlonzoEraOnly
+import           Cardano.Api.Feature.AlonzoEraOnwards
+import           Cardano.Api.Feature.BabbageEraOnwards
+import           Cardano.Api.Feature.ShelleyToAllegraEra
+import           Cardano.Api.Feature.ShelleyToAlonzoEra
+
+import qualified Cardano.Ledger.Alonzo.Core as Ledger
+import qualified Cardano.Ledger.Alonzo.Scripts as Ledger
+import qualified Cardano.Ledger.Babbage.Core as Ledger
+import qualified Cardano.Ledger.BaseTypes as Ledger
+import qualified Cardano.Ledger.Coin as Ledger
+
+import           GHC.Natural (Natural)
+import           Lens.Micro (Lens', lens)
+
+newtype ProtocolParameters era = ProtocolParameters
+  { unProtocolParameters :: Ledger.PParams (ShelleyLedgerEra era)
+  }
+
+protocolParametersL :: Lens' (ProtocolParameters era) (Ledger.PParams (ShelleyLedgerEra era))
+protocolParametersL = lens unProtocolParameters (\_ pp -> ProtocolParameters pp)
+
+emptyProtocolParameters :: ShelleyBasedEra era -> ProtocolParameters era
+emptyProtocolParameters w = shelleyBasedEraConstraints w $ ProtocolParameters Ledger.emptyPParams
+
+-- | Protocol version, major and minor. Updating the major version is
+-- used to trigger hard forks.
+protocolParamProtocolVersionL :: ShelleyBasedEra era -> Lens' (ProtocolParameters era) Ledger.ProtVer
+protocolParamProtocolVersionL w = shelleyBasedEraConstraints w $ protocolParametersL . Ledger.ppProtocolVersionL
+
+-- | The decentralization parameter. This is fraction of slots that
+-- belong to the BFT overlay schedule, rather than the Praos schedule.
+-- So 1 means fully centralised, while 0 means fully decentralised.
+--
+-- This is the \"d\" parameter from the design document.
+protocolParamDecentralizationL :: ShelleyToAlonzoEra era -> Lens' (ProtocolParameters era) Ledger.UnitInterval
+protocolParamDecentralizationL w = shelleyToAlonzoEraConstraints w $ protocolParametersL . Ledger.ppDL
+
+-- | Extra entropy for the Praos per-epoch nonce.
+--
+-- This can be used to add extra entropy during the decentralisation
+-- process. If the extra entropy can be demonstrated to be generated
+-- randomly then this method can be used to show that the initial
+-- federated operators did not subtly bias the initial schedule so that
+-- they retain undue influence after decentralisation.
+protocolParamExtraPraosEntropyL :: ShelleyToAlonzoEra era -> Lens' (ProtocolParameters era) Ledger.Nonce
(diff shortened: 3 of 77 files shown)


Jimbo4350 (changes requested):
I think this is more complicated than it needs to be

carbolymer (approved):

cardano-api/internal/Cardano/Api/Domain/ProtocolParameters.hs
```diff
@@ -0,0 +1,204 @@
...
+import qualified Cardano.Ledger.Babbage.Core as Ledger
+import qualified Cardano.Ledger.BaseTypes as Ledger
+import qualified Cardano.Ledger.Coin as Ledger
+
+import           GHC.Natural (Natural)
+import           Lens.Micro (Lens', lens)
+
+newtype ProtocolParameters era = ProtocolParameters
+  { unProtocolParameters :: Ledger.PParams (ShelleyLedgerEra era)
+  }
```
newhoggy, line 65: The goal of this PR is to be as thin a layer around ledger types as possible.  To that `ProtocolParameters` is a `newtype` around the corresponding ledger type with not additional type structure.

That the `newtype` exists at all is so that we have the ability to control JSON and serialisation instances.

cardano-api/internal/Cardano/Api/Domain/ProtocolParameters.hs
```diff
@@ -0,0 +1,314 @@
...
+  parseJSON =
+    case cardanoEra @era of
+      era ->
+        withObject "ProtocolParameters" $ \o -> do
+          inEraFeature era
+            (fail ("Supported Era " <> show era))
+            (\sbe ->
+              pure (shelleyBasedEraConstraints sbe $ emptyProtocolParameters sbe)
+                        -- Parser (ProtocolParameters era)
+                <**>  ( inEraFeature era (pure id) $ \w -> do
```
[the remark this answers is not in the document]
newhoggy: Added `parseJsonFieldByEra` to do the deduplication.

cardano-api/internal/Cardano/Api/Domain/ProtocolParameters.hs
```diff
@@ -0,0 +1,204 @@
...
+
+newtype ProtocolParameters era = ProtocolParameters
+  { unProtocolParameters :: Ledger.PParams (ShelleyLedgerEra era)
+  }
+
+protocolParametersL :: Lens' (ProtocolParameters era) (Ledger.PParams (ShelleyLedgerEra era))
+protocolParametersL = lens unProtocolParameters (\_ pp -> ProtocolParameters pp)
+
+emptyProtocolParameters :: ShelleyBasedEra era -> ProtocolParameters era
+emptyProtocolParameters w = shelleyBasedEraConstraints w $ ProtocolParameters Ledger.emptyPParams
```
newhoggy, line 170: Use the ledger to construct a default protocol parameters value.  Must be in `ShelleyBasedEra era` to do it.

cardano-api/internal/Cardano/Api/Domain/ProtocolParameters.hs
```diff
@@ -0,0 +1,204 @@
...
+
+-- | Protocol version, major and minor. Updating the major version is
+-- used to trigger hard forks.
+protocolParamProtocolVersionL :: ShelleyBasedEra era -> Lens' (ProtocolParameters era) Ledger.ProtVer
+protocolParamProtocolVersionL w = shelleyBasedEraConstraints w $ protocolParametersL . Ledger.ppProtocolVersionL
```
newhoggy: The rest of the file defines lenses which wrap the ledger one.

The context is that `cardano-api` uses GADT witnesses to decide whether an operation is valid in a given era  whereas `cardano-ledger` uses type class constraints.

The key idea is to take the ledger lenses which follow the ledger way and convert it into an api lens that uses the api way.

In order to do that, all we do is take a witness as an argument, use it to summon constraints and then using those constraints delegate to the ledger.

cardano-api/internal/Cardano/Api/Domain/ProtocolParameters.hs
```diff
@@ -80,8 +81,8 @@ protocolParamDecentralizationL w = shelleyToAlonzoEraConstraints w $ protocolPar
 -- randomly then this method can be used to show that the initial
 -- federated operators did not subtly bias the initial schedule so that
 -- they retain undue influence after decentralisation.
-protocolParamExtraPraosEntropyL :: ShelleyToAlonzoEra era -> Lens' (ProtocolParameters era) Ledger.Nonce
-protocolParamExtraPraosEntropyL w = shelleyToAlonzoEraConstraints w $ protocolParametersL . Ledger.ppExtraEntropyL
+protocolParamExtraPraosEntropyL :: ShelleyToAlonzoEra era -> Lens' (ProtocolParameters era) (Maybe PraosNonce)
+protocolParamExtraPraosEntropyL w = shelleyToAlonzoEraConstraints w $ protocolParametersL . Ledger.ppExtraEntropyL . unLedgerNonceL
```
newhoggy, line 193: If we ever want to expose an api type instead of a ledger type from our lenses, we can adjust them through lens composition.

cardano-api/internal/Cardano/Api/Domain/ProtocolParameters.hs
```diff
@@ -46,15 +56,104 @@ import qualified Cardano.Ledger.Alonzo.Core as Ledger
...
+          , "txFeePerByte"            .= justInEraFeature era (\w -> pp ^. protocolParamTxFeePerByteL w)
+          -- -- Alonzo era:
+          -- , "utxoCostPerWord"        .= protocolParamUTxOCostPerWord
+          -- , "costModels"             .= CostModels protocolParamCostModels
+          -- , "executionUnitPrices"    .= protocolParamPrices
+          -- , "maxTxExecutionUnits"    .= protocolParamMaxTxExUnits
+          -- , "maxBlockExecutionUnits" .= protocolParamMaxBlockExUnits
+          , "maxValueSize"            .= justInEraFeature era (\w -> pp ^. protocolParamMaxValueSizeL w)
+          , "collateralPercentage"    .= justInEraFeature era (\w -> pp ^. protocolParamCollateralPercentL w)
+          , "maxCollateralInputs"     .= justInEraFeature era (\w -> pp ^. protocolParamMaxCollateralInputsL w)
```
newhoggy, line 97: We can use `featureInEra` to generate `ToJSON` instances, but using `justInEraFeature` makes the code slightly shorter.

The nice thing here is we don't need to know what the type of the witness value `w`.  The lens function we choose determines that type for us.

Adding an extra field is just adding an extra line.

No decision needs to be made as to what type or constructor a field needs to be added to because we don't define any.  We just use the thin wrapper lenses.

If the ledger deprecates a field in the future, the only thing we need to adjust are the wrapper lenses.  This code will adapt accordingly with no changes.

cardano-api/internal/Cardano/Api/Domain/ProtocolParameters.hs
```diff
@@ -46,15 +56,104 @@ import qualified Cardano.Ledger.Alonzo.Core as Ledger
...
+          -- -- Babbage era:
+          -- , "utxoCostPerByte"        .= protocolParamUTxOCostPerByte
+          ]
+
+instance IsCardanoEra era => FromJSON (ProtocolParameters era) where
+  parseJSON =
+    case cardanoEra @era of
+      era ->
+        withObject "ProtocolParameters" $ \o -> do
```
newhoggy, line 106: This code is a little bit "interesting", but is understandable with explanation.

cardano-api/internal/Cardano/Api/Domain/ProtocolParameters.hs
```diff
@@ -46,15 +56,104 @@ import qualified Cardano.Ledger.Alonzo.Core as Ledger
...
+          pp <- inEraFeature era (fail ("Supported Era " <> show era)) (\w -> pure $ shelleyBasedEraConstraints w $ emptyProtocolParameters w)
```
newhoggy: The first thing is to generate a default `ProtocolParameters era` value.  The idea then is that for each potential field, we"mutate` this value to include and update to that field in the value.

cardano-api/internal/Cardano/Api/Domain/ProtocolParameters.hs
```diff
@@ -105,54 +104,64 @@ instance IsCardanoEra era => FromJSON (ProtocolParameters era) where
...
-          pp <- inEraFeature era (pure pp) $ \w -> do
-            v <- o .: "maxCollateralInputs"
-            pure (pp & protocolParamMaxCollateralInputsL w .~ v)
-          -- utxoCostPerByte <- o .:? "utxoCostPerByte"
-
-          pure pp
+          inEraFeature era
+            (fail ("Supported Era " <> show era))
+            (\sbe ->
+              pure (emptyProtocolParameters sbe)
```
newhoggy, line 125: We generate a default protocol parameters value just like before.

cardano-api/internal/Cardano/Api/Domain/ProtocolParameters.hs
```diff
@@ -105,54 +104,64 @@ instance IsCardanoEra era => FromJSON (ProtocolParameters era) where
...
+                <**>  ( inEraFeature era (pure id) $ \w -> do
+                          v <- o .: "protocolVersion"
+                          pure (& protocolParamProtocolVersionL w .~ v)
+                      )
+                <**>  ( inEraFeature era (pure id) $ \w -> do
+                          v <- o .: "extraPraosEntropy"
+                          pure (& protocolParamExtraPraosEntropyL w .~ v)
+                      )
```
newhoggy: The plan here is to do something analogous to this

```
defaultProtocolParameters
  & updateField1
  & updateField2
```

That is, we start with a default value of the type `ProtocolParameters era` then apply a function for each field that updates the value with the value for that field.

`updateField1` and `updateField2` have the type `ProtocolParameters era -> ProtocolParameters era`.

But we need to do this applicatively.

We can do this with the `<**>`.  This operator is exactly like the `<*>` operator but the arguments are reverse.

Observe:

(<*>) :: Applicative f => f (a -> b) -> f a -> f b
(<**>) :: Applicative f => f a -> f (a -> b) -> f b

What `<**>` allows us to do is write:

```
pure defaultProtocolParameters
  <**> updateField1
  <**> updateField2
```

And we are doing the exact same thing as the earlier example except applicatively which means this works with `optparse-applicative` and because every `Applicative` is a `Monad`, JSON parser works as well.

In the code , in each invocation of `<**>` we pass a value of the type `Parser (ProtocolParameters era -> ProtocolParameters era)`.

This parser will be `pure id` if the era is not supported.  If the era is supported however, we return a parser that mutates the `ProtocolParameter era` with an update to the relevant field.

cardano-api/internal/Cardano/Api/Domain/ProtocolParameters.hs
```diff
@@ -108,57 +123,19 @@ instance IsCardanoEra era => FromJSON (ProtocolParameters era) where
...
-                      )
+                <**>  parseJsonFieldByEra era o "protocolVersion"       (\w v -> protocolParamProtocolVersionL      w .~ v)
+                <**>  parseJsonFieldByEra era o "extraPraosEntropy"     (\w v -> protocolParamExtraPraosEntropyL    w .~ v)
+                <**>  parseJsonFieldByEra era o "maxBlockHeaderSize"    (\w v -> protocolParamMaxBlockHeaderSizeL   w .~ v)
+                <**>  parseJsonFieldByEra era o "maxBlockBodySize"      (\w v -> protocolParamMaxBlockBodySizeL     w .~ v)
+                <**>  parseJsonFieldByEra era o "maxTxSize"             (\w v -> protocolParamMaxTxSizeL            w .~ v)
+                <**>  parseJsonFieldByEra era o "txFeeFixed"            (\w v -> protocolParamTxFeeFixedL           w .~ v)
+                <**>  parseJsonFieldByEra era o "txFeePerByte"          (\w v -> protocolParamTxFeePerByteL         w .~ v)
+                <**>  parseJsonFieldByEra era o "minUTxOValue"          (\w v -> protocolParamMinUTxOValueL         w .~ v)
+                <**>  parseJsonFieldByEra era o "stakeAddressDeposit"   (\w v -> protocolParamStakeAddressDepositL  w .~ v)
```
newhoggy, line 134: The new `parseJsonFieldByEra` makes the code very pleasant.

Just like in the JSON printing case, JSON parsing is type-safe and era-sensitive.  The type `w` is automatically inferred to be whatever the lens we use demands.

If ledger changes removes the field in a future era, only the lens need change.  No change is required here.

cardano-api/internal/Cardano/Api/Domain/ProtocolParameters.hs
```diff
@@ -108,57 +123,19 @@ instance IsCardanoEra era => FromJSON (ProtocolParameters era) where
...
-                <**>  ( inEraFeature era (pure id) $ \w -> do
-                          v <- o .: "minUTxOValue"
-                          pure (protocolParamMinUTxOValueL w .~ v)
-                      )
-                <**>  ( inEraFeature era (pure id) $ \w -> do
-                          v <- o .: "stakeAddressDeposit"
-                          pure (protocolParamStakeAddressDepositL w .~ v)
-                      )
+                <**>  parseJsonFieldByEra era o "protocolVersion"       (\w v -> protocolParamProtocolVersionL      w .~ v)
+                <**>  parseJsonFieldByEra era o "extraPraosEntropy"     (\w v -> protocolParamExtraPraosEntropyL    w .~ v)
```
newhoggy, line 127: As a demonstration of a deprecate field handling, `protocolParamExtraPraosEntropyL` which demands `ShelleyToAlonzoEra era`, which means it is already deprecated, but no special handling is needed here.  `parseJsonFieldByEra` does all the era-switching for us thanks to `FeatureInEra`.

cardano-api/internal/Cardano/Api/Domain/ProtocolParameters.hs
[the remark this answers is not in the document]
newhoggy: See https://github.com/input-output-hk/cardano-api/pull/182/commits/7e2d409086385df91acb3d8725056023e10b5746#r1293399395

cardano-api/internal/Cardano/Api/Domain/ProtocolParameters.hs
newhoggy: One very relevant detail is in the context of `optparse-applicative` if the era is not supported and we use `pure id`, the CLI parser will recognise this as saying we do not accept a switch/argument so the CLI omits the unsupported CLI switch argument.

If the feature is supported an actual CLI parser that accepts the value for that field is return instead so the CLI help will include the field.

This is how the CLI can be made to be era-sensitive with very little maintenance work.

Jimbo4350 (changes requested):
I'm currently working on `EraBasedProtocolParametersUpdate` to construct a `PParams era`. We can revisit this after to see which parts are still relevant.

cardano-api/internal/Cardano/Api/Domain/ExecutionUnits.hs
```diff
@@ -0,0 +1,79 @@
...
+
+-- | Script execution units.
+--
+-- The units for how long a script executes for and how much memory it uses.
+-- This is used to declare the resources used by a particular use of a script.
+--
+-- This type is also used to describe the limits for the maximum overall
+-- execution units per transaction or per block.
+--
+data ExecutionUnits = ExecutionUnits
```
lehins, line 33: This could be a newtype around `ExUnits Nartural`

then all instances can be derived and conversion is just a `coerce`

cardano-api/internal/Cardano/Api/Domain/ProtVer.hs
```diff
@@ -0,0 +1,41 @@
...
+import           Control.Monad.Trans.Fail.String (runFail)
+import           Data.Aeson (FromJSON (..), ToJSON (..), object, withObject, (.:), (.=))
+import           Lens.Micro (Lens', lens)
+import           Numeric.Natural (Natural)
+
+newtype ProtVer = ProtVer
+  { unProtVer :: Ledger.ProtVer
+  }
+
+instance FromJSON ProtVer where
```
lehins, line 20: As mentioned in a private message this can be derived. The whole reason why json instances exist in ledger is for cardano-api

cardano-api/internal/Cardano/Api/Feature/AlonzoEraOnly.hs
```diff
@@ -0,0 +1,108 @@
...
+  , L.Crypto (L.EraCrypto (ShelleyLedgerEra era))
+  , L.Era (ShelleyLedgerEra era)
+  , L.EraCrypto (ShelleyLedgerEra era) ~ L.StandardCrypto
+  , L.EraPParams (ShelleyLedgerEra era)
+  , L.EraTx (ShelleyLedgerEra era)
+  , L.EraTxBody (ShelleyLedgerEra era)
+  , L.ExactEra L.AlonzoEra (ShelleyLedgerEra era)
+  , L.HashAnnotated (L.TxBody (ShelleyLedgerEra era)) L.EraIndependentTxBody L.StandardCrypto
+  , L.ShelleyEraTxBody (ShelleyLedgerEra era)
+  , L.ShelleyEraTxCert (ShelleyLedgerEra era)
```
lehins, line 85: All of this are redundant:
```suggestion
  , L.AlonzoEraPParams (ShelleyLedgerEra era)
  , L.EraCrypto (ShelleyLedgerEra era) ~ L.StandardCrypto
  , L.EraTx (ShelleyLedgerEra era)
  , L.ExactEra L.AlonzoEra (ShelleyLedgerEra era)
  , L.HashAnnotated (L.TxBody (ShelleyLedgerEra era)) L.EraIndependentTxBody L.StandardCrypto
  , L.ShelleyEraTxBody (ShelleyLedgerEra era)
  , L.ShelleyEraTxCert (ShelleyLedgerEra era)
```


in fact you could probably get away with a single constraint here:
```
  , ShelleyLedgerEra era ~ L.Alonzo

File: cardano-api/internal/Cardano/Api/Feature/AlonzoEraOnwards.hs [hunk 16]
```diff
@@ -0,0 +1,121 @@
...
+  , L.AlonzoEraPParams (ShelleyLedgerEra era)
+  , L.Crypto (L.EraCrypto (ShelleyLedgerEra era))
+  , L.Era (ShelleyLedgerEra era)
+  , L.EraCrypto (ShelleyLedgerEra era) ~ L.StandardCrypto
+  , L.EraPParams (ShelleyLedgerEra era)
+  , L.EraTx (ShelleyLedgerEra era)
+  , L.EraTxBody (ShelleyLedgerEra era)
+  , L.HashAnnotated (L.TxBody (ShelleyLedgerEra era)) L.EraIndependentTxBody L.StandardCrypto
+  , L.ShelleyEraTxBody (ShelleyLedgerEra era)
+  , L.ShelleyEraTxCert (ShelleyLedgerEra era)
```
lehins, line 94: Redundant, since they are inherited by other constraints that already present
```suggestion
  , L.EraCrypto (ShelleyLedgerEra era) ~ L.StandardCrypto
  , L.EraPParams (ShelleyLedgerEra era)
  , L.EraTx (ShelleyLedgerEra era)
  , L.HashAnnotated (L.TxBody (ShelleyLedgerEra era)) L.EraIndependentTxBody L.StandardCrypto
  , L.ShelleyEraTxBody (ShelleyLedgerEra era)
  , L.ShelleyEraTxCert (ShelleyLedgerEra era)
```

PR #196, opened 2023-08-17
Add consensus governance queries

carbolymer:
# Changelog

```yaml
- description: |
    Expose following queries from consensus:
      - GetGovState
      - GetDRepState
      - GetDRepStakeDistr
      - GetCommitteeState
      - GetConstitution
  type:
```

# Context
Implement consensus queries, similarly to: https://github.com/input-output-hk/cardano-ledger/pull/3648

# Checklist

  - golden tests
  - property tests
  - round trip tests
  - integration tests
- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/Query.hs
+++ b/cardano-api/internal/Cardano/Api/Query.hs
@@ -97,6 +97,8 @@ import qualified Cardano.Chain.Update.Validation.Interface as Byron.Update
 import qualified Cardano.Ledger.Api as L
 import           Cardano.Ledger.Binary
 import qualified Cardano.Ledger.Binary.Plain as Plain
+import qualified Cardano.Ledger.CertState as L
+import qualified Cardano.Ledger.Credential as L
 import qualified Cardano.Ledger.Credential as Shelley
 import           Cardano.Ledger.Crypto (Crypto)
 import qualified Cardano.Ledger.Shelley.API as Shelley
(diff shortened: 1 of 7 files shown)


cardano-api/internal/Cardano/Api/DRepMetadata.hs
```diff
@@ -90,3 +98,22 @@ validateAndHashDRepMetadata bs
       let mdh = DRepMetadataHash (Crypto.hashWith id bs)
       return (md, mdh)
   | otherwise = Left $ DRepMetadataInvalidLengthError 512 (BS.length bs)
+
+
+data DRepCredential
```
carbolymer, line 103: Probably no the best place to put this type. Suggestions are welcome. 😃

cardano-api/internal/Cardano/Api/Query.hs
```diff
@@ -316,7 +330,11 @@ instance NodeToClientVersionOf (QueryInShelleyBasedEra era result) where
   nodeToClientVersionOf (QueryPoolDistribution _) = NodeToClientV_14
   nodeToClientVersionOf (QueryStakeSnapshot _) = NodeToClientV_14
   nodeToClientVersionOf (QueryStakeDelegDeposits _) = NodeToClientV_15
-  nodeToClientVersionOf QueryConstitution = NodeToClientV_15
+  nodeToClientVersionOf QueryConstitution = NodeToClientV_16
```
carbolymer, line 333: Which version should be here? Quote from consensus repo:
```haskell
    | NodeToClientV_15
    -- ^ added `query` to NodeToClientVersionData
    | NodeToClientV_16
    -- ^ enabled @CardanoNodeToClientVersion11@, i.e., Conway and
    -- @GetStakeDelegDeposits@.
```
So `QueryStakeDelegDeposits` should be V16 as well?

cardano-api/internal/Cardano/Api/DRepMetadata.hs
Jimbo4350: We should be able to avoid this type. Where is it used?

cardano-api/internal/Cardano/Api/DRepMetadata.hs
```diff
@@ -13,6 +15,14 @@ module Cardano.Api.DRepMetadata (
     -- * Data family instances
     AsType(..),
     Hash(..),
+
+    -- * Auxilary aliases for ledger
+    DRepCredential,
```
Jimbo4350, line 20: Remember we are as much as possible trying to use the ledger types. Let's avoid adding any additional types (even type synonyms) as much as possible because the object is to rely on the ledger api. If you need ledger types in cardano-cli, expose them via `Cardano.Ledger.Api`

cardano-api/internal/Cardano/Api/Query.hs
Jimbo4350: You'll need to ask the consensus team @nfrisby

cardano-api/internal/Cardano/Api/DRepMetadata.hs
carbolymer: Removed

PR #220, opened 2023-08-29
Features we currently could use

newhoggy:
# Changelog

```yaml
- description: |
    Add the following features:
    * `AlonzoEraOnly`
    * `ShelleyToAllegraEra`
    * `BabbageEraOnwards`
    * `AlonzoEraOnwards`
    * `ShelleyToMaryEra`
    * `ShelleyToAlonzoEra`
  type:
```

# Context

These represent the era-ranges that are currently relevant to protocol parameters in the ledger.

# Checklist

  - golden tests
  - property tests
  - round trip tests
  - integration tests
- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/Feature/AlonzoEraOnly.hs
+++ b/cardano-api/internal/Cardano/Api/Feature/AlonzoEraOnly.hs
@@ -1,0 +1,108 @@
+{-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
+
+module Cardano.Api.Feature.AlonzoEraOnly
+  ( AlonzoEraOnly(..)
+  , IsAlonzoEraOnly(..)
+  , AnyAlonzoEraOnly(..)
+  , alonzoEraOnlyConstraints
+  , alonzoEraOnlyToCardanoEra
+  , alonzoEraOnlyToShelleyBasedEra
+  ) where
+
+import           Cardano.Api.Eras.Core
+import           Cardano.Api.Modes
+import           Cardano.Api.Query.Types
+
+import           Cardano.Binary
+import qualified Cardano.Crypto.Hash.Blake2b as Blake2b
+import qualified Cardano.Crypto.Hash.Class as C
+import qualified Cardano.Crypto.VRF as C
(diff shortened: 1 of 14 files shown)


cardano-api/internal/Cardano/Api/Feature/AlonzoEraOnly.hs
```diff
@@ -0,0 +1,108 @@
...
+  -> a
+alonzoEraOnlyConstraints = \case
+  AlonzoEraOnlyAlonzo  -> id
+
+alonzoEraOnlyToCardanoEra :: AlonzoEraOnly era -> CardanoEra era
+alonzoEraOnlyToCardanoEra = shelleyBasedToCardanoEra . alonzoEraOnlyToShelleyBasedEra
+
+alonzoEraOnlyToShelleyBasedEra :: AlonzoEraOnly era -> ShelleyBasedEra era
+alonzoEraOnlyToShelleyBasedEra = \case
+  AlonzoEraOnlyAlonzo  -> ShelleyBasedEraAlonzo
```
carbolymer, line 108: aren't those functions redundant? for what I understand it's possible to achieve the same using e.g. `maybeFeatureInEra`

cardano-api/src/Cardano/Api.hs
```diff
@@ -44,6 +44,48 @@ module Cardano.Api (
     shelleyToBabbageEraToCardanoEra,
     shelleyToBabbageEraToShelleyBasedEra,
 
+    ShelleyToAllegraEra(..),
```
carbolymer, line 47: Would it be possible to avoid creating types for ranges of eras? that's cumbersome and introduces overhead with every new era

carbolymer (commented):
https://github.com/input-output-hk/cardano-api/pull/220#discussion_r1308948662

cardano-api/internal/Cardano/Api/Feature/AlonzoEraOnly.hs
newhoggy: `maybeFeatureInEra` takes you from a large set to a smaller set only.  To expand we need something else.

cardano-api/src/Cardano/Api.hs
newhoggy: This is overhead, but we save elsewhere.  For example we currently case match on every era which is a lot of duplicated code.

newhoggy: See for example `fromShelleyTxOut`

cardano-api/internal/Cardano/Api/Feature/AlonzoEraOnly.hs
carbolymer: ah right, `featureInEra` takes `CardanoEra era`

carbolymer (approved):

PR #227, opened 2023-09-04
Update cardano-api to a newer cardano-ledger

coot:
# Changelog

```yaml
- description: |
    - Added `Cardano.Api.Certificate.filterUnRegDRepCreds`.
    - `queryStateForBalancedTx`  returns `DRep` credetnial.
    - `QueryDRepState` returns map indexed by `DRepCredential`.
    - `contructBalancedTx` and `evaluateTransactionBalance` takes `DRep` credential as an argument.
    - Updated IPC to `ouroboros-network-0.9.0.0` API

  type:
    - release        # related to a new release preparation
```

# Context

Make `cardano-api` compatible with a newer ledger input-output-hk/cardano-ledger@9e2f8151e3b9a0dde9faeb29a7dd2456e854427c.

# Checklist

  - golden tests
  - property tests
  - round trip tests
  - integration tests
- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/Address.hs
+++ b/cardano-api/internal/Cardano/Api/Address.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE LambdaCase #-}
--- a/cardano-api/internal/Cardano/Api/Convenience/Construction.hs
+++ b/cardano-api/internal/Cardano/Api/Convenience/Construction.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE DataKinds #-}
+
 -- | Convenience transaction construction functions
 --
 module Cardano.Api.Convenience.Construction (
@@ -24,6 +26,10 @@ import           Cardano.Api.TxBody
 import           Cardano.Api.Utils
 import           Cardano.Api.Value
 
+import qualified Cardano.Ledger.Api as L
+import qualified Cardano.Ledger.Credential as L
+import qualified Cardano.Ledger.Keys as L
+
 import qualified Data.List as List
 import qualified Data.Map.Strict as Map
 import           Data.Set (Set)
@@ -46,16 +52,17 @@ constructBalancedTx
   -> SystemStart
   -> Set PoolId       -- ^ The set of registered stake pools
   -> Map.Map StakeCredential Lovelace
+  -> Map.Map (L.Credential L.DRepRole L.StandardCrypto) Lovelace
   -> [ShelleyWitnessSigningKey]
   -> Either TxBodyErrorAutoBalance (Tx era)
 constructBalancedTx txbodcontent changeAddr mOverrideWits utxo lpp
                     ledgerEpochInfo systemStart stakePools
-                    stakeDelegDeposits shelleyWitSigningKeys = do
+                    stakeDelegDeposits drepDelegDeposits shelleyWitSigningKeys = do
 
   BalancedTxBody _ txbody _txBalanceOutput _fee
     <- makeTransactionBodyAutoBalance
          systemStart ledgerEpochInfo
-         lpp stakePools stakeDelegDeposits utxo txbodcontent
+         lpp stakePools stakeDelegDeposits drepDelegDeposits utxo txbodcontent
          changeAddr mOverrideWits
 
   let keyWits = map (makeShelleyKeyWitness txbody) shelleyWitSigningKeys
--- a/cardano-api/internal/Cardano/Api/Convenience/Query.hs
+++ b/cardano-api/internal/Cardano/Api/Convenience/Query.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE DataKinds #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeFamilies #-}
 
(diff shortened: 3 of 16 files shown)


cardano-api/internal/Cardano/Api/Convenience/Query.hs
```diff
@@ -90,6 +94,7 @@ queryStateForBalancedTx era allTxIns certs = runExceptT $ do
     & onNothing (left (EraConsensusModeMismatch (AnyConsensusMode CardanoMode) (getIsCardanoEraConstraint era $ AnyCardanoEra era)))
 
   let stakeCreds = Set.fromList $ mapMaybe (filterUnRegCreds sbe) certs
+      drepCreds  = undefined :: Set DRepCredential -- TODO: should `drepCreds` come from certs?
```
coot, line 97: @lehins were should `drepCreds` come from. They are passed to the `QueryDRepState` which returns [drepDeposit](https://input-output-hk.github.io/cardano-ledger/cardano-ledger-core/Cardano-Ledger-DRepDistr.html#t:DRepState) from `DRepState`.  The results of `queryStateForBalancedTx` are used to construct a transaction with `makeTransactionBodyAutoBalance` which is also update in this commit.

coot: `ConwayTxCert` contains [`ConwayRegDRep`](https://input-output-hk.github.io/cardano-ledger/cardano-ledger-conway/Cardano-Ledger-Conway-TxCert.html#t:ConwayGovCert), is this the right source?

lehins: Yes, you got it. `QueryDRepState` is the right query, except you need to look through the DRep unregistration certificates, instead of registration. Because the `drepCreds` need to be a `Map DRepCredential Coin`, where coin is the deposit that was left upon registration and will be refunded in this transaction that is being built. So here is the value that you need:
https://github.com/input-output-hk/cardano-ledger/blob/453d19da239d0eb123a98aabac2f74f5bfdb41f7/libs/cardano-ledger-core/src/Cardano/Ledger/DRepDistr.hs#L54

And this is the pattern synonym for the certificates that you need to collect from the transaction: https://github.com/input-output-hk/cardano-ledger/blob/453d19da239d0eb123a98aabac2f74f5bfdb41f7/libs/cardano-ledger-api/src/Cardano/Ledger/Api/Tx/Cert.hs#L60

coot: Thanks, here it is:

* [queryStateForBalancedTx](https://github.com/input-output-hk/cardano-api/blob/coot/cardano-ledger-conway-1.8/cardano-api/internal/Cardano/Api/Convenience/Query.hs?plain=1#L97)
* [filterUnRegDRepCreds](https://github.com/input-output-hk/cardano-api/blob/coot/cardano-ledger-conway-1.8/cardano-api/internal/Cardano/Api/Certificate.hs#L720-L726)

lehins: Perfect :+1:

cardano-api/internal/Cardano/Api/Address.hs
```diff
@@ -551,6 +558,10 @@ instance ToJSON StakeCredential where
         StakeCredentialByScript scriptHash ->
           ["stakingScriptHash" .= serialiseToRawBytesHexText scriptHash]
 
+newtype DRepCredential = DRepCredential { unDRepCredential :: Shelley.Credential 'Shelley.DRepRole StandardCrypto }
```
Jimbo4350, line 561: Let's use the edger type directly

cardano-api/internal/Cardano/Api/Convenience/Construction.hs
```diff
@@ -46,16 +46,17 @@ constructBalancedTx
   -> SystemStart
   -> Set PoolId       -- ^ The set of registered stake pools
   -> Map.Map StakeCredential Lovelace
+  -> Map.Map DRepCredential Lovelace
```
Jimbo4350, line 49: @newhoggy @carbolymer what do you think about a single sum type to handle stake and drep credentials?

Jimbo4350 (changes requested):

cardano-api/internal/Cardano/Api/Convenience/Construction.hs
newhoggy: I think if the ledger types are distinct, I prefer to keep them separate and if they are not, I prefer to new the ledger type itself or a `newtype` wrapper over it.

carbolymer: Isn't that what ledger `Credential` type already is?

cardano-api/internal/Cardano/Api/Address.hs
coot: Note that newtype wrappers can help you in the future protect upstream from breaking ledger changes.

Jimbo4350: I want to as much as possible not maintain any ledger related types in cardano-api.

coot: Resolved.

coot: Note that `newtypes` help with two things:

* protect upstream libraries
* hide unnecessary details (in this case not exposing two type parameters which are irrelevant for upstream users)

They also can simplify maintnance, since they limit what needs to changed in the provided API.

newhoggy (approved):

PR #244, opened 2023-09-14
Delete `TxFeesExplicitInEra` and `TxFeesImplicitInEra`

newhoggy:
# Changelog

```yaml
- description: |
    Delete `TxFeesExplicitInEra` and `TxFeesImplicitInEra`
    New `ByronEraOnly` feature
    Move `ShelleyBasedEra` into its own module as it is a legitimate feature
  type:
```

# Context

Feature types make these types unnecessary.

# Checklist

  - golden tests
  - property tests
  - round trip tests
  - integration tests
- [ ] Self-reviewed the diff


cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -983,7 +981,7 @@ makeTransactionBodyAutoBalance systemstart history lpp@(LedgerProtocolParameters
     let (dummyCollRet, dummyTotColl) = maybeDummyTotalCollAndCollReturnOutput txbodycontent changeaddr
     txbody1 <- first TxBodyError $ -- TODO: impossible to fail now
                createAndValidateTransactionBody txbodycontent1 {
-                 txFee  = TxFeeExplicit explicitTxFees $ Lovelace (2^(32 :: Integer) - 1),
+                 txFee  = TxFeeExplicit sbe $ Lovelace (2^(32 :: Integer) - 1),
```
newhoggy, line 984: We can just provide `sbe` as the witness instead of make up another one.

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -1566,24 +1565,19 @@ parseHash asType = do
...
-  MaryEra -> TxFeeExplicit TxFeesExplicitInMaryEra mempty
-  AlonzoEra -> TxFeeExplicit TxFeesExplicitInAlonzoEra mempty
-  BabbageEra -> TxFeeExplicit TxFeesExplicitInBabbageEra mempty
-  ConwayEra -> TxFeeExplicit TxFeesExplicitInConwayEra mempty
-
+defaultTxFee =
+  caseByronOrShelleyBasedEra
+    TxFeeImplicit
+    (\w -> TxFeeExplicit w mempty)
+    (cardanoEra @era)
```
newhoggy, line 1580: We can using casing functions now, avoiding the need to enumerate all eras.

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -2963,25 +2957,9 @@ fromLedgerTxReturnCollateral sbe txbody =
...
-      fromShelleyLovelace $ body ^. L.feeTxBodyL
-    ShelleyBasedEraBabbage ->
-      TxFeeExplicit TxFeesExplicitInBabbageEra $
-      fromShelleyLovelace $ body ^. L.feeTxBodyL
-    ShelleyBasedEraConway ->
-      TxFeeExplicit TxFeesExplicitInConwayEra $
-      fromShelleyLovelace $ body ^. L.feeTxBodyL
+  shelleyBasedEraConstraints sbe
+    $ TxFeeExplicit sbe
+    $ fromShelleyLovelace $ body ^. L.feeTxBodyL
```
newhoggy, line 2962: We can use `sbe` here as well and not have to enumerate all eras and use constraints functions to summon constraints.

Jimbo4350 (changes requested):
No, I disagree. It's going to be confusing having the era related code in different name spaces. The era itself isn't a feature, it represents a boundary IMO.

cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -984,7 +981,7 @@ makeTransactionBodyAutoBalance systemstart history lpp@(LedgerProtocolParameters
...
```
Jimbo4350, line 984: The explicit/implicit refactoring looks good. I would approve it in a separate PR.

newhoggy:
> No, I disagree. The era itself isn't a feature, it represents a boundary between feature sets IMO.

Maybe `Feature` is not the right word?  Stuff like `ShelleyToBabbageEra` and `ConwayEraOnwards` are more like era selectors for other things that are features.  `BabbageEraOnly` has that kind of shape.

carbolymer (approved):

Jimbo4350:
> > No, I disagree. The era itself isn't a feature, it represents a boundary between feature sets IMO.
> 
> Maybe `Feature` is not the right word? Stuff like `ShelleyToBabbageEra` and `ConwayEraOnwards` are more like era selectors for other things that are features. `BabbageEraOnly` has that kind of shape.
> 
> If we had another word to replace `Feature` that fit, I think it would work.

I don't think it's worth the churn. However the fee refactoring is good.

newhoggy:
`FeatureInEra` has been renamed to `Eon` so the misnaming doesn't apply anymore.

PR #299, opened 2023-10-05
Reduce constraint usage with eons

newhoggy:
# Changelog

```yaml
- description: |
    Reduce constraint usage with eons
  type:
```

# Context

By using eons to summon constraints our functions can have fewer or no constraints in their type signatures which simplifies code here as well as later in `cardano-cli`.

# Checklist

  - golden tests
  - property tests
  - round trip tests
  - integration tests
- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/Fees.hs
+++ b/cardano-api/internal/Cardano/Api/Fees.hs
@@ -48,9 +48,9 @@ module Cardano.Api.Fees (
 import           Cardano.Api.Address
 import           Cardano.Api.Certificate
 import           Cardano.Api.Eon.BabbageEraOnwards
-import           Cardano.Api.Eon.ByronToAllegraEra
 import           Cardano.Api.Eon.MaryEraOnwards
 import           Cardano.Api.Eon.ShelleyBasedEra
+import           Cardano.Api.Eon.ShelleyToAllegraEra
 import           Cardano.Api.Eras.Case
 import           Cardano.Api.Eras.Constraints
 import           Cardano.Api.Eras.Core
@@ -76,9 +76,7 @@ import qualified Cardano.Ledger.Coin as Ledger
 import           Cardano.Ledger.Credential as Ledger (Credential)
 import qualified Cardano.Ledger.Crypto as Ledger
 import qualified Cardano.Ledger.Keys as Ledger
-import           Cardano.Ledger.Mary.Value (MaryValue)
 import qualified Cardano.Ledger.Shelley.API.Wallet as Ledger (evaluateTransactionFee)
-import           Cardano.Ledger.UTxO as Ledger (EraUTxO)
 import qualified Ouroboros.Consensus.HardFork.History as Consensus
 import qualified PlutusLedgerApi.V1 as Plutus
 
@@ -588,10 +586,10 @@ evaluateTransactionBalance _ _ _ _ _ (ByronTxBody _) =
 
 evaluateTransactionBalance pp poolids stakeDelegDeposits drepDelegDeposits utxo
                            (ShelleyTxBody sbe txbody _ _ _ _) =
-    withLedgerConstraints
+    caseShelleyToAllegraOrMaryEraOnwards
+      evalAdaOnly
+      evalMultiAsset
       sbe
-      (shelleyBasedEraConstraints sbe evalAdaOnly)
-      (shelleyBasedEraConstraints sbe evalMultiAsset)
   where
     isRegPool :: Ledger.KeyHash Ledger.StakePool Ledger.StandardCrypto -> Bool
     isRegPool kh = StakePoolKeyHash kh `Set.member` poolids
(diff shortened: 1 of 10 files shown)


cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -608,81 +606,29 @@ evaluateTransactionBalance pp poolids stakeDelegDeposits drepDelegDeposits utxo
...
+        $ TxOutValue w . fromMaryValue
+        $ L.evalBalanceTxBody
+            pp
+            lookupDelegDeposit
+            lookupDRepDeposit
+            isRegPool
+            (toLedgerUTxO sbe utxo)
+            txbody
+
+    evalAdaOnly :: ShelleyToAllegraEra era -> TxOutValue era
```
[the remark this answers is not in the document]
newhoggy: Yes.  The original code had `ShelleyLedgerEra era` in the constraints so you could not call it in the `Byron` era anyway.

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -2851,43 +2788,21 @@ convScriptData era txOuts scriptWitnesses =
...
-        )
-        redeemers
-        datums
+  -> StrictMaybe (L.ScriptIntegrityHash (Ledger.EraCrypto (ShelleyLedgerEra era)))
+convPParamsToScriptIntegrityHash w txProtocolParams redeemers datums languages =
+  alonzoEraOnwardsConstraints w $
+    case txProtocolParams of
+      BuildTxWith Nothing -> SNothing
+      BuildTxWith (Just (LedgerProtocolParameters pp)) ->
+        Alonzo.hashScriptIntegrity (Set.map (L.getLanguageView pp) languages) redeemers datums
```
newhoggy, line 2805: `convPParamsToScriptIntegrityHash` simplified a lot for three reasons:

* The function was returning and `Either` when `Left` is never returned.  This imposed on callers to handle the `Left` case as well, which is unnecessary.
* The function is called with `ShelleyBasedEra` but the constraint makes it not possible for eras before alonzo anyway.  Changing this to `AlonzoEraOnwards` means we don't have to `case` on those earlier eras that never happen.
* `alonzoEraOnwardsConstraints` removes the need to handle the remaining cases.

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -2353,71 +2301,61 @@ fromLedgerTxOuts sbe body scriptdata =
     selectTxDatums  TxBodyNoScriptData                            = Map.empty
     selectTxDatums (TxBodyScriptData _ (Alonzo.TxDats' datums) _) = datums
 
-fromAlonzoTxOut :: forall era ledgerera.
-                   IsShelleyBasedEra era
-                => L.AlonzoEraTxOut ledgerera
-                => Ledger.EraCrypto ledgerera ~ StandardCrypto
-                => Ledger.Value ledgerera ~ MaryValue StandardCrypto
```
newhoggy, line 2360: Removing constraints like these means the caller no longer needs to supply them.

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -1588,109 +1588,76 @@ instance IsCardanoEra era => SerialiseAsCBOR (TxBody era) where
...
-         <> CBOR.encodeNullMaybe CBOR.encCBOR txmetadata
-      ShelleyBasedEraBabbage ->
-        CBOR.serialize' (L.eraProtVerLow @L.Babbage)
-          $ CBOR.encodeListLen 4
-         <> CBOR.encCBOR txbody
-         <> CBOR.encCBOR txscripts
-         <> CBOR.encCBOR (txScriptValidityToScriptValidity scriptValidity)
-         <> CBOR.encodeNullMaybe CBOR.encCBOR txmetadata
-      ShelleyBasedEraConway ->
-        CBOR.serialize' (L.eraProtVerLow @L.Babbage)
```
newhoggy, line 1648: This looks like a bug.  We say `Babbage` when doing `Conway` stuff.

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -1588,109 +1588,76 @@ instance IsCardanoEra era => SerialiseAsCBOR (TxBody era) where
...
-                            (TxBodyScriptData _ datums redeemers)
+  caseShelleyToMaryOrAlonzoEraOnwards
+    (const $ CBOR.serialize' (L.eraProtVerLow @(ShelleyLedgerEra era)) $ mconcat
+      [ CBOR.encodeListLen 3
+      , CBOR.encCBOR txbody
+      , CBOR.encCBOR txscripts
+      , CBOR.encodeNullMaybe CBOR.encCBOR txmetadata
+      ]
+    )
+    (const $ CBOR.serialize' (L.eraProtVerLow @(ShelleyLedgerEra era)) $ mconcat
```
newhoggy, line 1623: Using the `era` parameter means we can't get this wrong:

https://github.com/input-output-hk/cardano-api/pull/299/files#r1349452875

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -1951,23 +1916,18 @@ createTransactionBody sbe txBodyContent =
...
-                apiProtocolParameters
-                redeemers
-                datums
-                languages
+        let sData = convScriptData sbe apiTxOuts apiScriptWitnesses
+        let scriptIntegrityHash =
+              case sData of
+                TxBodyNoScriptData -> SNothing
+                TxBodyScriptData w datums redeemers ->
+                  convPParamsToScriptIntegrityHash w apiProtocolParameters redeemers datums languages
```
newhoggy, line 1930: Now that `convPParamsToScriptIntegrityHash` takes the more appropriate `AlonzoEraOnwards` witness (as opposed to `ShelleyBasedEra` we can make use of the witness supplied by `TxBodyScriptData`.

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -2353,71 +2301,61 @@ fromLedgerTxOuts sbe body scriptdata =
...
-      | Just d <- Map.lookup dh txdatums  = TxOutDatumInTx' w (ScriptDataHash dh) (fromAlonzoData d)
-      | otherwise                         = TxOutDatumHash w (ScriptDataHash dh)
-
-fromBabbageTxOut
-  :: forall ledgerera era.
-     L.BabbageEraTxOut ledgerera
-  => IsShelleyBasedEra era
-  => ShelleyLedgerEra era ~ ledgerera
-  => Ledger.EraCrypto ledgerera ~ StandardCrypto
-  => Ledger.Value ledgerera ~ MaryValue StandardCrypto
```
newhoggy, line 2387: Removing constraints relieves the caller of the burden of providing them.

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -2353,71 +2301,61 @@ fromLedgerTxOuts sbe body scriptdata =
...
-  => MaryEraOnwards era
-  -> AlonzoEraOnwards era
-  -> BabbageEraOnwards era
```
newhoggy, line 2390: We were previously demanding callers provide three witnesses.  We only need the `BabbageEraOnwards` because that's the most strict and the others are redundant.  Demanding fewer witness relieves the called of the burden of providing so many.

cardano-api/internal/Cardano/Api/TxBody.hs
Jimbo4350: I think this was implemented when there was no Conway era available.

cardano-api/internal/Cardano/Api/TxBody.hs
Jimbo4350: I think we can actually delete this function (or we should) because we no longer use the intermediate tx body format.

Jimbo4350 (approved):
Nice :+1:

PR #310, opened 2023-10-10
Fewer constraints in functions

newhoggy:
# Changelog

```yaml
- description: |
    Fewer constraints in functions
  type:
```

# Context

# Checklist

  - golden tests
  - property tests
  - round trip tests
  - integration tests
- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/Eon/MaryEraOnwards.hs
+++ b/cardano-api/internal/Cardano/Api/Eon/MaryEraOnwards.hs
@@ -31,6 +31,7 @@ import qualified Cardano.Ledger.BaseTypes as L
 import qualified Cardano.Ledger.Core as L
 import qualified Cardano.Ledger.Mary.Value as L
 import qualified Cardano.Ledger.SafeHash as L
+import qualified Cardano.Ledger.UTxO as L
 import qualified Ouroboros.Consensus.Protocol.Abstract as Consensus
 import qualified Ouroboros.Consensus.Protocol.Praos.Common as Consensus
 import qualified Ouroboros.Consensus.Shelley.Ledger as Consensus
@@ -76,6 +77,7 @@ type MaryEraOnwardsConstraints era =
   , L.EraPParams (ShelleyLedgerEra era)
   , L.EraTx (ShelleyLedgerEra era)
   , L.EraTxBody (ShelleyLedgerEra era)
+  , L.EraUTxO (ShelleyLedgerEra era)
   , L.HashAnnotated (L.TxBody (ShelleyLedgerEra era)) L.EraIndependentTxBody L.StandardCrypto
   , L.MaryEraTxBody (ShelleyLedgerEra era)
   , L.ShelleyEraTxBody (ShelleyLedgerEra era)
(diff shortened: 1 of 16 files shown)


cardano-api/internal/Cardano/Api/Eras/Case.hs
```diff
@@ -217,3 +219,17 @@ alonzoEraOnlyToAlonzoEraOnwards :: ()
   -> AlonzoEraOnwards era
 alonzoEraOnlyToAlonzoEraOnwards = \case
   AlonzoEraOnlyAlonzo -> AlonzoEraOnwardsAlonzo
+
+babbageEraOnwardsToMaryEraOnwards :: ()
+  => BabbageEraOnwards era
+  -> MaryEraOnwards era
+babbageEraOnwardsToMaryEraOnwards = \case
+  BabbageEraOnwardsBabbage -> MaryEraOnwardsBabbage
```
smelc, line 227: Aren't you possibly casting a `babbage` era into a `mary` one here? Why is this valid?

newhoggy: This does not change the era.  In the terms `BabbageEraOnwardsBabbage` and `MaryEraOnwardsBabbage`, the suffixes says the era is `Babbage` in both cases, hence there is not era cast.

What this is doing relaxing the eon type from one that _allows_ all eras from Babbage onwards to one that allows all eras from Mary onwards.  We say relax because the target eon is larger so the operation cannot fail.

smelc: Ohhh woow, I thought `MaryEraOnwardsBabbage` meant "an unknown era in [Mary, ..., Babbage]" :facepalm:

newhoggy: That would be `MaryToBabbageEra` which has these inhabitants: `MaryToBabbageEraMary` and `MaryToBabbageEraBabbage`.

cardano-api/internal/Cardano/Api/Eon/MaryEraOnwards.hs
```diff
@@ -76,6 +77,7 @@ type MaryEraOnwardsConstraints era =
   , L.EraPParams (ShelleyLedgerEra era)
   , L.EraTx (ShelleyLedgerEra era)
   , L.EraTxBody (ShelleyLedgerEra era)
+  , L.EraUTxO (ShelleyLedgerEra era)
```
smelc, line 80: How do you know that it is valid to add such a constraint here? Is this something you infer from the ledger?

smelc (approved):

cardano-api/internal/Cardano/Api/Eon/MaryEraOnwards.hs
newhoggy: If you can add and the the constraints function still compiles, in this case `maryEraOnwardsConstraints`, then adding the constraint is valid.

This constraint was added because code in `MaryEraOnwards` needed the constraint.

smelc: And the `maryEraOnwardsConstraints` obtains the constraints in context thanks to the various imports right?

So in the end, this list of constraints here show what we _actually_ need to use (as opposed to the full set of constraints if we were importing everything available)?

newhoggy: `maryEraOnwardsConstraints` obtains the constraints by `case` matching on the GADT.  The case match tells the compiler what the specific era that is being used is after which the compiler can figure out the constraints.

The imports don't play into it except to allow use to write the `XConstraints` type and the `Constraints` type tells the compiler which of those constraints it knows about should be captured for the continuation.

PR #312, opened 2023-10-12
New `MaryEraOnly` eon. Disjoint functions

newhoggy:
# Changelog

```yaml
- description: |
    New `MaryEraOnly` eon
    New functions:
    - `caseMaryEraOnlyOrAlonzoEraOnwards`
    - `disjointByronEraOnlyAndShelleyBasedEra`
    - `disjointAlonzoEraOnlyAndBabbageEraOnwards`
    Deprecate:
    - `noByronEraInShelleyBasedEra`.  Use `disjointByronEraOnlyAndShelleyBasedEra` instead.
  type:
```

# Context

`MaryEraOnly` is not used yet, its use is anticipated as there some Mary-specific code exists.

The `disjointX` functions are used to prove that two eons don't overlap so the cases associated with those eons don't have to be handled.

This is similar to how use sometimes do this `case someValue of {}`.

# Checklist

  - golden tests
  - property tests
  - round trip tests
  - integration tests
- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/TxBody.hs
+++ b/cardano-api/internal/Cardano/Api/TxBody.hs
@@ -2656,7 +2656,7 @@ convWithdrawals txWithdrawals =
 
 convTransactionFee :: ShelleyBasedEra era -> TxFee era -> Ledger.Coin
 convTransactionFee sbe = \case
-  TxFeeImplicit w  -> noByronEraInShelleyBasedEra sbe w
+  TxFeeImplicit w  -> disjointByronEraOnlyAndShelleyBasedEra w sbe
   TxFeeExplicit _ fee -> toShelleyLovelace fee
 
 convValidityInterval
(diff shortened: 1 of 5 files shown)


cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -2656,7 +2656,7 @@ convWithdrawals txWithdrawals =
 
 convTransactionFee :: ShelleyBasedEra era -> TxFee era -> Ledger.Coin
 convTransactionFee sbe = \case
-  TxFeeImplicit w  -> noByronEraInShelleyBasedEra sbe w
+  TxFeeImplicit w  -> disjointByronEraOnlyAndShelleyBasedEra w sbe
```
newhoggy, line 2659: Here we are using `disjointByronEraOnlyAndShelleyBasedEra` to prove that the `w` and `sbe` eons don't overlap so this case doesn't need to be handled.

In this case `w` happens to be `ByronEraOnly` which does not overlap with `ShelleyBasedEra`.

smelc (approved):
LGTM :beers:

PR #333, opened 2023-10-20
Simplify `createTransactionBody`

newhoggy:
# Changelog

```yaml
- description: |
    Simplify `createTransactionBody`
  type:
```

# Context

Includes commits from https://github.com/input-output-hk/cardano-api/pull/333 so review that first.

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/TxBody.hs
+++ b/cardano-api/internal/Cardano/Api/TxBody.hs
@@ -1804,46 +1804,46 @@ createTransactionBody
   -> TxBodyContent BuildTx era
   -> Either TxBodyError (TxBody era)
 createTransactionBody sbe txBodyContent =
-  shelleyBasedEraConstraints sbe $
-  let apiTxOuts = txOuts txBodyContent
-      apiScriptWitnesses = collectTxBodyScriptWitnesses sbe txBodyContent
-      apiScriptValidity = txScriptValidity txBodyContent
-      apiMintValue = txMintValue txBodyContent
-      apiProtocolParameters = txProtocolParams txBodyContent
-      apiCollateralTxIns = txInsCollateral txBodyContent
-      apiReferenceInputs = txInsReference txBodyContent
-      apiExtraKeyWitnesses = txExtraKeyWits txBodyContent
-      apiReturnCollateral = txReturnCollateral txBodyContent
-      apiTotalCollateral = txTotalCollateral txBodyContent
+  shelleyBasedEraConstraints sbe $ do
+    let apiTxOuts = txOuts txBodyContent
+        apiScriptWitnesses = collectTxBodyScriptWitnesses sbe txBodyContent
+        apiScriptValidity = txScriptValidity txBodyContent
+        apiMintValue = txMintValue txBodyContent
+        apiProtocolParameters = txProtocolParams txBodyContent
+        apiCollateralTxIns = txInsCollateral txBodyContent
+        apiReferenceInputs = txInsReference txBodyContent
+        apiExtraKeyWitnesses = txExtraKeyWits txBodyContent
+        apiReturnCollateral = txReturnCollateral txBodyContent
+        apiTotalCollateral = txTotalCollateral txBodyContent
 
-      -- Ledger types
-      collTxIns = convCollateralTxIns apiCollateralTxIns
-      refTxIns = convReferenceInputs apiReferenceInputs
-      returnCollateral = convReturnCollateral sbe apiReturnCollateral
-      totalCollateral = convTotalCollateral apiTotalCollateral
-      certs = convCertificates sbe $ txCertificates txBodyContent
-      txAuxData = toAuxiliaryData sbe (txMetadata txBodyContent) (txAuxScripts txBodyContent)
-      scripts = convScripts apiScriptWitnesses
-      languages = convLanguages apiScriptWitnesses
+        -- Ledger types
+        collTxIns = convCollateralTxIns apiCollateralTxIns
+        refTxIns = convReferenceInputs apiReferenceInputs
+        returnCollateral = convReturnCollateral sbe apiReturnCollateral
+        totalCollateral = convTotalCollateral apiTotalCollateral
+        certs = convCertificates sbe $ txCertificates txBodyContent
+        txAuxData = toAuxiliaryData sbe (txMetadata txBodyContent) (txAuxScripts txBodyContent)
+        scripts = convScripts apiScriptWitnesses
+        languages = convLanguages apiScriptWitnesses
 
-      setTxBodyFields txBody = txBody
-        & L.certsTxBodyL                .~ certs
-        & A.invalidHereAfterTxBodyL sbe .~ convValidityUpperBound sbe (txValidityUpperBound txBodyContent)
+        setTxBodyFields txBody = txBody
+          & L.certsTxBodyL                .~ certs
+          & A.invalidHereAfterTxBodyL sbe .~ convValidityUpperBound sbe (txValidityUpperBound txBodyContent)
 
-      mkTxBody :: ()
-        => ShelleyBasedEra era
-        -> TxBodyContent BuildTx era
-        -> Maybe (L.TxAuxData (ShelleyLedgerEra era))
-        -> L.TxBody (ShelleyLedgerEra era)
-      mkTxBody sbe' bc =
-        mkCommonTxBody
-          sbe'
-          (txIns bc)
-          (txOuts bc)
-          (txFee bc)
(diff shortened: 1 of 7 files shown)


cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -1794,178 +1799,135 @@ instance Error TxBodyError where
...
-        -> Maybe (L.TxAuxData (ShelleyLedgerEra era))
-        -> L.TxBody (ShelleyLedgerEra era)
-      mkTxBody sbe' bc =
-        mkCommonTxBody
-          sbe'
-          (txIns bc)
-          (txOuts bc)
-          (txFee bc)
-          (txWithdrawals bc)
+createTransactionBody sbe bc =
```
newhoggy, line 1806: The focus of this PR is to simplify this function.

The earlier version of this code case matches on `sbe` and handles each era therein separately resulting in a lot of duplicate code.

The new version of this function instead eliminates the large case expression and instead uses eons allow each feature to determine for themselves if they need to set their values into the TxBody.

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -1795,177 +1800,115 @@ instance Error TxBodyError where
...
+        txAuxData = toAuxiliaryData sbe (txMetadata bc) (txAuxScripts bc)
+        scripts = convScripts apiScriptWitnesses
+        languages = convLanguages apiScriptWitnesses
+        sData = convScriptData sbe apiTxOuts apiScriptWitnesses
+
+    setUpdateProposal <-
+      caseShelleyToBabbageOrConwayEraOnwards
+        (\w -> (L.apiUpdateTxBodyL w .~) <$> convTxUpdateProposal sbe (txUpdateProposal bc))
+        (const $ pure id)
+        sbe
```
newhoggy, line 1834: With this refactoring we can deal with each tx body feature in isolation without duplication.

This can be simplified further in a future PR.

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -1795,177 +1800,115 @@ instance Error TxBodyError where
...
+            & L.invalidHereAfterTxBodyL sbe .~ convValidityUpperBound sbe (txValidityUpperBound bc)
+            & modifyWith setUpdateProposal
+            & modifyWith setInvalidBefore
+            & modifyWith setMint
+            & modifyWith setScriptIntegrityHash
+            & modifyWith setCollateralInputs
+            & modifyWith setReqSignerHashes
+            & modifyWith setReferenceInputs
+            & modifyWith setCollateralReturn
+            & modifyWith setTotalCollateral
```
newhoggy, line 1896: `modityWith` shouldn't be necessary but `ghc`'s type inference isn't quite there.

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -3025,27 +2933,28 @@ makeShelleyTransactionBody sbe@ShelleyBasedEraBabbage
                              txMintValue,
                              txScriptValidity
                            } = do
-
+    let aOn = AllegraEraOnwardsBabbage
```
newhoggy, line 2936: This is just temporary because we need a witness.  A future PR will get rid of this.

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -3025,27 +2933,28 @@ makeShelleyTransactionBody sbe@ShelleyBasedEraBabbage
...
-           & L.updateTxBodyL              .~ update
-           & L.reqSignerHashesTxBodyL     .~ convExtraKeyWitnesses txExtraKeyWits
-           & L.mintTxBodyL                .~ convMintValue txMintValue
-           & L.scriptIntegrityHashTxBodyL .~ scriptIntegrityHash
+            & L.referenceInputsTxBodyL      .~ convReferenceInputs txInsReference
+            & L.collateralReturnTxBodyL     .~ convReturnCollateral sbe txReturnCollateral
+            & L.totalCollateralTxBodyL      .~ convTotalCollateral txTotalCollateral
+            & L.certsTxBodyL                .~ convCertificates sbe txCertificates
+            & L.invalidBeforeTxBodyL aOn    .~ convValidityLowerBound txValidityLowerBound
+            & L.invalidHereAfterTxBodyL sbe .~ convValidityUpperBound sbe txValidityUpperBound
```
newhoggy, line 2953: These lenses take witnesses instead of constraints.  This can be used in future to simplify code.

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -1795,177 +1800,115 @@ instance Error TxBodyError where
...
+        sData = convScriptData sbe apiTxOuts apiScriptWitnesses
+
+    setUpdateProposal <-
+      caseShelleyToBabbageOrConwayEraOnwards
+        (\w -> (L.apiUpdateTxBodyL w .~) <$> convTxUpdateProposal sbe (txUpdateProposal bc))
+        (const $ pure id)
+        sbe
 
-  in case sbe of
-       ShelleyBasedEraShelley -> do
```
newhoggy, line 1838: One of the goals of this PR is to eliminate to need to case match on the era and handle each era separately.  Doing away with this case expression vastly reduces code duplication and the potential for error.

cardano-api/internal/Cardano/Api/Utils.hs
```diff
@@ -133,3 +133,10 @@ bounded t = eitherReader $ \s -> do
   when (i < fromIntegral (minBound @a)) $ Left $ t <> " must not be less than " <> show (minBound @a)
   when (i > fromIntegral (maxBound @a)) $ Left $ t <> " must not greater than " <> show (maxBound @a)
   pure (fromIntegral i)
+
+-- | Aids type inference.  Use this function to ensure the value is a function
```
carbolymer, line 137: I think I don't understand that. There's no modification happening in this function, so the comment and naming is misleading.

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -1794,178 +1799,116 @@ instance Error TxBodyError where
...
+            & L.certsTxBodyL                .~ certs
+            & A.invalidHereAfterTxBodyL sbe .~ convValidityUpperBound sbe (txValidityUpperBound bc)
+            & modifyWith setUpdateProposal
+            & modifyWith setInvalidBefore
+            & modifyWith setMint
+            & modifyWith setScriptIntegrityHash
+            & modifyWith setCollateralInputs
+            & modifyWith setReqSignerHashes
+            & modifyWith setReferenceInputs
+            & modifyWith setCollateralReturn
```
carbolymer, line 1895: Invocations of `modifyWith` are not changing the value, so the before and after versions of the code have different meaning.

carbolymer (changes requested):
https://github.com/input-output-hk/cardano-api/pull/333/files#r1368451851

cardano-api/internal/Cardano/Api/Utils.hs
newhoggy: The function doesn't do anything and should behave the same as `id`.  The only reason it is used at all is to help with type-inference.  You can remove the function calls and observe that type-inference fails.

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -1798,184 +1799,116 @@ instance Error TxBodyError where
...
-              ledgerTxBody
-              scripts
-              (convScriptData sbe apiTxOuts apiScriptWitnesses)
-              txAuxData
-              apiScriptValidity
+    setMint <-
+      caseShelleyToAllegraOrMaryEraOnwards
+        (const $ pure id)
+        (const $ pure $ L.mintTxBodyL .~ convMintValue apiMintValue)
+        sbe
```
carbolymer, line 1846: Why do you need `>>=` here? Why not:
```suggestion
let setMint =
      caseShelleyToAllegraOrMaryEraOnwards
        (const $ id)
        (const $ L.mintTxBodyL .~ convMintValue apiMintValue)
        sbe
```

carbolymer (approved):

cardano-api/internal/Cardano/Api/TxBody.hs
newhoggy: This is mostly to make all of the units of code have the same shape rather than have some unit use `let` and others use `<-`.

You are right in that the `>>=` is not needed.

PR #350, opened 2023-10-31
foldBlocks update

Jimbo4350:
# Changelog

```yaml
- description: |
    Modify foldBlocks to allow the fold to terminate from the accumulator via the FoldStatus type.
    
    Modify foldBlocks to accumulate the chain tip rather than only immutable
    blocks (blocks that are k blocks away from the tip).
    
    Add debug mode to foldBlocks which forces it to error with information 
    about ledger states, client and server tip, number of requests in flight
    and the current IORef state.
  type:
```

# Context

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/LedgerState.hs
+++ b/cardano-api/internal/Cardano/Api/LedgerState.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE FlexibleContexts #-}
@@ -166,6 +167,7 @@ import qualified Ouroboros.Network.Protocol.ChainSync.Client as CS
 import qualified Ouroboros.Network.Protocol.ChainSync.ClientPipelined as CSP
 import           Ouroboros.Network.Protocol.ChainSync.PipelineDecision
 
+import           Control.DeepSeq
 import           Control.Error.Util (note)
 import           Control.Exception
 import           Control.Monad
@@ -240,7 +242,7 @@ data LedgerStateError
   -- ^ Encountered a rollback larger than the security parameter.
       SlotNo     -- ^ Oldest known slot number that we can roll back to.
       ChainPoint -- ^ Rollback was attempted to this point.
-  | DebugError String
+  | DebugError !String
   deriving (Show)
 
 instance Exception LedgerStateError
@@ -532,17 +534,17 @@ foldBlocks nodeConfigFilePath socketPath validationMode state0 accumulate = do
                           currentIORefState <- readIORef stateIORef
 
                           -- Useful for debugging:
-                          let ioRefErr = Just . DebugError
-                                              $ unlines [ "newClientTip: " <> show newClientTip
-                                                        , "newServerTip: " <> show newServerTip
-                                                        , "newLedgerState: " <> show (snd newLedgerState)
-                                                        , "knownLedgerStates: " <> show (extractHistory knownLedgerStates)
-                                                        , "committedStates: " <> show (extractHistory committedStates)
-                                                        , "numberOfRequestsInFlight: " <> show n
-                                                        , "k: " <> show (envSecurityParam env)
-                                                        , "Current IORef State: " <> show currentIORefState
-                                                        ]
-                          clientIdle_DoneNwithMaybeError n ioRefErr
+                          let !ioRefErr = DebugError . force
+                                            $ unlines [ "newClientTip: " <> show newClientTip
+                                                      , "newServerTip: " <> show newServerTip
+                                                      , "newLedgerState: " <> show (snd newLedgerState)
+                                                      , "knownLedgerStates: " <> show (extractHistory knownLedgerStates)
+                                                      , "committedStates: " <> show (extractHistory committedStates)
+                                                      , "numberOfRequestsInFlight: " <> show n
+                                                      , "k: " <> show (envSecurityParam env)
+                                                      , "Current IORef State: " <> show currentIORefState
+                                                      ]
+                          clientIdle_DoneNwithMaybeError n $ Just ioRefErr
 
                         ContinueFold -> return $ clientIdle_RequestMoreN newClientTip newServerTip n knownLedgerStates'
 
--- a/cardano-api/internal/Cardano/Api/LedgerState.hs
+++ b/cardano-api/internal/Cardano/Api/LedgerState.hs
@@ -34,6 +34,7 @@ module Cardano.Api.LedgerState
 
     -- * Traversing the block chain
   , foldBlocks
+  , FoldStatus(..)
   , chainSyncClientWithLedgerState
   , chainSyncClientPipelinedWithLedgerState
 
@@ -80,7 +81,6 @@ module Cardano.Api.LedgerState
 
   )
   where
-
 import           Cardano.Api.Block
 import           Cardano.Api.Certificate
 import           Cardano.Api.Eon.ShelleyBasedEra
@@ -160,6 +160,7 @@ import qualified Ouroboros.Consensus.Shelley.Eras as Shelley
 import qualified Ouroboros.Consensus.Shelley.Ledger.Block as Shelley
 import qualified Ouroboros.Consensus.Shelley.Ledger.Ledger as Shelley
 import           Ouroboros.Consensus.TypeFamilyWrappers (WrapLedgerEvent (WrapLedgerEvent))
+import           Ouroboros.Network.Block (blockNo)
 import qualified Ouroboros.Network.Block
 import qualified Ouroboros.Network.Protocol.ChainSync.Client as CS
 import qualified Ouroboros.Network.Protocol.ChainSync.ClientPipelined as CSP
@@ -167,7 +168,7 @@ import           Ouroboros.Network.Protocol.ChainSync.PipelineDecision
 
 import           Control.Error.Util (note)
 import           Control.Exception
-import           Control.Monad (when)
+import           Control.Monad
 import           Control.Monad.Trans.Class
 import           Control.Monad.Trans.Except
 import           Control.Monad.Trans.Except.Extra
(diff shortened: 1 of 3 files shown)


cardano-api/internal/Cardano/Api/LedgerState.hs
```diff
@@ -344,18 +348,27 @@ renderFoldBlocksError fbe = case fbe of
   FoldBlocksInitialLedgerStateError err -> renderInitialLedgerStateError err
   FoldBlocksApplyBlockError err -> "Failed when applying a block: " <> renderLedgerStateError err
 
+-- Type that lets us decide whether to continue or stop
```
carbolymer, line 351:
```suggestion
-- | Type that lets us decide whether to continue or stop
```

cardano-api/internal/Cardano/Api/LedgerState.hs
```diff
@@ -485,24 +499,50 @@ foldBlocks nodeConfigFilePath socketPath validationMode state0 accumulate = do
...
+                          -- and then terminate e.g a specific stake pool was registered
+                          let noError = Nothing
+                          in clientIdle_DoneNwithMaybeError n noError
+
+                        DebugFold -> do
+                          currentIORefState <- readIORef stateIORef
+
+                          -- Useful for debugging:
+                          let ioRefErr = Just $ DebugError $ unlines [ "newClientTip: " <> show newClientTip
+                                                                           , "newServerTip: " <> show newServerTip
```
carbolymer, line 534: this indentation looks off in github

cardano-api/internal/Cardano/Api/LedgerState.hs
```diff
@@ -485,24 +499,50 @@ foldBlocks nodeConfigFilePath socketPath validationMode state0 accumulate = do
...
+                    Left err -> clientIdle_DoneNwithMaybeError n (Just err)
                     Right newLedgerState -> do
                       let (knownLedgerStates', committedStates) = pushLedgerState env knownLedgerStates slotNo newLedgerState blockInMode
                           newClientTip = At currBlockNo
                           newServerTip = fromChainTip serverChainTip
-                      forM_ committedStates $ \(_, (ledgerState, ledgerEvents), currBlockMay) -> case currBlockMay of
-                          Origin -> return ()
+                      -- TODO: We are constantly overwriting an IORef which isn't ideal.
+                      -- We should be writing to a TChan so that we don't have to return the
+                      -- concatenation of all the ledger events in the accumulation function.
```
carbolymer, line 509: This comment isn't super clear to me. `TChan` can't be used as a drop-in replacement for `stateIORef :: IORef a` here, as how exactly accumulation is done is dictated by `accumulate` function. This means, that `newState` returned by `accumulate` doesn't have to be a FIFO stream (which is a `TChan` is), but can have any shape.
Can you clarify what do you mean here?

cardano-api/internal/Cardano/Api/LedgerState.hs
```diff
@@ -485,24 +499,50 @@ foldBlocks nodeConfigFilePath socketPath validationMode state0 accumulate = do
...
+                      foldStatuses <- forM knownLedgerStates' $ \(_, (ledgerState, ledgerEvents), currBlockMay) -> case currBlockMay of
```
carbolymer, line 510: If we're only interested in the first occurrence of a `FoldStatus` (this is what `foldDecision` does), this should be a set instead of a sequence. This way we would avoid accumulating all the unnecessary `foldStatus`es. `foldM` can be used here.

cardano-api/internal/Cardano/Api/LedgerState.hs
```diff
@@ -485,24 +499,50 @@ foldBlocks nodeConfigFilePath socketPath validationMode state0 accumulate = do
...
+                          let ioRefErr = Just $ DebugError $ unlines [ "newClientTip: " <> show newClientTip
+                                                                           , "newServerTip: " <> show newServerTip
+                                                                           , "newLedgerState: " <> show (snd newLedgerState)
+                                                                           , "knownLedgerStates: " <> show (extractHistory knownLedgerStates)
+                                                                           , "committedStates: " <> show (extractHistory committedStates)
+                                                                           , "numberOfRequestsInFlight: " <> show n
+                                                                           , "k: " <> show (envSecurityParam env)
+                                                                           , "Current IORef State: " <> show currentIORefState
+                                                                           ]
+                          clientIdle_DoneNwithMaybeError n ioRefErr
```
carbolymer, line 542: I'm not sure how soon this gets logged/written somewhere, but it may be worth enforcing strictness here
```suggestion
                          deepseq ioRefErr $ clientIdle_DoneNwithMaybeError n ioRefErr
```

carbolymer (changes requested):
https://github.com/input-output-hk/cardano-api/pull/350/files#r1379986767

cardano-api/internal/Cardano/Api/LedgerState.hs
Jimbo4350: This is left over from my debugging. I'll remove it.

cardano-api/internal/Cardano/Api/LedgerState.hs
carbolymer: Ok. It's that the current implementation does only check for the presence of an element in the list.

cardano-api/internal/Cardano/Api/LedgerState.hs
Jimbo4350: We discussed on slack but I will remove the comment regarding the `TChan`. I'll make a PR in the near future to address this.

cardano-api/internal/Cardano/Api/LedgerState.hs
Jimbo4350: The plan is to recurse on `knownLedgerStates'` instead and exit (or continue) depending on the value of `FoldStatus`. This will be in a follow up PR so I could keep the diff small here.

cardano-api/internal/Cardano/Api/LedgerState.hs
Jimbo4350: Fixed :+1:

cardano-api/internal/Cardano/Api/LedgerState.hs
Jimbo4350: So this would be ideal but I would need to derive `NFData` instances for types in consensus that currently don't exist.

cardano-api/internal/Cardano/Api/LedgerState.hs
```diff
@@ -238,13 +240,15 @@ data LedgerStateError
   -- ^ Encountered a rollback larger than the security parameter.
       SlotNo     -- ^ Oldest known slot number that we can roll back to.
       ChainPoint -- ^ Rollback was attempted to this point.
+  | DebugError String
```
carbolymer, line 243:
```suggestion
  | DebugError !String
```

cardano-api/internal/Cardano/Api/LedgerState.hs
Jimbo4350: https://github.com/input-output-hk/cardano-api/pull/353/commits/f73699815056a29e3c6a7a9f33079e1df96bf1ec

cardano-api/internal/Cardano/Api/LedgerState.hs
```diff
@@ -488,24 +502,50 @@ foldBlocks nodeConfigFilePath socketPath validationMode state0 accumulate = do
...
+                                              $ unlines [ "newClientTip: " <> show newClientTip
+                                                        , "newServerTip: " <> show newServerTip
+                                                        , "newLedgerState: " <> show (snd newLedgerState)
+                                                        , "knownLedgerStates: " <> show (extractHistory knownLedgerStates)
+                                                        , "committedStates: " <> show (extractHistory committedStates)
+                                                        , "numberOfRequestsInFlight: " <> show n
+                                                        , "k: " <> show (envSecurityParam env)
+                                                        , "Current IORef State: " <> show currentIORefState
+                                                        ]
+                          clientIdle_DoneNwithMaybeError n ioRefErr
```
carbolymer, line 545:
```suggestion
                          let !ioRefErr = DebugError
                                              $ unlines [ "newClientTip: " <> show newClientTip
                                                        , "newServerTip: " <> show newServerTip
                                                        , "newLedgerState: " <> show (snd newLedgerState)
                                                        , "knownLedgerStates: " <> show (extractHistory knownLedgerStates)
                                                        , "committedStates: " <> show (extractHistory committedStates)
                                                        , "numberOfRequestsInFlight: " <> show n
                                                        , "k: " <> show (envSecurityParam env)
                                                        , "Current IORef State: " <> show currentIORefState
                                                        ]
                          clientIdle_DoneNwithMaybeError n (Just ioRefErr)
```

cardano-api/internal/Cardano/Api/LedgerState.hs
carbolymer: Right. I guess we can get around using NFData with `BangPatterns`:
1. Make `DebugError` constructor field strict
2. use `let !ioRefError = DebugError`
3. wrap `ioRefErr` in `Just` when calling `clientIdle_DoneNwithMaybeError`, because `Maybe` is lazy

cardano-api/internal/Cardano/Api/LedgerState.hs
```diff
@@ -488,24 +503,50 @@ foldBlocks nodeConfigFilePath socketPath validationMode state0 accumulate = do
...
+                          -- This allow us to check for a specific condition in our accumulate function
+                          -- and then terminate e.g a specific stake pool was registered
+                          let noError = Nothing
+                          in clientIdle_DoneNwithMaybeError n noError
+
+                        DebugFold -> do
+                          currentIORefState <- readIORef stateIORef
+
+                          -- Useful for debugging:
+                          let ioRefErr = DebugError . force
```
carbolymer, line 536:
```suggestion
                          let !ioRefErr = DebugError . force
```
otherwise `ioError` will still be a thunk here, until we evaluate it somewhere

carbolymer (approved):

PR #353, opened 2023-11-03
Modify foldBlocks to recurse on ledger events 

Jimbo4350:
# Changelog

```yaml
- description: |
    Modify foldBlocks to recurse on ledger events instead of mapping over them. This allows finer grained control over when `foldBlocks` is stopped.
  type:
  - optimisation   # measurable performance improvements
```

# Context

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/LedgerState.hs
+++ b/cardano-api/internal/Cardano/Api/LedgerState.hs
@@ -509,20 +509,39 @@ foldBlocks nodeConfigFilePath socketPath validationMode state0 accumulate = do
                       let (knownLedgerStates', committedStates) = pushLedgerState env knownLedgerStates slotNo newLedgerState blockInMode
                           newClientTip = At currBlockNo
                           newServerTip = fromChainTip serverChainTip
-                      -- TODO: We are constantly overwriting an IORef which isn't ideal.
 
-                      foldStatuses <- forM knownLedgerStates' $ \(_, (ledgerState, ledgerEvents), currBlockMay) -> case currBlockMay of
-                          Origin -> pure ContinueFold
-                          At currBlock -> do
+                          ledgerStateSingleFold
+                            :: (SlotNo, (LedgerState, [LedgerEvent]), WithOrigin BlockInMode) -- Ledger events for a single block
+                            -> IO FoldStatus
+                          ledgerStateSingleFold (_, _, Origin) = return ContinueFold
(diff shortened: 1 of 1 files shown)


cardano-api/internal/Cardano/Api/LedgerState.hs
```diff
@@ -488,24 +502,69 @@ foldBlocks nodeConfigFilePath socketPath validationMode state0 accumulate = do
...
+                          ledgerStateSingleFold (_, _, Origin) = return ContinueFold
+                          ledgerStateSingleFold (_, (ledgerState, ledgerEvents), At currBlock) = do
+                            accumulatorState <- readIORef stateIORef
+                            (newState, foldStatus) <- accumulate
+                                                      env
+                                                      ledgerState
+                                                      ledgerEvents
+                                                      currBlock
+                                                      accumulatorState
+                            atomicWriteIORef stateIORef newState
```
carbolymer, line 523: It feels that we should use an underlying monad in `ClientStNext` to store state instead of using IORef. 
```
CSP.ClientStNext n BlockInMode ChainPoint ChainTip (StateT s IO) ()
```
would be cleaner I think.

carbolymer (approved):

cardano-api/internal/Cardano/Api/LedgerState.hs
Jimbo4350: Can you open a PR after this one? I'd be interested to see what you have in mind.

carbolymer: I took a closer look and it does not seem to be worth the effort. My point was to be able to use state monad in the protocol client to store the state (including an error). The problem with that is the change is quite extensive: we'd have to parameterize types used in the client e.g. `LocalNodeClientProtocolsForBlock` by an underlying monad additionally.
Moreover, the code in consensus we use has IO hardcoded there in the function argument e.g.:
```haskell
connectTo :: forall a b.
LocalSnocket
-> NetworkConnectTracers LocalAddress NodeToClientVersion
-> Versions
     NodeToClientVersion
     NodeToClientVersionData
     (OuroborosApplicationWithMinimalCtx
        'InitiatorMode LocalAddress ByteString IO a b)
-> FilePath
-> IO ()
```
This would require unlifting state into IO and then lifting it again, which is cumbersome.

In the end the semantics would be pretty much the same, but with more boilerplate.

PR #356, opened 2023-11-06
Delete `ByronToAllegraEra` eon

newhoggy:
# Changelog

```yaml
- description: |
    New `caseByronOrShelleyToAllegraOrMaryEraOnwards` function
    New `TxOutAdaOnlyByron` constructor
    Modify `TxOutAdaOnly` constructor to take `ShelleyToAllegraEra` instead of `ByronToAllegraEra`
    Delete `ByronToAllegraEra` eon
    Delete:
    - `caseByronToAllegraOrMaryEraOnwards`
    - `shelleyToAllegraEraToByronToAllegraEra`
    - `byronToAllegraEraConstraints`
    - `byronToAllegraEraToCardanoEra`
  type:
```

# Context

We desire to avoid eons that straddle the Byron/ShelleyBasedEra boundary.  Byron is very legacy so it ought to be a bit special and kept separate.

The reason we needed `ByronToAllegraEra` eon was because we had code that wanted to treat Byron era and subsequent eras the same way.  If we give Byron era it's own constructor, the need goes away.

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff


cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -886,7 +886,9 @@ deriving instance Show (TxInsReference build era)
 
 data TxOutValue era where
 
-  TxOutAdaOnly :: ByronToAllegraEra era -> Lovelace -> TxOutValue era
```
newhoggy, line 889: We split this constructor to avoid needing an extra eon.

carbolymer: Not sure that's worth it. We're removing one eon, but the additional constructor is not handling any special behaviour and makes code a bit confusing.

carbolymer (commented):
https://github.com/input-output-hk/cardano-api/pull/356/files#r1383315760
I'm not sure if this change is beneficial.

cardano-api/internal/Cardano/Api/TxBody.hs
newhoggy: Yeah, there's a tradeoff.

Jimbo4350 (changes requested):
Let's refrain from more changes to the api concerning eons. We need to decide a way forward.

newhoggy:
Closing for the moment.  Revisit later.

PR #360, opened 2023-11-08
Use ledger presentation of multi-asset values directly.  Lens to make this uniform over `ShelleyBasedEra`

newhoggy:
# Changelog

```yaml
- description: |
    Use ledger presentation of multi-asset values directly.  Lens to make this uniform over `ShelleyBasedEra`.
    
    Delete `ByronToAllegraEra`.
    
    New module `Cardano.Api.Ledger.Lens`.
    
    Modify `TxOutValue` to have `TxOutValueByron` and `TxOutValueShelleyBased` instead of `TxOutAdaOnly` and `TxOutValue` respectively.  `TxOutValueShelleyBased` now directly uses the ledger type instead of the `Value` type. 
    
    These functions have changed to either `L.Value (ShelleyLedgerEra era)` instead of `Value` or eons or both:
    - `genValue`
    - `genValueDefault`
    - `genValueForMinting`
  type:
```

# Context

This PR introduces adapter lenses `adaAssetL` and `multiAssetL`.

The former provides a uniform way to access the ADA asset in ledger values uniformly across all of `ShelleyBasedEra` which simplifies code considerably.

The latter provides an easy way to access all the non-ADA assets from Mary onwards.

The change also deletes `ByronToAllegraEra`

Further simplifications are possible but are left to future PRs to keep this PR small.

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff


cardano-api/internal/Cardano/Api/Eon/ByronToAllegraEra.hs
```diff
@@ -1,54 +0,0 @@
-{-# LANGUAGE ConstraintKinds #-}
-{-# LANGUAGE DataKinds #-}
-{-# LANGUAGE FlexibleContexts #-}
-{-# LANGUAGE GADTs #-}
-{-# LANGUAGE LambdaCase #-}
-{-# LANGUAGE RankNTypes #-}
-{-# LANGUAGE StandaloneDeriving #-}
-{-# LANGUAGE TypeFamilies #-}
-
-module Cardano.Api.Eon.ByronToAllegraEra
```
newhoggy, line 10: This PR deletes `ByronToAllegraEra`

cardano-api/internal/Cardano/Api/Eras/Case.hs
```diff
@@ -60,22 +57,6 @@ caseByronOrShelleyBasedEra l r = \case
...
-  -> CardanoEra era
-  -> a
-caseByronToAllegraOrMaryEraOnwards l r = \case
-  ByronEra   -> l ByronToAllegraEraByron
-  ShelleyEra -> l ByronToAllegraEraShelley
-  AllegraEra -> l ByronToAllegraEraAllegra
-  MaryEra    -> r MaryEraOnwardsMary
-  AlonzoEra  -> r MaryEraOnwardsAlonzo
-  BabbageEra -> r MaryEraOnwardsBabbage
-  ConwayEra  -> r MaryEraOnwardsConway
```
newhoggy, line 77: Deletes a case function because we don't need it anymore.

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -680,70 +678,58 @@ fromByronTxOut :: ByronEraOnly era -> Byron.TxOut -> TxOut ctx era
...
+        L.mkBasicTxOut (toShelleyAddr addr) value
+          & L.dataHashTxOutL .~ toAlonzoTxOutDataHash txoutdata
+      ShelleyBasedEraBabbage ->
+        L.mkBasicTxOut (toShelleyAddr addr) value
+          & L.datumTxOutL .~ toBabbageTxOutDatum txoutdata
+          & L.referenceScriptTxOutL .~ refScriptToShelleyScript cEra refScript
+      ShelleyBasedEraConway ->
+        L.mkBasicTxOut (toShelleyAddr addr) value
+          & L.datumTxOutL .~ toBabbageTxOutDatum txoutdata
+          & L.referenceScriptTxOutL .~ refScriptToShelleyScript cEra refScript
```
newhoggy, line 723: The `L.mkBasicTxOut (toShelleyAddr addr) value` is repeated `6` times.  Potential to simplify.

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -886,29 +872,50 @@ deriving instance Show (TxInsReference build era)
...
+    -> TxOutValue era
 
-  TxOutValue   :: MaryEraOnwards era -> Value -> TxOutValue era
+  TxOutValueShelleyBased
+    ::  ( Eq (Ledger.Value (ShelleyLedgerEra era))
+        , Show (Ledger.Value (ShelleyLedgerEra era))
+        )
+    => ShelleyBasedEra era
+    -> L.Value (ShelleyLedgerEra era)
+    -> TxOutValue era
```
newhoggy, line 886: This definition has two constructors like before, the main difference is that all Shelley eras are now treated the same and only Byron is different.  This is as it should be because the ledger treats Byron as legacy so it doesn't even have lens support.

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -946,23 +953,23 @@ lovelaceToTxOutValue :: ()
...
+    (\w -> TxOutValueShelleyBased w $ A.mkAdaValue w $ lovelaceToCoin ll)
     era
 
 txOutValueToLovelace :: TxOutValue era -> Lovelace
 txOutValueToLovelace tv =
   case tv of
-    TxOutAdaOnly _ l -> l
-    TxOutValue _ v -> selectLovelace v
+    TxOutValueByron         _   l -> l
+    TxOutValueShelleyBased  sbe v -> coinToLovelace $ v ^. A.adaAssetL sbe
```
newhoggy, line 966: Example of where we use `adaAssetL` to access the ADA value.

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -3141,33 +3147,36 @@ toShelleyTxOutAny :: forall ctx era ledgerera.
...
-    let cEra = shelleyBasedToCardanoEra sbe
-    in L.mkBasicTxOut (toShelleyAddr addr) (toMaryValue value)
-       & L.datumTxOutL .~ toBabbageTxOutDatum' txoutdata
-       & L.referenceScriptTxOutL .~ refScriptToShelleyScript cEra refScript
+      ShelleyBasedEraConway ->
+        L.mkBasicTxOut (toShelleyAddr addr) value
+          & L.datumTxOutL .~ toBabbageTxOutDatum' txoutdata
+          & L.referenceScriptTxOutL .~ refScriptToShelleyScript cEra refScript
+  where
+    cEra = shelleyBasedToCardanoEra sbe
```
newhoggy, line 3179: Opportunity to simplify because `L.mkBasicTxOut (toShelleyAddr addr) value` is repeated.  This wasn't the case previously.

cardano-api/internal/Cardano/Api/Value.hs
```diff
@@ -254,6 +274,15 @@ selectLovelace = quantityToLovelace . flip selectAsset AdaAssetId
 lovelaceToValue :: Lovelace -> Value
 lovelaceToValue = Value . Map.singleton AdaAssetId . lovelaceToQuantity
 
+lovelaceToCoin :: Lovelace -> Shelley.Coin
+lovelaceToCoin (Lovelace ll) = Shelley.Coin ll
+
+coinToLovelace :: Shelley.Coin -> Lovelace
+coinToLovelace (Shelley.Coin ll) = Lovelace ll
```
newhoggy, line 281: We could delete `Lovelace` and use the `Coin` ledger type directly to avoid conversions.

cardano-api/internal/Cardano/Api/Value.hs
```diff
@@ -281,6 +310,15 @@ toMaryValue v =
...
 
+toLedgerValue :: MaryEraOnwards era -> Value -> L.Value (ShelleyLedgerEra era)
+toLedgerValue w = maryEraOnwardsConstraints w toMaryValue
+
+fromLedgerValue :: ShelleyBasedEra era -> L.Value (ShelleyLedgerEra era) -> Value
+fromLedgerValue sbe v =
+  caseShelleyToAllegraOrMaryEraOnwards
+    (const (coinToValue v))
+    (const (fromMaryValue v))
+    sbe
```
newhoggy, line 321: These conversions were functions were introduced to convert between `Value` and `L.Value ledgerera`.  `Value` could actually be deleted if we use the ledger type directly everywhere and then these conversion functions go away.

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -680,70 +678,58 @@ fromByronTxOut :: ByronEraOnly era -> Byron.TxOut -> TxOut ctx era
...
-      txOutValue =
-        caseShelleyToAllegraOrMaryEraOnwards
-          (\w -> TxOutAdaOnly (shelleyToAllegraEraToByronToAllegraEra w) (fromShelleyLovelace (ledgerTxOut ^. L.valueTxOutL)))
-          (\w -> TxOutValue w (fromMaryValue (ledgerTxOut ^. L.valueTxOutL)))
-          sbe
-
-  let addressInEra :: AddressInEra era
-      addressInEra = shelleyBasedEraConstraints sbe $ fromShelleyAddr sbe $ ledgerTxOut ^. L.addrTxOutL
+fromShelleyTxOut sbe ledgerTxOut = shelleyBasedEraConstraints sbe $ do
+  let txOutValue = TxOutValueShelleyBased sbe $ ledgerTxOut ^. A.valueTxOutL sbe
```
newhoggy, line 731: If we use the ledger types directly we no longer have to convert to `cardano-api`'s representation so this simplifies considerably.

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -3141,33 +3147,36 @@ toShelleyTxOutAny :: forall ctx era ledgerera.
...
-    L.mkBasicTxOut (toShelleyAddr addr) (toMaryValue value)
+      ShelleyBasedEraMary ->
+        L.mkBasicTxOut (toShelleyAddr addr) value
 
-toShelleyTxOutAny _ (TxOut addr (TxOutValue MaryEraOnwardsAlonzo value) txoutdata _) =
-    L.mkBasicTxOut (toShelleyAddr addr) (toMaryValue value)
-    & L.dataHashTxOutL .~ toAlonzoTxOutDataHash' txoutdata
+      ShelleyBasedEraAlonzo ->
+        L.mkBasicTxOut (toShelleyAddr addr) value
+          & L.dataHashTxOutL .~ toAlonzoTxOutDataHash' txoutdata
```
newhoggy, line 3167: `dataHashTxOutL` is only used in Alonzo?  Why?  Can we delete it?

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -680,70 +678,57 @@ fromByronTxOut :: ByronEraOnly era -> Byron.TxOut -> TxOut ctx era
...
-
-toShelleyTxOut _ (TxOut addr (TxOutAdaOnly ByronToAllegraEraAllegra value) _ _) =
-    L.mkBasicTxOut (toShelleyAddr addr) (toShelleyLovelace value)
-
-toShelleyTxOut _ (TxOut addr (TxOutValue MaryEraOnwardsMary value) _ _) =
-    L.mkBasicTxOut (toShelleyAddr addr) (toMaryValue value)
-
-toShelleyTxOut _ (TxOut addr (TxOutValue MaryEraOnwardsAlonzo value) txoutdata _) =
-    L.mkBasicTxOut (toShelleyAddr addr) (toMaryValue value)
-    & L.dataHashTxOutL .~ toAlonzoTxOutDataHash txoutdata
```
newhoggy, line 718: This appears to be Alonzo only.  Deleting this can simplify the code a lot.

cardano-api/cardano-api.cabal
```diff
@@ -223,6 +223,7 @@ library
                         -- exposed by cardano-api-ledger
                         Cardano.Api.Ledger
 
+  reexported-modules:   Cardano.Api.Ledger.Lens
```
newhoggy, line 226: New export so we can use eon enabled from the CLI and from tests.

cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -827,34 +809,28 @@ makeTransactionBodyAutoBalance sbe systemstart history lpp@(LedgerProtocolParame
...
-          , mintedNonAda
-          , negateValue outgoingNonAda
-          ]
-
-    let changeTxOut = caseByronToAllegraOrMaryEraOnwards
-          (const $ lovelaceToTxOutValue era $ Lovelace (2^(64 :: Integer)) - 1)
-          (\w -> TxOutValue w (lovelaceToValue (Lovelace (2^(64 :: Integer)) - 1) <> nonAdaChange))
-          era
+          TxMintValue w v _ -> toLedgerValue w v
+    let change = mconcat [incoming, minted, negateLedgerValue sbe outgoing]
```
newhoggy, line 820: In the above code this differs from the previous code in that we also add all the Ada/Lovelace as well, but that doesn't change the behaviour because with throw away that value anyway with the computation of `changeWithMaxLovelace`.

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -793,25 +763,6 @@ fromShelleyTxOut sbe ledgerTxOut = do
...
-toAlonzoTxOutDataHash  TxOutDatumNone                        = SNothing
-toAlonzoTxOutDataHash (TxOutDatumHash _ (ScriptDataHash dh)) = SJust dh
-toAlonzoTxOutDataHash (TxOutDatumInline inlineDatumSupp _sd) =
-  case inlineDatumSupp :: BabbageEraOnwards AlonzoEra of {}
-
-fromAlonzoTxOutDataHash :: AlonzoEraOnwards era
-                        -> StrictMaybe (L.DataHash StandardCrypto)
-                        -> TxOutDatum ctx era
-fromAlonzoTxOutDataHash _    SNothing  = TxOutDatumNone
-fromAlonzoTxOutDataHash s (SJust dh)   = TxOutDatumHash s (ScriptDataHash dh)
```
newhoggy, line 811: A bunch of alonzo specific code disappears because we remove the feature that only worked in Alonzo.

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -2195,27 +2165,17 @@ fromLedgerTxOuts sbe body scriptdata =
...
-fromAlonzoTxOutDatum :: ()
-  => AlonzoEraOnwards era
-  -> Map (L.DataHash StandardCrypto) (L.Data ledgerera)
-  -> StrictMaybe (L.DataHash StandardCrypto)
-  -> TxOutDatum CtxTx era
-fromAlonzoTxOutDatum w txdatums = \case
-  SNothing -> TxOutDatumNone
-  SJust dh
-    | Just d <- Map.lookup dh txdatums  -> TxOutDatumInTx' w (ScriptDataHash dh) (fromAlonzoData d)
-    | otherwise                         -> TxOutDatumHash w (ScriptDataHash dh)
```
newhoggy, line 2218: A bunch of alonzo specific code disappears because we remove the feature that only worked in Alonzo and choose to no longer support.

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -3141,41 +3098,20 @@ toShelleyTxOutAny :: forall ctx era ledgerera.
...
-    L.mkBasicTxOut (toShelleyAddr addr) (toMaryValue value)
-
-toShelleyTxOutAny _ (TxOut addr (TxOutValue MaryEraOnwardsAlonzo value) txoutdata _) =
-    L.mkBasicTxOut (toShelleyAddr addr) (toMaryValue value)
-    & L.dataHashTxOutL .~ toAlonzoTxOutDataHash' txoutdata
-
-toShelleyTxOutAny sbe (TxOut addr (TxOutValue MaryEraOnwardsBabbage value) txoutdata refScript) =
-    let cEra = shelleyBasedToCardanoEra sbe
-    in L.mkBasicTxOut (toShelleyAddr addr) (toMaryValue value)
-       & L.datumTxOutL .~ toBabbageTxOutDatum' txoutdata
```
newhoggy, line 3163: A bunch of alonzo specific code disappears because we remove the feature that only worked in Alonzo and choose to no longer support.

cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -591,30 +596,6 @@ evaluateTransactionBalance sbe pp poolids stakeDelegDeposits drepDelegDeposits u
...
-    evalAdaOnly w =
-      shelleyToAllegraEraConstraints w
-        $ TxOutAdaOnly (shelleyToAllegraEraToByronToAllegraEra w) . fromShelleyLovelace
-        $ L.evalBalanceTxBody
-            pp
-            lookupDelegDeposit
-            lookupDRepDeposit
-            isRegPool
-            (toLedgerUTxO sbe utxo)
-            txbody
```
newhoggy, line 616: Because the lens adapter makes the ADA asset interface consistent across all of `ShelleyBasedEra` we no longer need to case on the eon to implement `evaluateTransactionBalance`.

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -680,70 +678,47 @@ fromByronTxOut :: ByronEraOnly era -> Byron.TxOut -> TxOut ctx era
...
-fromShelleyTxOut sbe ledgerTxOut = do
-  let txOutValue :: TxOutValue era
-      txOutValue =
-        caseShelleyToAllegraOrMaryEraOnwards
-          (\w -> TxOutAdaOnly (shelleyToAllegraEraToByronToAllegraEra w) (fromShelleyLovelace (ledgerTxOut ^. L.valueTxOutL)))
-          (\w -> TxOutValue w (fromMaryValue (ledgerTxOut ^. L.valueTxOutL)))
-          sbe
-
-  let addressInEra :: AddressInEra era
-      addressInEra = shelleyBasedEraConstraints sbe $ fromShelleyAddr sbe $ ledgerTxOut ^. L.addrTxOutL
```
newhoggy, line 746: More simplifications because the eon-enabled lens adapter makes ADA Asset appear uniform across all of `ShelleyBasedEra`.

cardano-api/cardano-api.cabal
```diff
@@ -170,6 +169,7 @@ library internal
                       , errors
                       , filepath
                       , formatting
+                      , groups
```
newhoggy, line 172: We import `groups` because ledger implements `Group` for multi-asset values which we can use to negative the entire `L.Value ledgerera`.

carbolymer (approved):

cardano-api/internal/Cardano/Api/Ledger/Lens.hs
```diff
@@ -165,3 +177,46 @@ votingProceduresTxBodyL w = conwayEraOnwardsConstraints w $ txBodyL . L.votingPr
...
+    (const coin)
+    (const (L.MaryValue (L.unCoin coin) mempty))
+    sbe
+
+adaAssetL :: ShelleyBasedEra era -> Lens' (L.Value (ShelleyLedgerEra era)) L.Coin
+adaAssetL sbe =
+  caseShelleyToAllegraOrMaryEraOnwards
+    adaAssetShelleyToAllegraEraL
+    adaAssetMaryEraOnwardsL
+    sbe
```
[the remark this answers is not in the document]
Jimbo4350: We need to move this lens to `cardano-ledger` in a follow up PR.

cardano-api/cardano-api.cabal
Jimbo4350: Why not just expose it?

cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -571,10 +570,15 @@ evaluateTransactionBalance sbe _ _ _ _ _ (ByronTxBody ByronEraOnlyByron _) =
   case sbe of {}
 
 evaluateTransactionBalance sbe pp poolids stakeDelegDeposits drepDelegDeposits utxo (ShelleyTxBody _ txbody _ _ _ _) =
-    caseShelleyToAllegraOrMaryEraOnwards
-      evalAdaOnly
-      evalMultiAsset
-      sbe
+  shelleyBasedEraConstraints sbe
+    $ TxOutValueShelleyBased sbe
```
Jimbo4350, line 575: :+1:

cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -828,21 +809,22 @@ makeTransactionBodyAutoBalance sbe systemstart history lpp@(LedgerProtocolParame
     -- However, since at this point we know how much non-Ada change to give
     -- we can use the true values for that.
 
-    let outgoingNonAda = mconcat [filterValue isNotAda v | (TxOut _ (TxOutValue _ v) _ _) <- txOuts txbodycontent]
-    let incomingNonAda = mconcat [filterValue isNotAda v | (TxOut _ (TxOutValue _ v) _ _) <- Map.elems $ unUTxO utxo]
+    let outgoingNonAda = mconcat [v | (TxOut _ (TxOutValueShelleyBased _ v) _ _) <- txOuts txbodycontent]
```
Jimbo4350: Where has the `isNotAda` filter gone?

cardano-api/internal/Cardano/Api/Ledger/Lens.hs
```diff
@@ -176,9 +179,17 @@ proposalProceduresTxBodyL w = conwayEraOnwardsConstraints w $ txBodyL . L.propos
 
 mkAdaOnlyTxOut :: ShelleyBasedEra era -> L.Addr (L.EraCrypto (ShelleyLedgerEra era)) -> L.Coin -> L.TxOut (ShelleyLedgerEra era)
 mkAdaOnlyTxOut sbe addr coin =
+  mkBasicTxOut sbe addr (mkAdaValue sbe coin)
+
+mkBasicTxOut :: ShelleyBasedEra era -> L.Addr (L.EraCrypto (ShelleyLedgerEra era)) -> L.Value (ShelleyLedgerEra era) -> L.TxOut (ShelleyLedgerEra era)
+mkBasicTxOut sbe addr value =
+  shelleyBasedEraConstraints sbe $ L.mkBasicTxOut addr value
+
+mkAdaValue :: ShelleyBasedEra era -> L.Coin -> L.Value (ShelleyLedgerEra era)
```
Jimbo4350, line 188: Another function that should live in the ledger

cardano-api/internal/Cardano/Api/Value.hs
Jimbo4350: We should in another PR.

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -712,7 +712,6 @@ toShelleyTxOut sbe = \case -- jky simplify
         L.mkBasicTxOut (toShelleyAddr addr) value
       ShelleyBasedEraAlonzo ->
         L.mkBasicTxOut (toShelleyAddr addr) value
-          & L.dataHashTxOutL .~ toAlonzoTxOutDataHash txoutdata
```
Jimbo4350:
```
     -- | A transaction output that only specifies the hash of the datum, but
     -- not the full datum value.
     --
     TxOutDatumHash   :: AlonzoEraOnwards era
                      -> Hash ScriptData
                      -> TxOutDatum ctx era
```
So this depends on the incoming `TxOut`.

cardano-api/internal/Cardano/Api/TxBody.hs
Jimbo4350: datum hashes weren't deprecated. See in ledger: `instance Crypto c => AlonzoEraTxOut (BabbageEra c) where`

Jimbo4350 (changes requested):

cardano-api/internal/Cardano/Api/Fees.hs
newhoggy: In the old code, `isNotAda` removes the ADA asset, which is equivalent to setting the ADA asset value to zero.  Later the ADA asset value is set by way of addition so the zero value is necessary:

> lovelaceToValue (Lovelace (2^(64 :: Integer)) - 1) <> nonAdaChange

The new code doesn't both to set it to zero because instead of adding `nonAdaChange` to the ADA asset value we overwrite it with `nonAdaChange` instead:

> `change & A.adaAssetL sbe .~ lovelaceToCoin maxLovelaceChange`

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -3178,14 +3138,6 @@ toShelleyTxOutAny sbe = \case
...
     cEra = shelleyBasedToCardanoEra sbe
 
-toAlonzoTxOutDataHash' :: TxOutDatum ctx AlonzoEra
-                       -> StrictMaybe (L.DataHash StandardCrypto)
-toAlonzoTxOutDataHash'  TxOutDatumNone                          = SNothing
-toAlonzoTxOutDataHash' (TxOutDatumHash _ (ScriptDataHash dh))   = SJust dh
-toAlonzoTxOutDataHash' (TxOutDatumInTx' _ (ScriptDataHash dh) _) = SJust dh
-toAlonzoTxOutDataHash' (TxOutDatumInline inlineDatumSupp _sd) =
-  case inlineDatumSupp :: BabbageEraOnwards AlonzoEra of {}
-
```
newhoggy, line 3179: The code above gets the hash and the code below also gets the hash as a `Datum`.

cardano-api/cardano-api.cabal
newhoggy: The code needs to live in the `internal` component because code in `internal` uses it.  this doesn't apply to the other modules because those are never imported by code in `internal`.

cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -827,34 +809,28 @@ makeTransactionBodyAutoBalance sbe systemstart history lpp@(LedgerProtocolParame
     -- of less than around 18 trillion ada  (2^64-1 lovelace).
     -- However, since at this point we know how much non-Ada change to give
     -- we can use the true values for that.
+    let maxLovelaceChange = Lovelace (2^(64 :: Integer)) - 1
+    let maxLovelaceFee = Lovelace (2^(32 :: Integer) - 1)
 
-    let outgoingNonAda = mconcat [filterValue isNotAda v | (TxOut _ (TxOutValue _ v) _ _) <- txOuts txbodycontent]
-    let incomingNonAda = mconcat [filterValue isNotAda v | (TxOut _ (TxOutValue _ v) _ _) <- Map.elems $ unUTxO utxo]
-    let mintedNonAda = case txMintValue txbodycontent1 of
+    let outgoing = mconcat [v | (TxOut _ (TxOutValueShelleyBased _ v) _ _) <- txOuts txbodycontent]
```
Jimbo4350, line 815: This now includes ada

cardano-api/internal/Cardano/Api/Ledger/Lens.hs
```diff
@@ -4,7 +4,15 @@
...
+  ( -- *Types
+    TxBody(..)
+
+    -- * Constructors
+  , mkAdaOnlyTxOut
+  , mkAdaValue
+
+    -- * Lenses
+  , strictMaybeL
   , L.invalidBeforeL
```
Jimbo4350, line 16: We need a clear distinction between ledger and cardano-api. We are mixing lenses here.

cardano-api/internal/Cardano/Api/TxBody.hs
Jimbo4350: The `TxOutValueByron` constructor will eventually be removed because:
```
makeByronTransactionBody :: ()
  => ByronEraOnly era
  -> TxBodyContent BuildTx era
  -> Either TxBodyError (TxBody era)
```

We don't need `TxBodyContent BuildTx Era` to construct a Byron tx. We are only interested in inputs, outputs and update proposals and we can parameterize `makeByronTransactionBody` as such.

PR #361, opened 2023-11-09
Merge `ShelleyLedgerEra` and `CardanoLedgerEra` to a single type family `LedgerEra`

newhoggy:
# Changelog

```yaml
- description: |
    Merge `ShelleyLedgerEra` and `CardanoLedgerEra` to a single type family `LedgerEra`
  type:
```

# Context

We previously defined two separate type families `ShelleyLedgerEra` and `CardanoLedgerEra` but never use `CardanoLedgerEra` anywhere.

`ShelleyLedgerEra` is basically the same as `CardanoLedgerEra` except Byron is missing.

There is no reason to have two type families because `CardanoLedgerEra` works everywhere `ShelleyLedgerEra` currently does.

This PR merges the two type families into the one `LedgerEra` type family (which includes Byron) which shortens the commonly used name.

Doing this opens the possibility of unifying Byron with the other eras for the minimal set of operations we care to maintain for hard forking purposes.

# How to trust this PR

Highlight important bits of the PR that will make the review faster. If there are commands the reviewer can run to observe the new behaviour, describe them.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/Eon/ShelleyBasedEra.hs
+++ b/cardano-api/internal/Cardano/Api/Eon/ShelleyBasedEra.hs
@@ -28,7 +28,7 @@ module Cardano.Api.Eon.ShelleyBasedEra
   , requireShelleyBasedEra
 
     -- ** Mapping to era types from the Shelley ledger library
-  , ShelleyLedgerEra
+  , LedgerEra
   , eraProtVerLow
 
   , ShelleyBasedEraConstraints
@@ -50,8 +50,6 @@ import qualified Cardano.Ledger.SafeHash as L
 import qualified Cardano.Ledger.UTxO as L
 import qualified Ouroboros.Consensus.Protocol.Abstract as Consensus
 import qualified Ouroboros.Consensus.Protocol.Praos.Common as Consensus
-import           Ouroboros.Consensus.Shelley.Eras as Consensus (StandardAllegra, StandardAlonzo,
-                   StandardBabbage, StandardConway, StandardMary, StandardShelley)
 import qualified Ouroboros.Consensus.Shelley.Ledger as Consensus
 
 import           Control.DeepSeq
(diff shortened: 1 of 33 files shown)


cardano-api/internal/Cardano/Api/Eon/ShelleyBasedEra.hs
```diff
@@ -313,25 +311,6 @@ shelleyBasedToCardanoEra ShelleyBasedEraAlonzo  = AlonzoEra
...
--- ledger library which allows writing conversion functions in a more generic
--- way.
---
-type family ShelleyLedgerEra era = ledgerera | ledgerera -> era where
-  ShelleyLedgerEra ShelleyEra = Consensus.StandardShelley
-  ShelleyLedgerEra AllegraEra = Consensus.StandardAllegra
-  ShelleyLedgerEra MaryEra    = Consensus.StandardMary
-  ShelleyLedgerEra AlonzoEra  = Consensus.StandardAlonzo
-  ShelleyLedgerEra BabbageEra = Consensus.StandardBabbage
-  ShelleyLedgerEra ConwayEra  = Consensus.StandardConway
```
newhoggy, line 333: Delete this.  Use `LedgerEra` instead.

cardano-api/internal/Cardano/Api/Eras/Core.hs
```diff
@@ -409,15 +409,15 @@ inAnyCardanoEra era a =
...
-  CardanoLedgerEra BabbageEra = L.BabbageEra L.StandardCrypto
-  CardanoLedgerEra ConwayEra  = L.ConwayEra  L.StandardCrypto
+type family LedgerEra era = ledgerera | ledgerera -> era where
+  LedgerEra ByronEra   = L.ByronEra   L.StandardCrypto
+  LedgerEra ShelleyEra = L.ShelleyEra L.StandardCrypto
+  LedgerEra AllegraEra = L.AllegraEra L.StandardCrypto
+  LedgerEra MaryEra    = L.MaryEra    L.StandardCrypto
+  LedgerEra AlonzoEra  = L.AlonzoEra  L.StandardCrypto
+  LedgerEra BabbageEra = L.BabbageEra L.StandardCrypto
+  LedgerEra ConwayEra  = L.ConwayEra  L.StandardCrypto
```
newhoggy, line 423: Rename `CardanoLedgerEra` which is never used to `LedgerEra` which we use everywhere in place of `ShelleyLedgerEra`.

smelc (approved):
LGTM :+1:

cardano-api/internal/Cardano/Api/Eras/Core.hs
carbolymer: I think the change is pretty harmless, but can you check if anything breaks in `cardano-cli`?

carbolymer (approved):

cardano-api/internal/Cardano/Api/Eras/Core.hs
newhoggy: Yep.  Already checked.

Jimbo4350 (changes requested):
This is not the direction we are taking cardano-api in. We do not care about all the eras, only mainnet and the upcoming era. The first step in this is to factor out the Byron era and only be concerned with shelley based eras. Consolidating the shelley based eras with cardano eras is moving in the wrong direction and prolongs this eventual separation.

newhoggy:
It's worth noting that this PR doesn't commit us to a particular direction.  On it's own it serves as a straightforward simplification. with the only impact on the rest of the code being a rename and removing a type we don't use anywhere.

Jimbo4350 (changes requested):
This is an unnecessary change. It is also incorrect because it implies the existence of types in Byron that do no exist. We will be able to simply things further after the Byron related code is isolated.

newhoggy:
What type is Byron type is being implied that doesn't exist?

PR #362, opened 2023-11-09
Isolation of Byron era 1/n

Jimbo4350:
# Changelog

```yaml
- description: |
    Isolation of Byron era 1/n. Changes made to cardano-api where Byron era was never being called in the first place.
  type:
```

# Context

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/gen/Test/Gen/Cardano/Api/Typed.hs
+++ b/cardano-api/gen/Test/Gen/Cardano/Api/Typed.hs
@@ -130,11 +130,12 @@ import           Cardano.Api hiding (txIns)
 import qualified Cardano.Api as Api
 import           Cardano.Api.Byron (KeyWitness (ByronKeyWitness),
                    WitnessNetworkIdOrByronAddress (..))
-import           Cardano.Api.Governance.Actions.VotingProcedure
-import           Cardano.Api.Script (scriptInEraToRefScript)
-import           Cardano.Api.Shelley
+import           Cardano.Api.Eon.AllegraEraOnwards (allegraEraOnwardsToShelleyBasedEra)
 import qualified Cardano.Api.Ledger as L
 import qualified Cardano.Api.Ledger.Lens as A
+import           Cardano.Api.Script (scriptInEraToRefScript)
+import           Cardano.Api.Shelley
+import qualified Cardano.Api.Shelley as ShelleyApi
 
 import qualified Cardano.Binary as CBOR
 import qualified Cardano.Crypto.Hash as Crypto
@@ -181,16 +182,11 @@ genAddressShelley = makeShelleyAddress <$> genNetworkId
                                        <*> genPaymentCredential
                                        <*> genStakeAddressReference
 
-genAddressInEra :: CardanoEra era -> Gen (AddressInEra era)
-genAddressInEra =
-  inEonForEra
-    (byronAddressInEra <$> genAddressByron)
-    (\sbe ->
-      Gen.choice
-        [ byronAddressInEra       <$> genAddressByron
-        , shelleyAddressInEra sbe <$> genAddressShelley
-        ]
-    )
+genAddressInEra :: ShelleyBasedEra era -> Gen (AddressInEra era)
+genAddressInEra sbe = shelleyAddressInEra sbe <$> genAddressShelley
+
+_genByronAddressInEra :: Gen (AddressInEra era)
+_genByronAddressInEra = byronAddressInEra <$> genAddressByron
 
 genKESPeriod :: Gen KESPeriod
 genKESPeriod = KESPeriod <$> Gen.word Range.constantBounded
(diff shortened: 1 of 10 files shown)


cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -2010,8 +2009,11 @@ createAndValidateTransactionBody :: ()
   => CardanoEra era
   -> TxBodyContent BuildTx era
   -> Either TxBodyError (TxBody era)
-createAndValidateTransactionBody =
-  caseByronOrShelleyBasedEra makeByronTransactionBody makeShelleyTransactionBody
+createAndValidateTransactionBody era txBodyContent =
```
Jimbo4350, line 2010: This is just a stop gap. We would have a separate function for Byron and parameterize `createAndValidateTransactionBody` on `ShelleyBasedEra era` instead.

cardano-api/internal/Cardano/Api/InMode.hs
```diff
@@ -69,8 +69,7 @@ data TxInMode where
   -- delegation certs.
   --
   TxInByronSpecial
-    :: ByronEraOnly era
-    -> Consensus.GenTx Consensus.ByronBlock
+    :: Consensus.GenTx Consensus.ByronBlock
```
Jimbo4350, line 72: The evidence is no longer necessary.

cardano-api/test/cardano-api-test/Test/Cardano/Api/Json.hs
```diff
@@ -27,27 +27,27 @@ prop_json_roundtrip_alonzo_genesis = H.property $ do
 
 prop_json_roundtrip_utxo :: Property
 prop_json_roundtrip_utxo = H.property $ do
```
Jimbo4350, line 29: We are only calling Babbage.

cardano-api/internal/Cardano/Api/InMode.hs
carbolymer: I don't think it was necessary in the first place, but it was a matter of convenience. Before you could write
```haskell
eon <- forEraMaybeEon
pure $ TxInByronSpecial eon foo
```
and after the change:
```haskell
_ :: ByronEraOnly <- forEarMaybeEon
pure $ TxInByronSpecial foo
```
so now you need to say twice that you're in byron.

cardano-api/test/cardano-api-test/Test/Cardano/Api/Typed/CBOR.hs
```diff
@@ -30,21 +30,21 @@ import           Test.Tasty.Hedgehog (testProperty)
 
 prop_roundtrip_txbody_CBOR :: Property
 prop_roundtrip_txbody_CBOR = H.property $ do
-  AnyCardanoEra era <- H.forAll $ Gen.element [minBound..AnyCardanoEra BabbageEra]
+  AnyShelleyBasedEra era <- H.forAll $ Gen.element [minBound..maxBound]
```
carbolymer, line 33: nice!

carbolymer (approved):

cardano-api/internal/Cardano/Api/InMode.hs
Jimbo4350: Can you show me where you now have to say twice that we are in Byron?

cardano-api/gen/Test/Gen/Cardano/Api/Typed.hs
```diff
@@ -711,12 +710,46 @@ genTxTotalCollateral =
 genTxFee :: CardanoEra era -> Gen (TxFee era)
 genTxFee =
   caseByronOrShelleyBasedEra
-    (pure . TxFeeImplicit)
+    undefined -- (pure . TxFeeImplicit)
```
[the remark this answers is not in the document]
Jimbo4350: This was a mistake

cardano-api/internal/Cardano/Api/InMode.hs
carbolymer: In the current usages you don't have to, because byron was hardcoded everywhere. My remark was about general usage, where you could create an era-specific value, mentioning era just once: in the value constructor.
If we're going to isolate and hardcode byron code, I guess it won't matter much.

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -838,8 +840,7 @@ deriving instance Show (TxInsReference build era)
 data TxOutValue era where
 
   TxOutValueByron
-    :: ByronEraOnly era
-    -> Lovelace
+    :: Lovelace
```
newhoggy, line 843: If we're going the path of pushing Byron to legacy status and have it separate we should move this constructor to its own type and then to a Byron-specific module so users of our API never have to see/worry/handle it.

PR #365, opened 2023-11-11
Fix round trip for empty `Value`. Add tests.

newhoggy:
# Changelog

```yaml
- description: |
    Fix round trip for empty `Value`. Add tests.
  type:
```

# Context

Resolves this issue https://github.com/input-output-hk/cardano-cli/pull/435#discussion_r1390131249

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff


cardano-api/test/cardano-api-golden/Test/Golden/Cardano/Api/Value.hs
```diff
@@ -0,0 +1,127 @@
...
+hprop_roundtrip_Value_parse_renderPretty :: Property
+hprop_roundtrip_Value_parse_renderPretty =
+  property $ do
+    ledgerValue <- forAll $ genValueDefault MaryEraOnwardsConway
+    let value = fromLedgerValue ShelleyBasedEraConway ledgerValue
+    H.noteShow_ value
+    tripping
+      value
+      renderValuePretty
+      (Parsec.parse parseValue "" . Text.unpack)
```
smelc, line 44: Since this is almost the same test as the first one, could we make the first one test the two rendering functions, by introducing an additional quantification?

smelc: I mean something like this:

```hs
forAll (Gen.element [renderValue, renderValuePretty]) >>= \renderer ->
      tripping
        value
        renderer
        (Parsec.parse parseValue "" . Text.unpack)
```

Untested because `ghc: panic!` :upside_down_face:

cardano-api/test/cardano-api-golden/Test/Golden/Cardano/Api/Value.hs
```diff
@@ -0,0 +1,127 @@
...
+
+
+hprop_goldenValue_1_lovelace :: Property
+hprop_goldenValue_1_lovelace =
+  H.propertyOnce $ do
+    valueList <- pure [(Api.AdaAssetId, 1)]
+    value     <- pure $ Text.unpack $ Api.renderValuePretty $ Api.valueFromList valueList
+
+    H.diffVsGoldenFile value "test/cardano-api-golden/files/golden/Cardano/Api/Value/value-ada-1.json"
```
smelc, line 53: Do we need the burden of tracking an extra file here? You could just put the expected value in the source code right?

cardano-api/test/cardano-api-golden/Test/Golden/Cardano/Api/Value.hs
```diff
@@ -0,0 +1,127 @@
...
+
+hprop_goldenValue1 :: Property
+hprop_goldenValue1 =
+  H.propertyOnce $ do
+    policyId  <- pure $ Api.PolicyId "a0000000000000000000000000000000000000000000000000000000"
+    assetName <- pure $ Api.AssetName "asset1"
+    valueList <- pure [(Api.AssetId policyId assetName, 1)]
+    value     <- pure $ Text.unpack $ Api.renderValuePretty $ Api.valueFromList valueList
+
+    H.diffVsGoldenFile value "test/cardano-api-golden/files/golden/Cardano/Api/Value/value-asset1-1.json"
```
smelc, line 63: Same thing here: do we need the burden of tracking an extra file here? Maybe put the expected value directly?

cardano-api/test/cardano-api-golden/Test/Golden/Cardano/Api/Value.hs
newhoggy: I think it's better separate because in the case of failure it is easier to identify which one caused the problem.

cardano-api/test/cardano-api-golden/Test/Golden/Cardano/Api/Value.hs
newhoggy: We prefer golden tests because we could _choose_ to change the format in which case we delete the golden file and regenerate it and don't have to hand-craft the expected result in a test and still be able to see the diff.

PR #368, opened 2023-11-14
Use `L.Coin` instead of `Lovelace`

newhoggy:
# Changelog

```yaml
- description: |
    Use `L.Coin` instead of `Lovelace`
    Delete:
    - `toShelleyLovelace`
    - `fromShelleyLovelace`
    - `lovelaceToCoin`
    - `coinToLovelace`
  type:
```

# Context

A lot of the code is converting between `L.Coin` and `Lovelace`.  Switching to `L.Coin` saves a lot of conversions.

Note that because we now use the ledger type, it also follows that we use the ledger type class instances.  This can have knock-on changes.  See the golden file diffs for examples.

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/Orphans.hs
+++ b/cardano-api/internal/Cardano/Api/Orphans.hs
@@ -20,6 +20,7 @@ import qualified Cardano.Ledger.Alonzo.PParams as Ledger
 import qualified Cardano.Ledger.Babbage.PParams as Ledger
 import           Cardano.Ledger.BaseTypes (strictMaybeToMaybe)
 import qualified Cardano.Ledger.BaseTypes as Ledger
+import qualified Cardano.Ledger.Coin as L
 import qualified Cardano.Ledger.Conway.PParams as Ledger
 import           Cardano.Ledger.Crypto (StandardCrypto)
 import qualified Cardano.Ledger.Crypto as CC (Crypto)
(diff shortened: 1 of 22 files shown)


cardano-api/internal/Cardano/Api/Orphans.hs
```diff
@@ -56,6 +57,16 @@ deriving instance Data CBOR.DeserialiseFailure
...
 
+-- | These instances originally existed on the Lovelace type.
+-- As the Lovelace type is deleted and we use L.Coin instead,
+-- these instances are added to L.Coin.  The instances are
+-- purely for the convenience of writing expressions involving
+-- L.Coin but be aware that not all uses of these typeclasses
+-- are valid.
+deriving newtype instance Real L.Coin
+deriving newtype instance Integral L.Coin
+deriving newtype instance Num L.Coin
```
newhoggy, line 68: These instances are added because we currently rely on them heavily on the `Lovelace` type.

Some consideration is required as to whether we want to remove these and stay with the instances that ledger exports.

Jimbo4350: Where are these used in the api/cli?

Jimbo4350 (approved):
LGTM! Before merging, lets discuss your comment:
> not all uses of these typeclasses are valid.

cardano-api/internal/Cardano/Api/Orphans.hs
newhoggy: All over the place.

For example when we use:

* literals like `0`
* operators like `+`, `-`, `*`, `/`.
* conversions like `fromIntegeral`

carbolymer (approved):

carbolymer:
Superceded by https://github.com/IntersectMBO/cardano-api/pull/475

PR #370, opened 2023-11-15
Use `Pretty` for rendering errors instead of `Show`

newhoggy:
# Changelog

```yaml
- description: |
    Use `Pretty` for rendering errors instead of `Show`
  type:
```

# Context

`Pretty` is a typeclass from the `prettyprinter` package.

It has support for powerful formatting options such as colour and layout.

By switching to `Pretty` we are able to use these features.  Furthermore, `plutus` errors and types have `Pretty` instances so can utilise that to pretty print their error messages in an ergonomic way.

This package is also used by `optparse-applicative` which means it is already a dependency.

`Pretty` also doesn't suffer the problem with `Show` where many of the instances render strings that are not friendly to users.

Ledger also have `Show` instances that may not behave the way we want.

Our `Pretty` instances can help us break away from this status quo.

In order to minimise change, this PR does not utilise these features.

Instead, it preserves formatting of all errors exactly as before leaving future PRs to tidy up the formatting and layout of our rendered errors.

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/DRepMetadata.hs
+++ b/cardano-api/internal/Cardano/Api/DRepMetadata.hs
@@ -31,6 +31,7 @@ import qualified Cardano.Ledger.Keys as Shelley
 import           Data.ByteString (ByteString)
 import qualified Data.ByteString as BS
 import           Data.Either.Combinators (maybeToRight)
+import           Prettyprinter
 
 -- ----------------------------------------------------------------------------
 -- DRep metadata
@@ -68,13 +69,13 @@ data DRepMetadataValidationError
   deriving Show
 
 instance Error DRepMetadataValidationError where
-  displayError = \case
+  prettyError = \case
     DRepMetadataInvalidLengthError maxLen actualLen ->
       mconcat
         [ "DRep metadata must consist of at most "
-        , show maxLen
+        , pretty maxLen
         , " bytes, but it consists of "
-        , show actualLen
+        , pretty actualLen
         , " bytes."
         ]
 
--- a/cardano-api/internal/Cardano/Api/Eras/Core.hs
+++ b/cardano-api/internal/Cardano/Api/Eras/Core.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE ConstraintKinds #-}
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DerivingVia #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE LambdaCase #-}
@@ -49,6 +50,7 @@ module Cardano.Api.Eras.Core
   ) where
 
 import           Cardano.Api.HasTypeProxy
+import           Cardano.Api.Pretty
 
 import qualified Cardano.Ledger.Api as L
 
@@ -236,9 +238,11 @@ data CardanoEra era where
      ConwayEra  :: CardanoEra ConwayEra
      -- when you add era here, change `instance Bounded AnyCardanoEra`
 
-deriving instance Eq   (CardanoEra era)
-deriving instance Ord  (CardanoEra era)
-deriving instance Show (CardanoEra era)
+deriving instance Eq    (CardanoEra era)
+deriving instance Ord   (CardanoEra era)
+deriving instance Show  (CardanoEra era)
+
+deriving via (ShowOf (CardanoEra era)) instance Pretty (CardanoEra era)
 
 instance ToJSON (CardanoEra era) where
    toJSON ByronEra   = "Byron"
--- a/cardano-api/internal/Cardano/Api/Error.hs
+++ b/cardano-api/internal/Cardano/Api/Error.hs
@@ -1,6 +1,8 @@
 {-# LANGUAGE DeriveFunctor #-}
 {-# LANGUAGE ExistentialQuantification #-}
 {-# LANGUAGE GADTSyntax #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE OverloadedStrings #-}
 
 -- | Class of errors used in the Api.
 --
@@ -10,23 +12,25 @@ module Cardano.Api.Error
   , ErrorAsException(..)
   , FileError(..)
   , fileIOExceptT
+  , displayError
   ) where
+
+import           Cardano.Api.Pretty
 
 import           Control.Exception (Exception (..), IOException, throwIO)
 import           Control.Monad.Except (throwError)
 import           Control.Monad.IO.Class (MonadIO)
 import           Control.Monad.Trans.Except (ExceptT)
 import           Control.Monad.Trans.Except.Extra (handleIOExceptT)
+import           Prettyprinter
 import           System.Directory (doesFileExist)
 import           System.IO (Handle)
 
-class Show e => Error e where
-
-    displayError :: e -> String
+class Error e where
+  prettyError :: e -> Doc ann
 
 instance Error () where
-    displayError () = ""
-
+  prettyError () = ""
 
 -- | The preferred approach is to use 'Except' or 'ExceptT', but you can if
 -- necessary use IO exceptions.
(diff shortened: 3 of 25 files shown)


cardano-api/internal/Cardano/Api/Error.hs
```diff
@@ -10,23 +12,25 @@ module Cardano.Api.Error
...
 import           Control.Monad.Trans.Except (ExceptT)
 import           Control.Monad.Trans.Except.Extra (handleIOExceptT)
+import           Prettyprinter
 import           System.Directory (doesFileExist)
 import           System.IO (Handle)
 
-class Show e => Error e where
-
-    displayError :: e -> String
+class Error e where
```
carbolymer, line 29: Why not just use `Pretty a` instead?

cardano-api/internal/Cardano/Api/DRepMetadata.hs
```diff
@@ -68,13 +69,13 @@ data DRepMetadataValidationError
   deriving Show
 
 instance Error DRepMetadataValidationError where
-  displayError = \case
+  prettyError = \case
     DRepMetadataInvalidLengthError maxLen actualLen ->
       mconcat
```
carbolymer, line 74: use `hsep` instead

cardano-api/internal/Cardano/Api/Pretty.hs
```diff
@@ -0,0 +1,55 @@
...
+  ) where
+
+import qualified Data.Text.Lazy as TextLazy
+import           Prettyprinter
+import           Prettyprinter.Render.Terminal
+
+type Ann = AnsiStyle
+
+prettyToString :: Doc AnsiStyle -> String
+prettyToString =  TextLazy.unpack . prettyToText
```
carbolymer, line 25:
```suggestion
prettyToString = show
```
[The show instance](https://hackage.haskell.org/package/prettyprinter-1.7.1/docs/src/Prettyprinter.Internal.html#line-2199) is doing exactly the same, but without going through `unpack` which is [`O(n)`](https://hackage.haskell.org/package/text-2.1/docs/Data-Text-Lazy.html#v:unpack)

cardano-api/internal/Cardano/Api/Eras/Core.hs
```diff
@@ -236,9 +237,12 @@ data CardanoEra era where
...
 
-deriving instance Eq   (CardanoEra era)
-deriving instance Ord  (CardanoEra era)
-deriving instance Show (CardanoEra era)
+deriving instance Eq    (CardanoEra era)
+deriving instance Ord   (CardanoEra era)
+deriving instance Show  (CardanoEra era)
+
+instance Pretty (CardanoEra era) where
+  pretty = pretty . show
```
carbolymer, line 245:
```suggestion
  pretty = viaShow
```
I think we can go further and for such use cases create a newtype wrapper for deriving via, e.g.:
```haskell
data PrettyViaShow a = PrettyViaShow a
instance Show a => Show (PrettyViaShow a) where show (PrettyViaShow a) = show a
instance Show a => Pretty (PrettyViaShow a) where pretty = viaShow
```

and then you could write
```haskell
deriving via (PrettyViaShow (Cardano era)) instance Pretty (Cardano era)
```
(but I guess it's less characters than my first proposed change :D )

cardano-api/internal/Cardano/Api/DRepMetadata.hs
newhoggy: I went for the conservation change.  Happy to have a second pass to clean it up.

cardano-api/internal/Cardano/Api/Error.hs
```diff
@@ -55,20 +67,23 @@ data FileError e = FileError FilePath e
...
+    FileDoesNotExistError path ->
+      "Error file not found at: " <> pretty path
+    FileIOError path ioe ->
+      pretty path <> ": " <> pretty (displayException ioe)
+    FileError path e ->
+      pretty path <> ": " <> prettyError e
 
 instance Error IOException where
-  displayError = show
+  prettyError = pretty . show
```
carbolymer, line 86:
```suggestion
  prettyError = viaShow
```

cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -351,49 +351,59 @@ data ScriptExecutionError =
...
-     "The pointers that can be resolved are: " <> show resolveable
-
-  displayError (ScriptErrorMissingCostModel language) =
-      "No cost model was found for language " <> show language
+  prettyError = \case
+    ScriptErrorMissingTxIn txin ->
+      "The supplied UTxO is missing the txin " <> pretty (renderTxIn txin)
+
+    ScriptErrorTxInWithoutDatum txin ->
+      mconcat
```
carbolymer, line 359: use `hsep`

cardano-api/internal/Cardano/Api/Eras/Core.hs
newhoggy: I approve of `deriving via` in general.  Can add that in a new PR.

cardano-api/internal/Cardano/Api/Pretty.hs
```diff
@@ -0,0 +1,55 @@
...
+magenta = annotate (color Magenta)
+
+cyan :: Doc AnsiStyle -> Doc AnsiStyle
+cyan = annotate (color Cyan)
+
+white :: Doc AnsiStyle -> Doc AnsiStyle
+white = annotate (color White)
+
+pshow :: Show a => a -> Doc ann
+pshow = pretty . show
```
carbolymer, line 55: why not use `viaShow`?

cardano-api/internal/Cardano/Api/ProtocolParameters.hs
```diff
@@ -1856,18 +1857,20 @@ data ProtocolParametersError
...
-  displayError PParamsErrorMissingAlonzoProtocolParameter = mconcat
-    [ "The Alonzo era protocol parameters in use is missing one or more of the "
-    , "following fields: UTxOCostPerWord, CostModels, Prices, MaxTxExUnits, "
-    , "MaxBlockExUnits, MaxValueSize, CollateralPercent, MaxCollateralInputs. Did "
-    , "you intend to use an Alonzo era protocol parameters value?"
-    ]
-
+  prettyError = \case
+    PParamsErrorMissingMinUTxoValue (AnyCardanoEra era) ->
+      mconcat
```
carbolymer, line 1862: use `hsep`

cardano-api/internal/Cardano/Api/TxBody.hs
```diff
@@ -341,6 +341,9 @@ instance Eq TxOutInAnyEra where
       Just Refl -> out1 == out2
       Nothing   -> False
 
+instance Pretty TxOutInAnyEra where
+  pretty = pretty . show
```
carbolymer, line 345:
```suggestion
  pretty = viaShow
```

carbolymer (commented):
I like the change, but I have a few remarks.

cardano-api/internal/Cardano/Api/Error.hs
newhoggy: Because of this:

```
instance Error () where
  prettyError () = ""
```

`instance Pretty ()` already exists and does something different.

We had used `oops` to remove the need for this instead, but we chose not to adopt `oops` in `cardano-api` for the interim so we'll need to keep `Error`.

cardano-api/internal/Cardano/Api/DRepMetadata.hs
newhoggy: `hsep` work, so I'll just make the change.

cardano-api/internal/Cardano/Api/Error.hs
carbolymer: Where are we using this instance?

cardano-api/internal/Cardano/Api/DRepMetadata.hs
newhoggy: Note, `hsep` behaves different to `mconcat` in that it inserts spaces, so something it is not directly substitutable.

newhoggy: I'll hold off on this.

cardano-api/internal/Cardano/Api/Pretty.hs
newhoggy: Done!

cardano-api/internal/Cardano/Api/Fees.hs
newhoggy: Postponing

cardano-api/internal/Cardano/Api/Error.hs
newhoggy: Done!

cardano-api/internal/Cardano/Api/Eras/Core.hs
newhoggy: Done!

cardano-api/internal/Cardano/Api/Error.hs
newhoggy: There are `5` occurences in the `cardano-api` and more than `25` occurences in `cardano-cli`.

newhoggy: Moreover, the reason `Error` existed in the first place instead of using show is that one instance.  This PR isn't attempting to change the status quo.

carbolymer: I guess there's no easy way around `FileError` here. Let's leave it as it is.

Jimbo4350 (approved):

PR #379, opened 2023-11-18
`ToJSON` instance for `TxValidationError`

newhoggy:
# Changelog

```yaml
- description: |
    `ToJSON` instance for `TxValidationError`
  type:
```

# Context

This is needed for reporting errors in machine readable JSON.

Note that the error tree is not completely converted to JSON structure.  Some sub-branches remain as strings because the ledger does not export enough symbols to go all the way down to the leaves.

In any case, these instances are a stop-gap until the instances are implemented in the ledger proper.

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/InMode.hs
+++ b/cardano-api/internal/Cardano/Api/InMode.hs
@@ -219,59 +219,32 @@ toConsensusTxId (TxIdInMode ConwayEra txid) =
 -- transaction to a local node. The errors are specific to an era.
 --
 data TxValidationError era where
+  ByronTxValidationError
+    :: Consensus.ApplyTxErr Consensus.ByronBlock
+    -> TxValidationError ByronEra
 
-     ByronTxValidationError
-       :: Consensus.ApplyTxErr Consensus.ByronBlock
-       -> TxValidationError ByronEra
-
-     ShelleyTxValidationError
-       :: ShelleyBasedEra era
-       -> Consensus.ApplyTxErr (Consensus.ShelleyBlock (ConsensusProtocol era) (ShelleyLedgerEra era))
-       -> TxValidationError era
+  ShelleyTxValidationError
+    :: ShelleyBasedEra era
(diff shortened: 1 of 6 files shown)


cardano-api/internal/Cardano/Api/InMode.hs
```diff
@@ -220,20 +220,22 @@ toConsensusTxId (TxIdInMode ConwayEra txid) =
 --
 data TxValidationError era where
   ByronTxValidationError
-    :: Consensus.ApplyTxErr Consensus.ByronBlock
-    -> TxValidationError ByronEra
+    :: ByronEraOnly era
```
Jimbo4350, line 228: `ByronEraOnly era` is unnecessary here.

Jimbo4350 (dismissed):

cardano-api/internal/Cardano/Api/InMode.hs
newhoggy: Removed

cardano-api/internal/Cardano/Api/Via/ShowOf.hs
```diff
@@ -0,0 +1,23 @@
...
+  ( ShowOf(..)
+  ) where
+
+import           Data.Aeson
+import qualified Data.Aeson.Key as Key
+import           Data.Aeson.Types
+import qualified Data.Text as Text
+import           Prettyprinter
+
+newtype ShowOf a = ShowOf a
```
smelc, line 11: I think a little motivation comment would be helpful in the long term. I believe here the main point of `ShowOf` is that you want to capture when `Show` can be lifted to `ToJSON`, right? I think stating this motivation would help maintenance in the long term (and onboarding of newcomers to the codebase (speaking for a friend)).

smelc (approved):
LGTM

carbolymer (approved):

PR #402, opened 2023-12-13
Era handling

Jimbo4350:
# Changelog

```yaml
- description: |
    Implement Era GADT and UseEra class as an alternative to the existing era handling code
  type:
```

# Context

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/Protocol/Version.hs
+++ b/cardano-api/internal/Cardano/Api/Protocol/Version.hs
@@ -1,0 +1,102 @@
+{-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilyDependencies #-}
+{-# LANGUAGE TypeOperators #-}
+
+-- | This module defines the protocol versions corresponding to the eras in the Cardano blockchain.
+module Cardano.Api.Protocol.Version
+  ( BabbageEra
+  , ConwayEra
+  , Era (..)
+  , MaxSupportedVersion
+  , MinSupportedVersion
+  , SupportedProtocolVersionRange
+  , UseEra
+  , VersionToSbe
+  , getProtocolVersion
+  , protocolVersionToSbe
+  ) where
+
+import           Cardano.Api.Eon.ShelleyBasedEra (ShelleyBasedEra (..))
+import qualified Cardano.Api.Eras.Core as Api
+
+import           GHC.TypeLits
+
+-- | Users typically interact with the latest features on the mainnet or experiment with features
+-- from the upcoming era. Hence, the protocol versions are limited to the current mainnet era
+-- and the next era (upcoming era).
+--
+-- - 'MinSupportedVersion': Represents the minimum protocol version, aligning with the current
+--   era on mainnet (Babbage era).
+-- - 'MaxSupportedVersion': Represents the maximum protocol version, corresponding to the next
+--   era planned for Cardano's mainnet (Conway era).
+
+-- | The minimum supported protocol version, corresponding to the Babbage era on Cardano's mainnet.
+type MinSupportedVersion = 8 :: Nat
+
+-- | The maximum supported protocol version, representing the upcoming Conway era.
+type MaxSupportedVersion = 9 :: Nat
+
+type BabbageEra = 8 :: Nat
+type ConwayEra = 9 :: Nat
+
+type SupportedProtocolVersionRange (version :: Nat) =
+    ( MinSupportedVersion <= version
+    , version <= MaxSupportedVersion
+    )
+
+-- Allows us to gradually change the api without breaking things.
+-- This will eventually be removed.
+type family VersionToSbe (version :: Nat) where
+  VersionToSbe BabbageEra = Api.BabbageEra
+  VersionToSbe ConwayEra = Api.ConwayEra
+
+-- | Represents the eras in Cardano's blockchain history.
+--
+-- Instead of enumerating every possible era, we use two constructors:
+-- 'CurrentEra' and 'UpcomingEra'. This design simplifies the handling
+-- of eras, especially for 'cardano-api' consumers who are primarily concerned
+-- with the current mainnet era and the next era for an upcoming hardfork.
+--
+-- Usage:
+-- - 'CurrentEra': Reflects the era currently active on mainnet.
+-- - 'UpcomingEra': Represents the era planned for the next hardfork.
+--
+-- After a hardfork, 'cardano-api' should be updated promptly to reflect
+-- the new mainnet era in 'CurrentEra'.
+--
+-- Each era is associated with a specific protocol version:
+-- - 'BabbageEra' for 'CurrentEra'
+-- - 'ConwayEra' for 'UpcomingEra'
+data Era (version :: Nat) where
+  -- | The era currently active on Cardano's mainnet.
+  CurrentEra
+    :: SupportedProtocolVersionRange BabbageEra
+    => Era BabbageEra
+  -- | The era planned for the next hardfork on Cardano's mainnet.
+  UpcomingEra
+    :: SupportedProtocolVersionRange ConwayEra
+    => Era ConwayEra
+
+
+protocolVersionToSbe
+  :: Era version
+  -> ShelleyBasedEra (VersionToSbe version)
+protocolVersionToSbe CurrentEra = ShelleyBasedEraBabbage
+protocolVersionToSbe UpcomingEra = ShelleyBasedEraConway
+
+-------------------------------------------------------------------------
+
+-- | Type class interface for the 'Era' type.
+
+class UseEra (version :: Nat) where
+  getProtocolVersion :: Era version
+
+instance UseEra BabbageEra where
+  getProtocolVersion = CurrentEra
+
+instance UseEra ConwayEra where
+  getProtocolVersion = UpcomingEra
--- a/cardano-api/cardano-api.cabal
+++ b/cardano-api/cardano-api.cabal
@@ -118,6 +118,7 @@ library internal
                         Cardano.Api.OperationalCertificate
                         Cardano.Api.Pretty
                         Cardano.Api.Protocol
+                        Cardano.Api.Protocol.Version
                         Cardano.Api.ProtocolParameters
                         Cardano.Api.Query
                         Cardano.Api.Query.Expr
@@ -230,6 +231,7 @@ library
                         Cardano.Api.ChainSync.Client
                         Cardano.Api.ChainSync.ClientPipelined
                         Cardano.Api.Crypto.Ed25519Bip32
+                        Cardano.Api.Experimental
                         Cardano.Api.Shelley
                         -- TODO: Eliminate Cardano.Api.Ledger when
                         -- cardano-api only depends on modules
--- a/cardano-api/internal/Cardano/Api/ReexposeLedger.hs
+++ b/cardano-api/internal/Cardano/Api/ReexposeLedger.hs
@@ -94,6 +94,7 @@ module Cardano.Api.ReexposeLedger
   , urlToText
   , textToUrl
   , portToWord16
+  , ProtVer(..)
   , strictMaybeToMaybe
   , maybeToStrictMaybe
 
@@ -118,9 +119,9 @@ import           Cardano.Ledger.Api.Tx.Cert (pattern AuthCommitteeHotKeyTxCert,
                    pattern ResignCommitteeColdTxCert, pattern RetirePoolTxCert,
                    pattern UnRegDRepTxCert, pattern UnRegDepositTxCert, pattern UnRegTxCert)
 import           Cardano.Ledger.Babbage.Core (CoinPerByte (..))
-import           Cardano.Ledger.BaseTypes (DnsName, Network (..), StrictMaybe (..), Url,
-                   boundRational, dnsToText, maybeToStrictMaybe, portToWord16, strictMaybeToMaybe,
-                   textToDns, textToUrl, unboundRational, urlToText)
+import           Cardano.Ledger.BaseTypes (DnsName, Network (..), ProtVer (..), StrictMaybe (..),
+                   Url, boundRational, dnsToText, maybeToStrictMaybe, portToWord16,
+                   strictMaybeToMaybe, textToDns, textToUrl, unboundRational, urlToText)
 import           Cardano.Ledger.Binary (Annotated (..))
 import           Cardano.Ledger.CertState (DRepState, csCommitteeCredsL)
 import           Cardano.Ledger.Coin (Coin (..), addDeltaCoin, toDeltaCoin)
--- a/cardano-api/internal/Cardano/Api/TxBody.hs
+++ b/cardano-api/internal/Cardano/Api/TxBody.hs
@@ -152,6 +152,15 @@ module Cardano.Api.TxBody (
     AsType(AsTxId, AsTxBody, AsByronTxBody, AsShelleyTxBody, AsMaryTxBody),
 
     getTxBodyContent,
+
+    -- Temp
+    validateTxIns,
+    guardShelleyTxInsOverflow,
+    validateTxOuts,
+    validateMetadata,
+    validateMintValue,
+    validateTxInsCollateral,
+    validateProtocolParameters,
   ) where
 
 import           Cardano.Api.Address
--- a/cardano-api/src/Cardano/Api/Experimental.hs
+++ b/cardano-api/src/Cardano/Api/Experimental.hs
@@ -1,0 +1,15 @@
+module Cardano.Api.Experimental
+  ( -- * New Era interface
+    BabbageEra
+  , ConwayEra
+  , Era
+  , MaxSupportedVersion
+  , MinSupportedVersion
+  , SupportedProtocolVersionRange
+  , UseEra
+  , VersionToSbe
+  , getProtocolVersion
+  , protocolVersionToSbe
+  ) where
+
+import           Cardano.Api.Protocol.Version


cardano-api/internal/Cardano/Api/Protocol/Version.hs
```diff
@@ -0,0 +1,63 @@
...
+data SomeProtocolVersion version where
+  CurrentProtocolVersion
+    :: SupportedProtocolVersionRange BabbageEra
+    => SomeProtocolVersion BabbageEra
+  ExperimentalProtocolVersion
+    :: SupportedProtocolVersionRange ConwayEra
+    => SomeProtocolVersion ConwayEra
+
+
+type family VersionToEra version where
```
Jimbo4350, line 50: This is temporary and allows us to change `cardano-api` bit by bit.

cardano-api/internal/Cardano/Api/Protocol/Version.hs
```diff
@@ -0,0 +1,91 @@
...
+import qualified Data.Set as Set
+import           GHC.TypeLits
+
+-- Users interacting with Cardano are likely only interested in using the latest
+-- features available on mainnet and experimenting with the upcoming era as this becomes
+-- available. Therefore we restrict the choices of protocol version to what is currently
+-- on mainnet and what is in the upcoming era.
+
+-- | Minimum supported version. Corresponds to Babbage era.
+type MinSupportedVersion = 8 :: Nat
```
carbolymer, line 28: Will this be our new protocol version thing, or is this planned to be correlated somehow with node-to-client, or shelley node-to-client protocol version?

cardano-api/internal/Cardano/Api/Protocol/Version.hs
```diff
@@ -0,0 +1,91 @@
...
+  validateTxOuts sbe txOuts
+  validateMetadata txMetadata
+  validateMintValue txMintValue
+  validateTxInsCollateral txInsCollateral languages
+  validateProtocolParameters txProtocolParams languages
+
+  case p of
+    CurrentProtocolVersion ->
+      guardShelleyTxInsOverflow (map fst txIns)
+    ExperimentalProtocolVersion -> pure ()
```
carbolymer, line 90: I think this is a good example of my doubts with using "current" and "experimental" as mutable pointers to eras. When we transition between eras by switching eras in `SomeProtocolVersion` definition, this place will change its behaviour without triggering compilation error.

An ideal solution in my opinion would be to get value-level `version` and do the check.

cardano-api/internal/Cardano/Api/Protocol/Version.hs
```diff
@@ -0,0 +1,91 @@
...
+-- available. Therefore we restrict the choices of protocol version to what is currently
+-- on mainnet and what is in the upcoming era.
+
+-- | Minimum supported version. Corresponds to Babbage era.
+type MinSupportedVersion = 8 :: Nat
+
+-- | Maximum supported version. Corresponds to Conway era.
+type MaxSupportedVersion = 9 :: Nat
+
+type BabbageEra = 8 :: Nat
```
carbolymer, line 33: I like the idea of using type-level naturals, which gives us `Ord` and `Enum`.

cardano-api/internal/Cardano/Api/Protocol/Version.hs
```diff
@@ -0,0 +1,91 @@
...
+type ConwayEra = 9 :: Nat
+
+type SupportedProtocolVersionRange version =
```
carbolymer, line 36: Can you try to force compilation error due to mismatch, I'm curious how it would look like?

cardano-api/internal/Cardano/Api/Protocol/Version.hs
Jimbo4350: This is a good point. Looking into this.

cardano-api/internal/Cardano/Api/Protocol/Version.hs
Jimbo4350: Just pushed a commit that lets you play with a toy example.

cardano-api/internal/Cardano/Api/Protocol/Version.hs
Jimbo4350: It corresponds to version defined in `ouroboros-consensus`: https://github.com/input-output-hk/ouroboros-consensus/blob/4322b57bf683cd7ac28bc3e3b9fef3fc30fbc0f2/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/Praos.hs#L120

and https://github.com/input-output-hk/ouroboros-consensus/blob/4322b57bf683cd7ac28bc3e3b9fef3fc30fbc0f2/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Node/Protocol/Cardano.hs#L229

Ideally they would export these as standalone values and we can just use them.

Ledger has min and max supported versions:
```
-- | Minimum supported version
type MinVersion = 0

-- | Maximum supported version. This is the protocol version of the next upcoming era
type MaxVersion = 10
```
but they extend over what we really want to support.

cardano-api/internal/Cardano/Api/Protocol/Version.hs
Jimbo4350: https://github.com/input-output-hk/cardano-api/pull/402#discussion_r1426920651

Jimbo4350:
>  We'll definitely want the compiler to guide us on a version bump.
Only the type synonyms will change. Have a look at my most recent commit (the commit message aswell) and play with it so see how GHC complains when the type synonyms are adjusted in the value constructors of `SomeProtocolVersion`. 

> * We might have several types like `SomeProtocolVersion`, one for each "generation" of supported and experimental version.
I believe my most recent commit also deals with this concern unless I'm misunderstanding.

cardano-api/internal/Cardano/Api/Protocol/Version.hs
Jimbo4350: > is this planned to be correlated somehow with node-to-client, or shelley node-to-client protocol version?

It's worth exploring this (albeit not right now). It would be useful to have a single type class that gives you these values along with GHC complaining about any unmatched cases.

smelc:
I agree that using type-level naturals seems quite natural and smooth :+1: 

But I'm wondering how this interacts with code that is going to support older versions than what is in `SupportedProtocolVersionRange`. Would we also introduce `type AlonzoEra = 7 :: Nat` and friends? Or would the "old" stuff not be updated with this new system (like the `legacy` part of cardano-cli?)

cardano-api/internal/Cardano/Api/Protocol/Version.hs
carbolymer: with:
```haskell
  Foo
    :: SupportedProtocolVersionRange PostConwayEra
    => SomeProtocolVersion PostConwayEra
```
this:
```haskell
err :: ()
err = (undefined :: SomeProtocolVersion v -> ()) Foo
```
gives:
```
internal/Cardano/Api/Protocol/Version.hs:136:50: error: [GHC-64725]
    • Cannot satisfy: PostConwayEra <= MaxSupportedVersion
    • In the first argument of ‘undefined ::
                                  SomeProtocolVersion v -> ()’, namely
        ‘Foo’
      In the expression: (undefined :: SomeProtocolVersion v -> ()) Foo
      In an equation for ‘err’:
          err = (undefined :: SomeProtocolVersion v -> ()) Foo
    |
136 | err = (undefined :: SomeProtocolVersion v -> ()) Foo
    |                
```

Jimbo4350: Yep this is working as intended. `PostConwayEra` exceeds our upper bound of `BabbageEra`. We would need to adjust  `MaxSupportedVersion` to include `PostConwayEra`.

Jimbo4350:
> I agree that using type-level naturals seems quite natural and smooth 👍
> 
> But I'm wondering how this interacts with code that is going to support older versions than what is in `SupportedProtocolVersionRange`. Would we also introduce `type AlonzoEra = 7 :: Nat` and friends? Or would the "old" stuff not be updated with this new system (like the `legacy` part of cardano-cli?)

So we are operating on the assumption that consumers of cardano-api and cli are only interested in the latest era. Because of this we would not introduce `Nat`s for anything prior to `BabbageEra`.

cardano-api/internal/Cardano/Api/Protocol/Version.hs
```diff
@@ -0,0 +1,156 @@
...
+-- For the wallet team. Exposing a class interface side by side should be easy.
+classExample
+  :: UseProtocolVersion version
+  => [(TxIn, BuildTxWith BuildTx (UpdatedWitness version))]
+  -> Either TxBodyError ()
+classExample = example getProtocolVersion
+
+
+classExampleUsage :: Either TxBodyError ()
+classExampleUsage = classExample @BabbageEra []
```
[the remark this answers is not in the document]
Jimbo4350: So there is a relationship between the protocol version and the era but the average user likely does not need to know/care about this. What about a rename to some thing like: 
```
class UseEra era where 
  getEra :: Era era
  
 data Era (era :: Nat) where
  CurrentEra
    :: SupportedEras BabbageEra
    => Era BabbageEra
  UpcomingEra
    :: SupportedEras ConwayEra
    => Era ConwayEra
```
Users don't need to be aware that a particular protocol version corresponds to an era.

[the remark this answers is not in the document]
Jimbo4350: > One use case we have upcoming is that we want to deliberately stay with BabbageEra when creating transactions, but be able to read BabbageEra or ConwayEra transactions (the latter is not so important as we should be able to deserialize a BabbageEra tx from ConwayEra blocks).

You're interested in this deserialisation for the purposes of experimenting the the ConwayEra correct? After the hardfork, our plan would be to drop support for BabbageEra transactions.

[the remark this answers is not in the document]
ch1bo: > You're interested in this deserialisation for the purposes of experimenting the the ConwayEra correct?

Not exactly. We are interested in deserializing `Babbage` transactions from `Conway` blocks as our codebase is not fully parameterized on the era, but uses one hard-coded era (currently `Babbage`) in both: creating and observing transactions. That means, that when building a `hydra-node` version, we need to pick an era we do talk/understand. We do want to build a `hydra-node` that works in `Babbage` and `Conway` so we would want to build/inspect `Babbage` transactions (and also expose a `Babbage` ledger on L2).. until we drop support for `Babbage` and switch to full `Conway`. Hope that makes sense? NB: Making our codebase parameterized by `era` will require us to still decide which era we want to produce transactions for in a given build. Unless we make it configurable (a) or run-time switching (b) .. which we would like to avoid due to unnecessariy complexity in usage (a) or code (b).

> It would be useful if you could point to your code where you currently using this functionality.

We are currently only getting started on that, so nothing to point to yet.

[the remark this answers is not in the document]
ch1bo: Regarding the snippet: `classExample :: UseEra era => [(TxIn, BuildTxWith BuildTx (UpdatedWitness era))] -> Either TxBodyError ()` would make more sense to me as a user :+1: 
Would that mean I use it as `classExample @CurrentEra []` and for any given version of `cardano-api` that would be either `Babbage` or `Conway`?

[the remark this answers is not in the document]
Jimbo4350: Your use case makes sense. What I'm proposing will give access to both Babbage and Conway transaction serialization and those will be available until after mainnet has hard forked to Conway. So I think this will be fine.

> Would that mean I use it as classExample @CurrentEra [] and for any given version of cardano-api that would be either Babbage or Conway?

It would still be `classExample @BabbageEra []`. I think we would want some breakage when mainnet is in the `ConwayEra` and we need to deprecate `BabbageEra`. Users should be forced to acknowledge the change. However it would be a simple type synonym change rather than having to update the `CurrentEra` value constructor.

cardano-api/internal/Cardano/Api/Protocol/Version.hs
```diff
@@ -0,0 +1,105 @@
...
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilyDependencies #-}
+{-# LANGUAGE TypeOperators #-}
+
+-- | This module defines the protocol versions corresponding to the eras in the Cardano blockchain.
+module Cardano.Api.Protocol.Version
+  ( BabbageEra
+  , ConwayEra
+  , Era
```
Jimbo4350, line 13: I need to expose the Era constructors

carbolymer: But `Era` line is not exporting them.

Do we really need to export them? I find them to be footguns. It's fine to use them internally for eras infrastructure code, but it seems dangerous to use them directly in feature toggling code.

carbolymer (approved):

cardano-api/internal/Cardano/Api/Protocol/Version.hs
```diff
@@ -0,0 +1,105 @@
...
+-- - 'ConwayEra' for 'UpcomingEra'
+data Era (version :: Nat) where
+  -- | The era currently active on Cardano's mainnet.
+  CurrentEra
+    :: SupportedProtocolVersionRange BabbageEra
+    => Era BabbageEra
+  -- | The era planned for the next hardfork on Cardano's mainnet.
+  UpcomingEra
+    :: SupportedProtocolVersionRange ConwayEra
+    => Era ConwayEra
```
carbolymer, line 82: How about changing `Era` definition a bit to:
```suggestion
  -- | The era currently active on Cardano's mainnet.
  BabbageEra
    :: SupportedProtocolVersionRange BabbageEra
    => Era BabbageEra
  -- | The era planned for the next hardfork on Cardano's mainnet.
  ConwayEra
    :: SupportedProtocolVersionRange ConwayEra
    => Era ConwayEra

pattern CurrentEra :: Era BabbageEra
pattern CurrentEra = BabbageEra

pattern UpcomingEra :: Era ConwayEra
pattern UpcomingEra = ConwayEra

{-# COMPLETE CurrentEra, UpcomingEra #-}
```
and export constructors, but not patterns. This could be useful when we would like to rely on era-dependent behaviour, but have compile-time checks for eras.

carbolymer, line 82: Those constraints seem redundant. The should be easily recoverable wherever needed.
```suggestion
    :: Era BabbageEra
  -- | The era planned for the next hardfork on Cardano's mainnet.
  UpcomingEra
    :: Era ConwayEra
```

cardano-api/internal/Cardano/Api/Protocol/Version.hs
```diff
@@ -0,0 +1,105 @@
...
+
+type family VersionToEra (version :: Nat) where
```
carbolymer, line 84: This type family is fine, but I'm not sure about naming. Here `Era` means our old `Era`, but in this module we're using `Era` also as a name for the new type. This may be a bit confusing for library users, which era type (new version or old) is used in the name. Maybe we should use `data Version` instead of `data Era`?

In general, that would be only an issue for the transition period, before the codebase abandons eons usage.

cardano-api/internal/Cardano/Api/Protocol/Version.hs
Jimbo4350: I envision two interfaces: the typeclass `UseEra` and the GADT `Era`. So our functions will either require satisfaction of the constraint or choosing one of the GADT constructors.

cardano-api/internal/Cardano/Api/Protocol/Version.hs
Jimbo4350: We already have compile time checks for eras via `SupportedProtocolVersionRange`. The issue with your approach is we now have to be continuously adding and remove data constructors. The patterns seem like an extra unnecessary step because it amounts to the same thing: the user would only experience the change of the type synonym in the pattern definition.

cardano-api/internal/Cardano/Api/Protocol/Version.hs
Jimbo4350: This is confusing but it will be internal. I'll add a comment. We can come back to the name change.

Jimbo4350: I've changed it to `VersionToSbe`. It's not great but this interface hinges on the fact that protocol versions map to specific eras.

cardano-api/internal/Cardano/Api/Protocol/Version.hs
carbolymer: So it's purpose is to trigger a compiler error in case there are constructors, which are outside of the supported version range. Do I understand it correctly?

carbolymer: >The patterns seem like an extra unnecessary step because it amounts to the same thing: the user would only experience the change of the type synonym in the pattern definition.

I would say it's a different thing. My concern is that if an user decides to write a function like:
```haskell
doThing :: Era era -> ()
doThing = \case
  CurrentEra -> enableFeature
  UpcomingEra -> disableFeature
```
he won't get any errors when we switch current eras, but the application will change its behaviour unnoticed.
My point is not to export type constructors representing `CurrentEra` and `UpcomingEra`, to prevent creating situations like that.

Jimbo4350: I see what you're saying. Let me think about this further.

Jimbo4350: So when we modify `SupportedProtocolVersionRange`'s range, users will receive a compilation error saying the current era they are instantiating to is not supported. They will have to update their type synonyms.

Jimbo4350: https://github.com/IntersectMBO/cardano-api/pull/414

PR #403, opened 2023-12-15
Use ShelleyNTC for version check in node queries

carbolymer:
# Changelog

```yaml
- description: |
    Use ShelleyNTC for version check in node queries
  type:
```

# Context

Fixes https://github.com/input-output-hk/cardano-api/issues/391

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/IPC/Monad.hs
+++ b/cardano-api/internal/Cardano/Api/IPC/Monad.hs
@@ -1,6 +1,13 @@
+{-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DerivingVia #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE GeneralisedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeOperators #-}
+{-# OPTIONS_GHC -Wno-redundant-constraints #-}
 
 module Cardano.Api.IPC.Monad
   ( LocalStateQueryExpr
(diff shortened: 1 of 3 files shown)


cardano-api/internal/Cardano/Api/IPC/Monad.hs
```diff
@@ -86,21 +121,121 @@ setupLocalStateQueryExpr waitDone mPointVar' resultVar' ntcVersion f =
...
+  case isQuerySupprted shelleyNtcVersion q of
+    Left e -> pure $ Left e
+    Right isShelleyNtcSupported
+      | ntcVersion >= minNtcVersion && isShelleyNtcSupported ->
+          fmap Right . LocalStateQueryExpr . ReaderT $ \_ -> ContT $ \f -> pure $
+            Net.Query.SendMsgQuery q $
+              Net.Query.ClientStQuerying
+              { Net.Query.recvMsgResult = f
+              }
+      | otherwise -> pure . Left $ UnsupportedNtcVersionError minNtcVersion ntcVersion
```
carbolymer, line 172: * [ ] TODO: include shelley NTC version info and/or use separate error for it

Jimbo4350 (changes requested):
I haven't look at this yet but I'm prematurely requesting changes as this needs to be carefully considered.

amesgen (commented):
The new logic here looks correct to me.

We also talked about this in the Consensus Office Hours today, and came to the conclusion that we can avoid `cardano-api` having to implement this more complicated logic by a simple change in Consensus, see #863 and #862.

carbolymer:
Thanks @amesgen ! I think I prefer to rely on the consensus change and your assumption:
```
In order to reduce the number of possible version combinations, we currently follow the convention that all `ShelleyNodeToClientVersion`s in one `CardanoNodeToClientVersionX` are equal.
```
this way we can just use `NodeToClientVersion` and avoid introducing redundant code.
@Jimbo4350 thoughts?

carbolymer:
Closing because of: https://github.com/IntersectMBO/cardano-api/pull/403#issuecomment-1880960379

PR #424, opened 2024-01-12
Implement checkLedgerStateConditions

Jimbo4350:
# Changelog

```yaml
- description: |
    Implement checkLedgerStateConditions. This new function gives direct access to the `NewEpochState` which contains the ledger state. It also requires an `EpochNo` upon which it will terminate if the supplied condition is not met.
  type:
```

# Context

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/LedgerState.hs
+++ b/cardano-api/internal/Cardano/Api/LedgerState.hs
@@ -39,6 +39,10 @@ module Cardano.Api.LedgerState
   , chainSyncClientWithLedgerState
   , chainSyncClientPipelinedWithLedgerState
 
+   -- * Ledger state conditions
+  , LedgerStateCondition(..)
+  , checkLedgerStateCondition
+
    -- * Errors
   , LedgerStateError(..)
   , FoldBlocksError(..)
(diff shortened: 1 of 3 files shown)


cardano-api/internal/Cardano/Api/LedgerState.hs
```diff
@@ -1708,3 +1719,211 @@ constructGlobals
...
+
+data LedgerStateCondition
+  = ConditionMet
+  | ConditionNotMet
+
+-- | Reconstructs the ledger state and applies a supplied condition to it
+-- for every block. This function only terminations if the condition
+-- is met or we have reached the termination epoch. We need to provide
+-- a termination epoch otherwise we would indefinitely apply blocks
+-- until we have reached the tip of the chain.
```
carbolymer, line 1735:
```suggestion
-- for every block. This function only terminates if the condition
-- is met or we have reached the termination epoch. We need to provide
-- a termination epoch otherwise we would indefinitely apply blocks.
```
Typo. Also I think the last part was redundant, since the tip is moving and we would be applying blocks indefinitely. Do I get this right?

carbolymer, line 1735:
```suggestion
```
I think the last part was redundant, since the tip is moving and we would be applying blocks indefinitely. Do I get this right?

cardano-api/internal/Cardano/Api/LedgerState.hs
```diff
@@ -1708,3 +1719,211 @@ constructGlobals
...
+
+
+--------------------------------------------------------------------------
+
+data LedgerStateCondition
+  = ConditionMet
+  | ConditionNotMet
+
+-- | Reconstructs the ledger state and applies a supplied condition to it
+-- for every block. This function only terminations if the condition
```
carbolymer, line 1732:
```suggestion
-- for every block. This function only terminates if the condition
```
A typo.

carbolymer (approved):

Issue #384, opened 2023-11-23
Remove cardano-api's ProtocolParameters (it has been deprecated for a while)


smelc:
From @Jimbo4350 [here on Slack](https://input-output-rnd.slack.com/archives/C031GFLC3DF/p1700679433867939?thread_ts=1700673271.498199&cid=C031GFLC3DF) (context: https://github.com/input-output-hk/cardano-cli/pull/455 caused https://github.com/input-output-hk/cardano-cli/issues/472 and https://github.com/input-output-hk/cardano-cli/issues/474):

> A potential way forward would be to:
• Define JSON instances for cardano-api's `LedgerProtocolParameters` that are identical to what we have for `ProtocolParameters`. 
• We should be able to begin the deprecation of `cardano-api`'s `ProtocolParameters` . 
• Talk to QA about relying fully on the ledger's `PParams` JSON instances for Conway on wards.

smelc:
cc @mkoura

gitmachtl:
@smelc @mkoura @Jimbo4350 just to give you also some context about external tools. we heavily use the protocol-parameters.json output to use it for the build-raw command and calculate things and switch tool behavior on protocol version...
For example, during the pool registration certificate generation query the parameters, and than look at the current minPoolCost value to validate that the generated certificate is within the allowed boundaries. Later on when it comes to the pool registration, query the parameters and look at the stakePoolDeposit parameter to calculate the right amount for the build-raw command. Also for example i am using my own minOutUTXO value function i created to be independend from the inbuilt one of the cli. Also looking at values like protocolVersion.major, utxoCostPerByte, minUTxOValue, utxoCostPerWord. When it comes to the stake key registration, looking for the stakeAddressDeposit parameter to calculate the correct amounts for the transaction. And so on ... 

All these actions must be possible on complete offline mode without a node running. 

So whatever change will be made in the future, please keep such things in mind. CLI tools like Jormanager, CN-Tools, StakepoolOperatorScripts are building transactions via build-raw to do explicit era transaction (often needed for hw-wallet restrictions) and so on. We are not using the "easier" build command for this. 

Best regards, Martin

Jimbo4350:
Thanks @gitmachtl. Just to confirm you would like us to maintain cardano-api's existing JSON serialization of `ProtocolParameters`

gitmachtl:
[...]

PR #457, opened 2024-02-18
Add golden tests to `ProtocolParameters` serialization

palas:
This PR:
- Adds a golden test for ProtocolParameters
- Adds property based tests for comparing ProtocolParameters and PParams for the different eras
- Fixes the implementation of `toAlonzoPParams`

In order to make the tests work, this PR is built upon a branch of [an old commit](https://github.com/IntersectMBO/cardano-ledger/compare/3c2c45c640b98c57f36e67a0b84baf0119f9407b...3975dc36034238677351958224535a0a992529e0) of `cardano-ledger`.

The rebased version of this branch is proposed in [this PR in cardano-ledger](https://github.com/IntersectMBO/cardano-ledger/pull/4129).

# Changelog

```yaml
- description: |
    Added a golden test for ProtocolParameters, added three property-based tests for comparing ProtocolParameters and PParams for the different eras, and fixed the implementation of `toAlonzoPParams` function.
  type:
```

# Context

This is part of an effort to unify the JSON instances of `ProtocolParameters` and `PParams` to eventually migrate and use `PParams` directly in `cardano-api` instead of `ProtocolParameters`

# How to trust this PR

Check that the properties and golden tests are the right ones. And check that the changes I did to the conversion functions make sense. Possible check together with [this PR in cardano-ledger](https://github.com/IntersectMBO/cardano-ledger/pull/4129).

# Checklist

- [x] Self-reviewed the diff


cardano-api/test/cardano-api-test/Test/Cardano/Api/ProtocolParameters.hs
```diff
@@ -0,0 +1,145 @@
...
+                          [ testProperty "ShelleyEra" $ ppParamsRoundtrip ShelleyEra
+                          , testProperty "AlonzoEra" $ ppParamsRoundtrip AlonzoEra
+                          , testProperty "BabbageEra" $ ppParamsRoundtrip BabbageEra
+                          ]
+              ]
+
+
+-- | Ensure serialization of PParams is the same as that of ProtocolParams
+protocolParametersSerializeTheSame :: forall era. ToJSON (PParams (ShelleyLedgerEra era)) => CardanoEra era -> Property
+protocolParametersSerializeTheSame era =
```
Jimbo4350, line 50: :+1:  A comment explicitly saying we are comparing the JSON serialization of cardano-ledger's `PParams` and cardano-api's `ProtocolParameters` would be useful.

cardano-api/test/cardano-api-test/Test/Cardano/Api/ProtocolParameters.hs
```diff
@@ -0,0 +1,145 @@
...
+    patchProtocolParamsJSON :: LBS.ByteString -> Maybe LBS.ByteString
+    patchProtocolParamsJSON s = LBS.fromStrict . prettyPrintJSON <$> (patchProtocolParamRepresentation =<< decode s)
+      where
+        patchProtocolParamRepresentation :: Object -> Maybe Object
+        patchProtocolParamRepresentation o = do filters <- filtersForEra era
+                                                replace "committeeTermLength" "committeeMaxTermLength"
+                                                  =<< replace "minCommitteeSize" "committeeMinSize"
+                                                        (applyFilters filters o)
+
+        filtersForEra :: CardanoEra era -> Maybe [String]
```
Jimbo4350, line 125: A comment here saying we incorrectly rendered these fields in the JSON serialization of `ProtocolParameters` and now we have to filter them out when comparing against the JSON format of `PParams` would be useful.

cardano-api/test/cardano-api-test/Test/Cardano/Api/ProtocolParameters.hs
```diff
@@ -0,0 +1,145 @@
...
+patchProtocolParamsJSONOrFail :: MonadFail m => CardanoEra era -> LBS.ByteString -> m LBS.ByteString
+patchProtocolParamsJSONOrFail era b = maybe (fail "Cannot fix JSON") return $ patchProtocolParamsJSON b
+  where
+    patchProtocolParamsJSON :: LBS.ByteString -> Maybe LBS.ByteString
+    patchProtocolParamsJSON s = LBS.fromStrict . prettyPrintJSON <$> (patchProtocolParamRepresentation =<< decode s)
+      where
+        patchProtocolParamRepresentation :: Object -> Maybe Object
+        patchProtocolParamRepresentation o = do filters <- filtersForEra era
+                                                replace "committeeTermLength" "committeeMaxTermLength"
+                                                  =<< replace "minCommitteeSize" "committeeMinSize"
```
Jimbo4350, line 122: Is this for Conway?

Jimbo4350 (commented):
Looking good :+1:

cardano-api/test/cardano-api-test/Test/Cardano/Api/ProtocolParameters.hs
```diff
@@ -0,0 +1,104 @@
...
+                                         } <- forAll $ genValidSerializedPair era
+                 case eitherDecode serializedProtocolParams :: Either String (PParams (ShelleyLedgerEra era)) of
+                   Left err -> fail err
+                   Right pParams -> prettyPrintJSON pParams === LBS.toStrict serializedProtocolParams
+
+-------------------------
+-- Auxiliary generator --
+-------------------------
+
+data ValidatedSerializedPair era = ValidatedSerializedPair { serializedProtocolParams ::  LBS.ByteString
```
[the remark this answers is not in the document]
Jimbo4350: Maybe even add haddocks pointing out the types they correspond to.

cardano-api/test/cardano-api-test/Test/Cardano/Api/ProtocolParameters.hs
palas: Yes, it is because of this comment: https://github.com/IntersectMBO/cardano-ledger/pull/4129#discussion_r1507373498

cabal.project
```diff
@@ -26,6 +26,43 @@ package cardano-api
...
+  tag: 3975dc36034238677351958224535a0a992529e0
+  subdir: eras/conway/impl
+  --sha256: sha256-PUZzjs9GvIe2EkV9iGb+Lo+CCVA24x4mQ6oGkEsDPOc=
+
+source-repository-package
+  type: git
+  location: https://github.com/IntersectMBO/cardano-ledger.git
+  tag: 3975dc36034238677351958224535a0a992529e0
+  subdir: libs/cardano-ledger-core
+  --sha256: sha256-PUZzjs9GvIe2EkV9iGb+Lo+CCVA24x4mQ6oGkEsDPOc=
```
newhoggy, line 63: What's the background to this SRP?  Is this a branch a backport.  Is the PR waiting for a ledger release or an upgrade in `cardano-api` to a newer ledger?

cardano-api/test/cardano-api-golden/Test/Golden/Cardano/Api/ProtocolParameters.hs
```diff
@@ -0,0 +1,108 @@
...
+  testGroup "golden ProtocolParameter tests"
+            [ testProperty "ShelleyPParams" $
+                goldenProtocolParametersToPParams (undefined :: (ShelleyPParams Identity (ShelleyEra StandardCrypto)))
+            , testProperty "AlonzoPParams" $
+                goldenProtocolParametersToPParams (undefined :: (AlonzoPParams Identity (AlonzoEra StandardCrypto)))
+            , testProperty "BabbagePParams" $
+                goldenProtocolParametersToPParams (undefined :: (BabbagePParams Identity (BabbageEra StandardCrypto)))
+            ]
+
+goldenProtocolParametersToPParams :: forall pp. FromJSON pp => pp -> Property
```
newhoggy, line 50: Perhaps we can use `Proxy pp` instead of `pp` so we can avoid using `undefined` above?

cardano-api/test/cardano-api-golden/Test/Golden/Cardano/Api/ProtocolParameters.hs
```diff
@@ -0,0 +1,108 @@
...
+
+import           Hedgehog (Property, property, success)
+import qualified Hedgehog.Extras.Aeson as H
+import           Hedgehog.Internal.Property (failWith)
+import           Test.Tasty (TestTree, testGroup)
+import           Test.Tasty.Hedgehog (testProperty)
+
+test_golden_ProtocolParameters :: TestTree
+test_golden_ProtocolParameters = testProperty "golden ProtocolParameters" $ do
+  H.goldenTestJsonValuePretty legacyCardanoApiProtocolParameters "test/cardano-api-golden/files/golden/LegacyProtocolParameters"
```
newhoggy, line 37: Should this golden file have a `.json` file extension?

cardano-api/test/cardano-api-test/Test/Cardano/Api/ProtocolParameters.hs
```diff
@@ -0,0 +1,155 @@
...
+protocolParametersAreCompatible era =
+   property $ do ValidatedSerializedPair { serializedProtocolParameters
+                                         , serializedPParams = _
+                                         } <- forAll $ genValidSerializedPair era
+                 case eitherDecode serializedProtocolParameters :: Either String (PParams (ShelleyLedgerEra era)) of
+                   Left err -> fail err
+                   Right _ -> success
+
+-- | Ensure that deserializing using PParams FromJSON instance and then serializing again using PParams ToJSON
+-- | instance results in the same thing
```
newhoggy, line 72: My understanding is only the first line in haddock comments needs a `|`?

cardano-api/test/cardano-api-test/Test/Cardano/Api/ProtocolParameters.hs
```diff
@@ -0,0 +1,155 @@
...
+        patchProtocolParamRepresentation o = do filters <- filtersForEra era
+                                                replace "committeeTermLength" "committeeMaxTermLength"
+                                                  =<< replace "minCommitteeSize" "committeeMinSize"
+                                                        (applyFilters filters o)
+
+        -- | Legacy ProtocolParams ToJSON renders all fields from all eras in all eras,
+        -- | because it is the same data type for every era. But this is not backwards compatible
+        -- | because it means that new eras can modify the fields in old eras. For this reason, when
+        -- | comparing to PParams we use this function to filter fields that don't belong to
+        -- | particular era we are testing.
```
newhoggy, line 134: This function isn't top-level.  Does it's comment actually appear in the generated documentation?  If not we can drop the `|`.

cabal.project
palas: Oh, apparently the later, I didn't realise, good point. I could try to do that

cardano-api/test/cardano-api-test/Test/Cardano/Api/ProtocolParameters.hs
palas: I was wondering when I wrote it, but couldn't find out

cardano-api/test/cardano-api-golden/Test/Golden/Cardano/Api/ProtocolParameters.hs
```diff
@@ -0,0 +1,109 @@
...
+
+import           Hedgehog (Property, property, success)
+import qualified Hedgehog.Extras.Aeson as H
+import           Hedgehog.Internal.Property (failWith)
+import           Test.Tasty (TestTree, testGroup)
+import           Test.Tasty.Hedgehog (testProperty)
+
+test_golden_ProtocolParameters :: TestTree
+test_golden_ProtocolParameters = testProperty "golden ProtocolParameters" $ do
+  H.goldenTestJsonValuePretty legacyCardanoApiProtocolParameters "test/cardano-api-golden/files/golden/LegacyProtocolParameters.json"
```
carbolymer, line 38: Something is off here. This is supposed to produce pretty json, but it appears it's not doing that?

cardano-api/test/cardano-api-test/Test/Cardano/Api/ProtocolParameters.hs
```diff
@@ -0,0 +1,155 @@
...
+-- | Ensure that cardano-api's legacy ProtocolParameter serialization can be deserialized by cardano-ledger's PParams FromJSON instance
+protocolParametersAreCompatible :: forall era. ( ToJSON (PParams (ShelleyLedgerEra era))
+                                               , FromJSON (PParams (ShelleyLedgerEra era))
+                                               ) => CardanoEra era -> Property
+protocolParametersAreCompatible era =
+   property $ do ValidatedSerializedPair { serializedProtocolParameters
+                                         , serializedPParams = _
+                                         } <- forAll $ genValidSerializedPair era
+                 case eitherDecode serializedProtocolParameters :: Either String (PParams (ShelleyLedgerEra era)) of
+                   Left err -> fail err
```
carbolymer, line 68: I think[ `void . leftFail`](https://hackage.haskell.org/package/hedgehog-extras-0.6.1.0/docs/Hedgehog-Extras-Test-Base.html#v:leftFail) could work here

cardano-api/test/cardano-api-test/Test/Cardano/Api/ProtocolParameters.hs
```diff
@@ -0,0 +1,155 @@
...
+                                        , "maxValueSize", "utxoCostPerByte" ]
+        filtersForEra AlonzoEra  = Just [ "minUTxOValue" ]
+        filtersForEra BabbageEra = Just [ "decentralization", "extraPraosEntropy", "minUTxOValue" ]
+        filtersForEra _ = Nothing
+
+        applyFilters :: [String] -> Object -> Object
+        applyFilters filters o = foldl' (flip Aeson.delete) o (map Aeson.fromString filters)
+
+        replace :: String -> String -> Object -> Maybe Object
+        replace src dest o =
```
carbolymer, line 147:
```suggestion
        renameKey :: String -> String -> Object -> Maybe Object
        renameKey srcKey destKey o =
```
took me a minute to decipher what this function does 😄

cardano-api/test/cardano-api-golden/Test/Golden/Cardano/Api/ProtocolParameters.hs
palas: It is not "pretty", but is not single line either. I think it is because there are many elements in the list and it is not prepared for that. Or do you mean the fact that CREATE_GOLDEN_TESTS doesn't work? I checked and that is hardcoded :/

carbolymer (approved):
LGTM, but get an approval from @Jimbo4350 too

cardano-api/test/cardano-api-golden/Test/Golden/Cardano/Api/ProtocolParameters.hs
```diff
@@ -0,0 +1,109 @@
...
+  testGroup "golden ProtocolParameter tests"
+            [ testProperty "ShelleyPParams" $
+                goldenProtocolParametersToPParams (Proxy :: Proxy (ShelleyPParams Identity (ShelleyEra StandardCrypto)))
+            , testProperty "AlonzoPParams" $
+                goldenProtocolParametersToPParams (Proxy :: Proxy (AlonzoPParams Identity (AlonzoEra StandardCrypto)))
+            , testProperty "BabbagePParams" $
+                goldenProtocolParametersToPParams (Proxy :: Proxy (BabbagePParams Identity (BabbageEra StandardCrypto)))
+            ]
+
+goldenProtocolParametersToPParams :: forall pp. FromJSON pp => Proxy pp -> Property
```
Jimbo4350, line 51: `goldenLegacyProtocolParametersToPParams` ?

cardano-api/test/cardano-api-golden/files/golden/LegacyProtocolParameters.json
```diff
@@ -0,0 +1 @@
...14,115,116,117,118,119,120,121,122,123,124,125,126,127,128,129,130,131,132,133,134,135,136,137,138,139,140,141,142,143,144,145,146,147,148,149,150,151,152,153,154,155,156,157,158,159,160,161,162,163,164,165,166,167,168,169,170,171,172,173,174,175,176,177,178,179,180,181,182,183,184,185,186,187,188,189,190,191,192,193,194,195,196,197,198,199,200,201,202,203,204,205,206,207,208,209,210,211,212,213,214,215,216,217,218,219,220,221,222,223,224,225,226,227,228,229,230,231,232,233]},"decentralization":0.52,"executionUnitPrices":{"priceMemory":0.2,"priceSteps":0.3},"extraPraosEntropy":"853a190a0ee6ca25de8b22f69624799a8db530d561dce9183029fe7a2ed29c6f","maxBlockBodySize":5000,"maxBlockExecutionUnits":{"memory":3400,"steps":5600},"maxBlockHeaderSize":1200,"maxCollateralInputs":10,"maxTxExecutionUnits":{"memory":2300,"steps":4300},"maxTxSize":3000,"maxValueSize":10,"minPoolCost":3500000,"minUTxOValue":3000000,"monetaryExpansion":0.23,"poolPledgeInfluence":0.54,"poolRetireMaxEpoch":4,"protocolVersion":{"major":2,"minor":3},"stakeAddressDeposit":10000000,"stakePoolDeposit":1000000000,"stakePoolTargetNum":100,"treasuryCut":0.1,"txFeeFixed":1500000,"txFeePerByte":2000000,"utxoCostPerByte":1000000}
```
Jimbo4350, line 1: Where is "PlutusV1"?

cardano-api/test/cardano-api-test/Test/Cardano/Api/ProtocolParameters.hs
```diff
@@ -0,0 +1,155 @@
...
+  , -- | Serialized cardano-ledger's `PParams` as a ByteString.
+    serializedPParams            :: LBS.ByteString
+  } deriving Show
+
+
+-- | Produces a pair of a valid cardano-api's legacy ProtocolParameters and corresponding cardano-ledger's PParams by doing a round trip
+genValidSerializedPair :: forall era. ToJSON (PParams (ShelleyLedgerEra era)) => CardanoEra era -> Gen (ValidatedSerializedPair era)
+genValidSerializedPair era = do
+  unrefinedProtocolParameters <- genProtocolParameters era
+  case (do unrefinedPParams <- convertToLedgerProtocolParameters sbe unrefinedProtocolParameters :: (Either ProtocolParametersConversionError (LedgerProtocolParameters era))
```
Jimbo4350, line 102: This is my first time seeing `do` notation inside a `case` expression like this. Can you factor out the expression inside the `do` notation and return a variable for the result which can then be `case`d on?

cardano-api/test/cardano-api-test/Test/Cardano/Api/ProtocolParameters.hs
```diff
@@ -0,0 +1,155 @@
...
+           return $ ValidatedSerializedPair { serializedProtocolParameters = LBS.fromStrict $ prettyPrintJSON refinedProtocolParams
+                                            , serializedPParams = LBS.fromStrict $ prettyPrintJSON . unLedgerProtocolParameters $ refinedPParams
+                                            }) of
+     Right result -> return result
+     Left _ -> genValidSerializedPair era
+    where
+      sbe :: ShelleyBasedEra era
+      sbe = toShelleyBased era
+
+      toShelleyBased :: CardanoEra era -> ShelleyBasedEra era
```
Jimbo4350, line 114: `genProtocolParameters` should not be parameterized on `CardanoEra era`. We need to sort that out in a separate PR.

cardano-api/test/cardano-api-test/Test/Cardano/Api/ProtocolParameters.hs
```diff
@@ -0,0 +1,155 @@
...
+      toShelleyBased = inEonForEra (error "Not a Shelley-based era") id
+
+
+patchProtocolParamsJSONOrFail :: MonadFail m => CardanoEra era -> LBS.ByteString -> m LBS.ByteString
+patchProtocolParamsJSONOrFail era b = maybe (fail "Cannot fix JSON") return $ patchProtocolParamsJSON b
```
Jimbo4350, line 119: If this fails and all we see is "Cannot fix JSON" it's not going to be obvious what went wrong. Why are we "fixing" the JSON?

cardano-api/test/cardano-api-golden/Test/Golden/Cardano/Api/ProtocolParameters.hs
```diff
@@ -0,0 +1,109 @@
...
+goldenProtocolParametersToPParams _ =
+  property $ case decodedLegacyCardanoApiProtocolParameters of
+               Left err -> failWith Nothing ("could not decode: " <> show err)
```
Jimbo4350, line 54: "goldenProtocolParametersToPParams.could not decode: " <> show err

cardano-api/test/cardano-api-test/Test/Cardano/Api/ProtocolParameters.hs
```diff
@@ -0,0 +1,155 @@
...
+import qualified Data.ByteString.Lazy as LBS
+import           Data.Foldable (foldl')
+
+import           Test.Gen.Cardano.Api.Typed (genProtocolParameters)
+
+import           Hedgehog (Gen, Property, forAll, property, success, (===))
+import           Test.Tasty (TestTree, testGroup)
+import           Test.Tasty.Hedgehog (testProperty)
+
+tests :: TestTree
```
Jimbo4350, line 29: One additional comment here. We should indicate the reason why we don't test Conway on wards is that we fully adopt ledger's `PParams` type.

cardano-api/test/cardano-api-test/Test/Cardano/Api/ProtocolParameters.hs
Jimbo4350: Add a comment explaining the reason for the change ^^

Jimbo4350 (commented):
Looking good but I left some additional comments. It may come off as nit picking but I want this code to be as accessible as possible. Leaving comments explaining the obvious will save us headache in the future if we have to come back to this code.

Jimbo4350 (approved):
LGTM! However you need to fix the cost model golden file and include the "PlutusV1" cost model.

cardano-api/test/cardano-api-test/Test/Cardano/Api/ProtocolParameters.hs
```diff
@@ -0,0 +1,155 @@
...
+                                               ) => CardanoEra era -> Property
+protocolParametersAreCompatible era =
+   property $ do ValidatedSerializedPair { serializedProtocolParameters
+                                         , serializedPParams = _
+                                         } <- forAll $ genValidSerializedPair era
+                 case eitherDecode serializedProtocolParameters :: Either String (PParams (ShelleyLedgerEra era)) of
+                   Left err -> fail err
+                   Right _ -> success
+
+-- | Ensure that deserializing using PParams FromJSON instance and then serializing again using PParams ToJSON
```
Jimbo4350, line 71: This comment is misleading. You are round tripping between `ProtocolParameters` and `PParams era`.

cardano-api/test/cardano-api-test/Test/Cardano/Api/ProtocolParameters.hs
palas: Hm, actually, I am not sure, maybe it makes sense, because the protocol parameters are different for different eras, so they get validated differently

cardano-api/test/cardano-api-test/Test/Cardano/Api/ProtocolParameters.hs
palas: That is definitely a better name, I also added a small comment

cardano-api/test/cardano-api-golden/Test/Golden/Cardano/Api/ProtocolParameters.hs
palas: Ah, you were right, I used the wrong function. Fixed now

cardano-api/test/cardano-api-golden/files/golden/LegacyProtocolParameters.json
palas: Added it now

cardano-api/test/cardano-api-test/Test/Cardano/Api/ProtocolParameters.hs
palas: I've added this explanation:
```text
Originally, cardano-api used a different type than cardano-ledger to represent
protocol parameters. From conway on, we aim to unify those types and use PParams.
These tests aim to ensure backwards compatibility between the legacy type ProtocolParams
and PParams for eras before conway. Conway should use PParams directly, so we don't
provide any tests for it.
```

smelc:
Contributes to fixing https://github.com/IntersectMBO/cardano-api/issues/384

PR #458, opened 2024-02-20
Improve Github Actions

erikd:
# Changelog

```yaml
- description: Improve Github Actions
  type:
```

# Context

Aiming to reduce the size of the GHA spec and to use the DEXV GHA helpers.

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -59,9 +59,12 @@ jobs:
     - name: Cabal update
       run: cabal update
 
+    # A dry run `build all` operation does *NOT* downlaod anything, it just looks at the package
+    # indices to generate an install plan.
     - name: Build dry run
       run: cabal build all --enable-tests --dry-run --minimize-conflict-set
 
+    # From the install plan we generate a dependency list.
     - name: Record dependencies
       id: record-deps
       run: |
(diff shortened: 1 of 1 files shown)


erikd:
Had a very quick look at the build speed. I found a recent `ghc-9.8` build on `Linux` that took over 26 minutes while a the same build configuration with the new caching took 23 minutes.

.github/workflows/haskell.yml
```diff
@@ -60,49 +60,29 @@ jobs:
...
-        dist-dir: dist-newstyle
-        store-path: ${{ steps.setup-haskell.outputs.cabal-store }}
-        threads: 16
-        archive-uri: https://iohk.cache.haskellworks.io/${{ env.CABAL_CACHE_VERSION }}/${{ runner.os }}/${{ matrix.cabal }}/${{ matrix.ghc }}
-        skip: "${{ vars.CACHE_URI != '' }}"
-        enable-save: false
+        path: |
+          ${{ steps.setup-haskell.outputs.cabal-store }}
+          dist-newstyle
+        key: cache-${{ env.CABAL_CACHE_VERSION }}-${{ runner.os }}-${{ matrix.ghc }}-${{ hashFiles('dependencies.txt') }}
```
smelc, line 82: > hashFiles('dependencies.txt')

I see that previously the cache wasn't indexed by the dependencies. So is there a chance this PR will reduce the occurrences of the cabal cache being corrupted (i.e. less bumps of CABAL_CACHE_VERSION required)?

smelc:
Seems like a sane simplification :+1:

.github/workflows/haskell.yml
```diff
@@ -60,49 +60,29 @@ jobs:
...
-      uses: action-works/cabal-cache-s3@v1
+    - name: Record dependencies
+      id: record-deps
+      run: |
+        # The tests call out to msys2 commands. We generally do not want to mix toolchains, so
+        # we are very deliberate about only adding msys64 to the path where absolutely necessary.
+        ${{ (runner.os == 'Windows' && '$env:PATH=("C:\msys64\mingw64\bin;{0}" -f $env:PATH)') || '' }}
+        cat dist-newstyle/cache/plan.json | jq -r '."install-plan"[].id' | sort | uniq > dependencies.txt
+
+    - name: Install dependencies
```
Jimbo4350, line 73: Silly question but don't we need to run the `Cache Cabal store` job first before running  `cabal build all --enable-tests --only-dependencies`?

.github/workflows/haskell.yml
```diff
@@ -60,49 +60,29 @@ jobs:
...
+      run: cabal build all --enable-tests --only-dependencies -j --ghc-option=-j4
+
+    - name: Cache Cabal store
+      uses: actions/cache@v3
```
carbolymer, line 77: Any reason not to use `v4`?

.github/workflows/haskell.yml
erikd: No, specifically want to run it after. We want to build and then store.

Jimbo4350 (approved):

erikd:
As @Jimbo4350 pointed out, there was no action to restore the cabal cache. Fixing that now.

.github/workflows/haskell.yml
erikd: Thanks foe making me look more closely. That made me realise that there is a cache store, but no **restore** action.

angerman (approved):
So the control flow is:
- dry-run build to get the proper dependency lists.
- use dependencies list as primary key for the cache.
- RESTORE the cache.
- BUILD DEPENDENCIES.
- STORE the cache.
- BUILD
- TEST
- ARTIFACTS.

Looks good to me.

PR #481, opened 2024-03-11
Exclude non-global dependencies from cache-key computation

newhoggy:
# Changelog

```yaml
- description: |
    Exclude non-global dependencies from cache-key computation
  type:
```

# Context

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff


carbolymer (approved):
Nice, thanks! Can you add this to cardano-cli as well?

smelc:
@newhoggy> there are a number of entries in the build plan that don't have the `style` field set. Wouldn't it be safer to remove entries that have the field set to `local` rather than filtering out those that don't have the field set to `global` (so also excluding those that don't have the field set).

Here is an excerpt of the difference (filtering by `.style != "local"` on the left, filterd by `.style == "global"` on the right): 
image

newhoggy:
Thanks for the suggestion.  I've switch to `.style != "local"`.

PR #498, opened 2024-03-26
Replace unsafeMergeVotingProcedures by mergeVotingProcedures

smelc:
# Changelog

```yaml
- description: |
    Replace unsafeMergeVotingProcedures by mergeVotingProcedures, that handles incompatible votes and return an error
  type:
```

# Context

* This is the cardano-api part of https://github.com/IntersectMBO/cardano-cli/pull/681
* The return type for the error case is a bit different from the version in https://github.com/IntersectMBO/cardano-cli/pull/681. This version makes more sense for general-purpose callers and we can adapt the CLI's caller when we'll update.

# How to trust this PR

See https://github.com/IntersectMBO/cardano-cli/pull/681

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/Governance/Actions/VotingProcedure.hs
+++ b/cardano-api/internal/Cardano/Api/Governance/Actions/VotingProcedure.hs
@@ -30,7 +30,9 @@ import qualified Cardano.Ledger.Api as L
 import           Cardano.Ledger.Core (EraCrypto)
 import qualified Cardano.Ledger.Core as L
 
+import           Control.Monad (foldM)
 import qualified Data.Map as Map
+import qualified Data.Set as Set
 import           Data.Text (Text)
 import qualified Data.Text.Encoding as Text
 import           GHC.Generics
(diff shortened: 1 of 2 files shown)


cardano-api/internal/Cardano/Api/Governance/Actions/VotingProcedure.hs
```diff
@@ -154,16 +156,28 @@ singletonVotingProcedures _ voter govActionId votingProcedure =
...
+            , [L.GovActionId (L.EraCrypto (ShelleyLedgerEra era))]) -- and its conflicting votes (multiples votes for the same governance action);
+            (VotingProcedures era)                                  -- or the merged votes.
+mergeVotingProcedures vpsa vpsb =
+  VotingProcedures . L.VotingProcedures <$> foldM mergeVotesOfOneVoter Map.empty allVoters
+  where
+    mapa = L.unVotingProcedures (unVotingProcedures vpsa)
+    mapb = L.unVotingProcedures (unVotingProcedures vpsb)
+    allVoters = Set.union (Map.keysSet mapa) (Map.keysSet mapb)
+    mergeVotesOfOneVoter acc voter =
+      Map.union acc <$> case (Map.lookup voter mapa, Map.lookup voter mapb) of
```
Jimbo4350, line 174: Why not use [differenceWith](https://hackage.haskell.org/package/containers-0.7/docs/Data-Map-Strict.html#v:differenceWith) since  all we care about are conflicts? Then we can just do a `union` on the maps if there are none.

smelc: @Jimbo4350> but we need to report the conflicts. How would we do that with `differenceWith`, that is not within an error monad/`Either`-return type?

Jimbo4350: https://github.com/IntersectMBO/cardano-api/commit/04b75f6be6157c401da222ab368e74782ae02771

smelc: > I think we should also property test this function.

If you're fine with me spending time doing that, yes, :100: This function has nice properties indeed.

Regarding https://github.com/IntersectMBO/cardano-api/commit/ae88ae4915991d6e49dc77f25dbbd91992939e1f, sorry I disagree: my version doesn't separate doing the job from checking conflicts, it's doing both at the same time; which is more intuitive, at the cost of very little duplication (only the symmetry of `(Just _, Nothing)` causes repetition). I rather keep my version for future readers and maintainers of the code.

smelc: I added a `newtype` wrapping the error case and tracked adding properties tests in https://github.com/IntersectMBO/cardano-api/issues/499/, so I'm enqueueing for merge.

newhoggy: `differenceWith` is only able to compare the keys.

If we have to `GovActionId`s that are equal according to `Eq`, is that considered a conflict?

Jimbo4350 (approved):

cardano-api/internal/Cardano/Api/Governance/Actions/VotingProcedure.hs
smelc: @newhoggy> indeed we only compare the keys (`GovActionId`). So, if there are two equal `GovActionId` that map to the same `VotingProcedures`, a conflict will be reported; whereas it's not really a conflict, it's more of a duplication.

PR #511, opened 2024-04-08
Implement estimateBalancedTxBody and estimateOrCalculateBalancedTxBody

Jimbo4350:
# Changelog

```yaml
- description: |
    Implement estimateBalancedTxBody and estimateOrCalculateBalancedTxBody
  type:
```

# Context

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff


cardano-api/internal/Cardano/Api/Fees.hs
carbolymer, line 1: Can you add `Show` and `Error` instances for error types?

cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -87,6 +93,159 @@ import           Lens.Micro ((.~), (^.))
...
+        )
+
+  
+data TxFeeEstimationError era
+  = TxFeeEstimationTransactionTranslationError (TransactionValidityError era)
+  | TxFeeEstimationScriptExecutionError (TxBodyErrorAutoBalance era)
+  | TxFeeEstimationBalanceError (TxBodyErrorAutoBalance era)
+  | TxFeeEstimationxBodyError TxBodyError
+  | TxFeeEstimationFinalConstructionError TxBodyError
+  | TxFeeEstimationOnlyMaryOnwardsSupported
```
carbolymer, line 134:
```suggestion
  | TxFeeEstimationOnlyMaryOnwardsSupportedError
```

carbolymer (approved):

cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -679,6 +838,28 @@ data BalancedTxBody era
...
+      -- | Total value of UTxOs being spent
+      Value 
+      -- | Plutus script execution units
+      (Map ScriptWitnessIndex ExecutionUnits) 
+      -- | The number of key witnesses still to be added to the transaction.
+      Int
+      -- | The number of Byron key witnesses still to be added to the transaction.
+      Int
+      -- | The total size in bytes of reference scripts
+      Int
```
newhoggy, line 861: Might it be worth having having these sum-type constructors wrap record types so they can have field names?

cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -29,9 +29,14 @@ module Cardano.Api.Fees (
     evaluateTransactionBalance,
 
     -- * Automated transaction building
+    estimateBalancedTxBody,
+    estimateOrCalculateBalancedTxBody,
```
newhoggy, line 33: I'm thinking it would be good to have unit tests or golden tests for these new functions.

Even if it we have very basic ones to start with.

cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -786,17 +965,22 @@ makeTransactionBodyAutoBalance sbe systemstart history lpp@(LedgerProtocolParame
                  txReturnCollateral = dummyCollRet,
                  txTotalCollateral = dummyTotColl
                }
-
+        -- NB: This has the potential to over estimate the fees!!
```
newhoggy, line 968: I'm unsure if this comment applies to the above or below code.

Extra newline would help.

cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -786,17 +965,22 @@ makeTransactionBodyAutoBalance sbe systemstart history lpp@(LedgerProtocolParame
...
             (const (TxReturnCollateralNone, TxTotalCollateralNone))
             (\w ->
-              calcReturnAndTotalCollateral w
-                fee pp (txInsCollateral txbodycontent) (txReturnCollateral txbodycontent)
-                (txTotalCollateral txbodycontent) changeaddr utxo
+              let collIns = case txInsCollateral txbodycontent of
+                              TxInsCollateral _ collIns' -> collIns'
+                              TxInsCollateralNone -> mempty
+                  collateralOuts  = catMaybes [ Map.lookup txin (unUTxO utxo) | txin <- collIns]
+                  totalPotentialCollateral = mconcat $ map (\(TxOut _ txOutVal _ _) -> txOutValueToLovelace txOutVal) collateralOuts
```
newhoggy, line 980: Would be good to add a comment about how `utxo` differs from `totalPotentialCollateral` that we should use the latter instead.

cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -841,66 +1025,99 @@ makeTransactionBodyAutoBalance sbe systemstart history lpp@(LedgerProtocolParame
...
-
-        case (txReturnCollateral, txTotalCollateral) of
+   era :: CardanoEra era
+   era = shelleyBasedToCardanoEra sbe
+
+-- In the event of spending the exact amount of lovelace in
+-- the specified input(s), this function excludes the change
+-- output. Note that this does not save any fees because by default
+-- the fee calculation includes a change address for simplicity and
+-- we make no attempt to recalculate the tx fee without a change address.
```
newhoggy, line 1035: Can this be upgraded this to a haddock comment?

cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -841,66 +1025,99 @@ makeTransactionBodyAutoBalance sbe systemstart history lpp@(LedgerProtocolParame
...
+     else Left $ TxBodyErrorMinUTxONotMet (txOutInAnyEra (toCardanoEra sbe) txout) minUTxO
+
+balanceCheck
+  :: ShelleyBasedEra era
+  -> Ledger.PParams (ShelleyLedgerEra era)
+  -> AddressInEra era
+  -> TxOutValue era
+  -> Either (TxBodyErrorAutoBalance era) ()
+balanceCheck sbe bpparams changeaddr balance
+ | txOutValueToLovelace balance == 0 && onlyAda (txOutValueToValue balance) = return ()
```
newhoggy, line 1062: What's the reason we can avoid `checkMinUTxOValue` if `Value` only contains ADA?

cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -841,66 +1025,99 @@ makeTransactionBodyAutoBalance sbe systemstart history lpp@(LedgerProtocolParame
...
+   -> (TxReturnCollateral CtxTx era, TxTotalCollateral era)
+calcReturnAndTotalCollateral _ _ _ TxInsCollateralNone _ _ _ _= (TxReturnCollateralNone, TxTotalCollateralNone)
+calcReturnAndTotalCollateral _ _ _ _ rc@TxReturnCollateral{} tc@TxTotalCollateral{} _ _ = (rc,tc)
+calcReturnAndTotalCollateral retColSup fee pp' TxInsCollateral{} txReturnCollateral txTotalCollateral cAddr totalAvailableAda =
+   do
+     let colPerc = pp' ^. Ledger.ppCollateralPercentageL
+     -- We must first figure out how much lovelace we have committed
+     -- as collateral and we must determine if we have enough lovelace at our
+     -- collateral tx inputs to cover the tx
+     let --txOuts = catMaybes [ Map.lookup txin utxo' | txin <- collIns]
```
newhoggy, line 1108: There is commented out code above.

cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -841,66 +1025,99 @@ makeTransactionBodyAutoBalance sbe systemstart history lpp@(LedgerProtocolParame
...
+         totalCollateralLovelace = totalAvailableAda --mconcat $ map (\(TxOut _ txOutVal _ _) -> txOutValueToLovelace txOutVal) txOuts
```
newhoggy, line 1109: There is commented out code above.

cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -841,66 +1025,99 @@ makeTransactionBodyAutoBalance sbe systemstart history lpp@(LedgerProtocolParame
...
+         requiredCollateral@(L.Coin reqAmt) = fromIntegral colPerc * fee
+         totalCollateral = TxTotalCollateral retColSup . L.rationalToCoinViaCeiling
+                                                       $ reqAmt % 100
+         -- Why * 100? requiredCollateral is the product of the collateral percentage and the tx fee
+         -- We choose to multiply 100 rather than divide by 100 to make the calculation
+         -- easier to manage. At the end of the calculation we then use % 100 to perform our division
+         -- and round up.
```
newhoggy, line 1116: Where does the round up happen?  `rationalToCoinViaFloor` takes the floor, so I assume that's a round down?

cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -916,55 +1133,72 @@ makeTransactionBodyAutoBalance sbe systemstart history lpp@(LedgerProtocolParame
...
+          TxMintNone -> mempty
+          TxMintValue _ v _ -> v
+    in mconcat [incoming, minted, negateValue outgoing]
+
+-- | This is used in the balance calculation in the event where
+-- the user does not supply the UTxO(s) they intend to spend
+-- but they must supply their total balance of ADA.
+-- TODO: Include multiassets
+createFakeUTxO :: ShelleyBasedEra era -> TxBodyContent BuildTx era -> Coin -> UTxO era
+createFakeUTxO sbe txbodycontent totalAdaInUTxO =
```
newhoggy, line 1161: I don't understand the word "fake".  I think that choice of word requires elaboration.

I think the function documentation should say what the function does rather than (or at least in addition to) what it is used for.

newhoggy: Is this like a placeholder utxo that later gets updated with a proper one?

cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -916,55 +1133,72 @@ makeTransactionBodyAutoBalance sbe systemstart history lpp@(LedgerProtocolParame
...
+  in UTxO $ Map.fromList $ zip singleTxIn singleTxOut
+
+updateTxOut :: ShelleyBasedEra era -> Coin -> TxOut CtxUTxO era -> TxOut CtxUTxO era
+updateTxOut sbe updatedValue txout =
+  let ledgerout = shelleyBasedEraConstraints sbe $ toShelleyTxOut sbe txout & L.coinTxOutL .~ updatedValue
+  in fromShelleyTxOut sbe ledgerout
+
+-- Essentially we check for the existence of collateral inputs. If they exist we
+-- create a fictitious collateral return output. Why? Because we need to put dummy values
+-- to get a fee estimate (i.e we overestimate the fee.)
```
newhoggy, line 1175: So to clarify we return dummy collateral return output in the case where collateral inputs existed but collateral return outputs are not provided?

cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -841,66 +1025,99 @@ makeTransactionBodyAutoBalance sbe systemstart history lpp@(LedgerProtocolParame
...
+     let --txOuts = catMaybes [ Map.lookup txin utxo' | txin <- collIns]
+         totalCollateralLovelace = totalAvailableAda --mconcat $ map (\(TxOut _ txOutVal _ _) -> txOutValueToLovelace txOutVal) txOuts
+         requiredCollateral@(L.Coin reqAmt) = fromIntegral colPerc * fee
+         totalCollateral = TxTotalCollateral retColSup . L.rationalToCoinViaCeiling
+                                                       $ reqAmt % 100
+         -- Why * 100? requiredCollateral is the product of the collateral percentage and the tx fee
+         -- We choose to multiply 100 rather than divide by 100 to make the calculation
+         -- easier to manage. At the end of the calculation we then use % 100 to perform our division
+         -- and round up.
+         enoughCollateral = totalCollateralLovelace * 100 >= requiredCollateral
```
newhoggy, line 1117: Where is `enoughCollateral` used?

newhoggy, line 1117: I'm thinking it would be good to move `enoughCollateral` closer to where it is used down below.

cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -679,6 +838,28 @@ data BalancedTxBody era
       (TxOut CtxTx era) -- ^ Transaction balance (change output)
       L.Coin    -- ^ Estimated transaction fee
 
+data FeeEstimationMode era
+  = CalculateWithSpendableUTxO -- ^ Accurate fee calculation.
+      -- | Spendable UTxO
+      (UTxO era)
+      SystemStart
+      LedgerEpochInfo
+      -- | Override key witnesses
```
newhoggy, line 847: Override number of key witnesses?

cardano-api/internal/Cardano/Api/Fees.hs
Jimbo4350: Definitely

cardano-api/internal/Cardano/Api/Fees.hs
Jimbo4350: I'll update the comment.

cardano-api/internal/Cardano/Api/Fees.hs
Jimbo4350: I'm not sure I understand what comment to put here. `totalPotentialCollateral` comes from the specified collateral outputs. Can you suggest something?

cardano-api/internal/Cardano/Api/Fees.hs
Jimbo4350: If the lovelace component of the `balance` is 0 and the `balance` only consists of lovelace then there is no balance and therefore no need to check for a minimum output value.

cardano-api/internal/Cardano/Api/Fees.hs
Jimbo4350: So the _returnCollateral_ is floored. This means that the `requiredCollateral` is rounded upwards.

cardano-api/internal/Cardano/Api/Fees.hs
Jimbo4350: So `evaluateTransactionBalance` calls [evalBalanceTxBody](https://github.com/IntersectMBO/cardano-ledger/blob/315e78d1baed5efcd18206f3d5113ffefc81de74/libs/cardano-ledger-api/src/Cardano/Ledger/Api/Tx/Body.hs#L96) which required a `UTxO` value. However this eventually calls [getConsumedMaryValue](https://github.com/IntersectMBO/cardano-ledger/blob/315e78d1baed5efcd18206f3d5113ffefc81de74/eras/mary/impl/src/Cardano/Ledger/Mary/UTxO.hs#L70) which retrieves the balance from the transaction itself. 

So I had to create a "fake" UTxO to still use `evalBalanceTxBody` however this will fail for transactions containing multi-assets, refunds and withdrawals. I need to add a comment here and a disclaimer in the cli for the time being that this will balance ADA only transactions for now. I wanted to do some more testing before adding in support for these others things.

cardano-api/internal/Cardano/Api/Fees.hs
Jimbo4350: Because the required collateral depends on the tx fee, we need to put in dummy collateral values otherwise we will risk underestimating the tx fee.

cardano-api/internal/Cardano/Api/Fees.hs
newhoggy: Oh I see, the code above was previously inside `calcReturnAndTotalCollateral` and has been moved out.

cardano-api/internal/Cardano/Api/Fees.hs
newhoggy: What effect does `balance = 0` have on the behaviour of `estimateOrCalculateBalancedTxBody`?

What happens if a transaction with `balance = 0` is submitted?

newhoggy: Oh I understand now.  So if we produce a tx out, it must have minimum ADA, but if we don't produce a txout in the first place we don't ned.

Okay all good.

cardano-api/internal/Cardano/Api/Fees.hs
newhoggy: Is this a problem with how `evalBalanceTxBody` is defined and can `evalBalanceTxBody` be rewritten to accomodate this?

I think your explanation is good and would be useful in the comments.

cardano-api/internal/Cardano/Api/Fees.hs
newhoggy: Can you mention that the required collateral depends on the tx fee in the comments as well?

newhoggy (approved):

cardano-api/internal/Cardano/Api/Fees.hs
Jimbo4350: > Is this a problem with how evalBalanceTxBody is defined and can evalBalanceTxBody be rewritten to accomodate this?

I want to raise this with Alexey.

PR #517, opened 2024-04-10
Sort metadata keys for no-schema json for canonical CBOR

carbolymer:
# Changelog

```yaml
- description: |
    Sort metadata keys for no-schema json for canonical CBOR
  type:
```

# Context

This PR changes sorting for `--json-metadata-no-schema` JSON to [canonical CBOR](https://datatracker.ietf.org/doc/html/rfc7049#section-3.9). It changes transaction file output CBOR representation (reorders metadata keys for this case), so it's marked as a breaking change.

Note that there's always `--json-metadata-schema` JSON format available which allows to specify metadata keys ordering. [See example](https://github.com/IntersectMBO/cardano-node/issues/4335#issuecomment-2047984996).

Fixes https://github.com/IntersectMBO/cardano-node/issues/4335

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/TxMetadata.hs
+++ b/cardano-api/internal/Cardano/Api/TxMetadata.hs
@@ -57,6 +57,7 @@ import           Cardano.Api.SerialiseCBOR (SerialiseAsCBOR (..))
 import qualified Cardano.Ledger.Binary as CBOR
 import qualified Cardano.Ledger.Shelley.TxAuxData as Shelley
 
+import qualified Codec.CBOR.Magic as CBOR
 import           Control.Applicative (Alternative (..))
 import           Control.Monad (guard, when)
 import qualified Data.Aeson as Aeson
@@ -85,13 +86,11 @@ import qualified Data.Text.Lazy.Builder as Text.Builder
 import qualified Data.Vector as Vector
 import           Data.Word
 
-{- HLINT ignore "Use lambda-case" -}
-
 -- ----------------------------------------------------------------------------
 -- TxMetadata types
 --
 
-newtype TxMetadata = TxMetadata (Map Word64 TxMetadataValue)
+newtype TxMetadata = TxMetadata { unTxMetadata :: Map Word64 TxMetadataValue }
     deriving (Eq, Show)
 
 data TxMetadataValue = TxMetaMap    [(TxMetadataValue, TxMetadataValue)]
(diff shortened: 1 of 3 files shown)


cardano-api/internal/Cardano/Api/TxMetadata.hs
```diff
@@ -556,9 +555,11 @@ metadataValueFromJsonNoSchema = conv
       $ Vector.toList vs
 
     conv (Aeson.Object kvs) =
-        fmap TxMetaMap
+        fmap
+        ( TxMetaMap
+        . sortCanonicalForCbor
+        )
       . traverse (\(k,v) -> (,) (convKey k) <$> conv v)
-      . List.sortOn fst
```
carbolymer, line 561: The JSON map was previously sorted lexicographically by keys and then it was processed down the line as a list of pairs. So this place was deciding about the sorting order of the object fields' names.

cardano-api/test/cardano-api-test/Test/Cardano/Api/Metadata.hs
```diff
@@ -29,63 +33,151 @@ import           Test.Tasty.Hedgehog (testProperty)
...
-    case Aeson.decodeStrict' jsonStr of
-      Nothing -> Hedgehog.failure
-      Just json -> do
-        Hedgehog.annotateShow json
-        metadataFromJson TxMetadataJsonNoSchema json === Right metadata
-
+matchMetadata :: HasCallStack
+              => TxMetadataJsonSchema
+              -> ByteString -- ^ json string to test
+              -> TxMetadata -- ^ expected metadata
```
smelc, line 175:
```suggestion
              -> ByteString -- ^ Json string to test
              -> TxMetadata -- ^ Expected metadata
```

smelc (approved):
Great work :heart:

cardano-api/internal/Cardano/Api/TxMetadata.hs
Jimbo4350: So this will work but I think a more robust solution would be to do something similar to `HashableScriptData`. We should serialize the metadata to CBOR once and carry around the canonical representation.

carbolymer: I see. This would complicate things a bit because here we're going through ledger type: https://github.com/IntersectMBO/cardano-api/blob/03de67a58e044c41c62ab8683bffac52771bcd37/cardano-api/internal/Cardano/Api/TxMetadata.hs#L135 which is using ledger type (isomorphic to ours): https://github.com/IntersectMBO/cardano-ledger/blob/dbce4e4a3ad508b2bfd2df897b617c19b7f049de/eras/shelley/impl/src/Cardano/Ledger/Shelley/TxAuxData.hs#L95
So I guess I'd have to modify `SerialiseAsCBOR` to handle different metadata types 🤔

Jimbo4350 (approved):

PR #518, opened 2024-04-12
Release cardano-api-8.45.0.0

smelc:
# Changelog

```yaml
- description: |
    Release cardano-api-8.45.0.0
  type:
  - release        # related to a new release preparation
```

# Context

CHaP PR: https://github.com/IntersectMBO/cardano-haskell-packages/pull/742

### DIFF
--- a/cardano-api/CHANGELOG.md
+++ b/cardano-api/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Changelog for cardano-api
 
+## 8.45.0.0
+
+- Sort metadata keys for no-schema json for canonical CBOR
+  (breaking, bugfix)
+  [PR 517](https://github.com/IntersectMBO/cardano-api/pull/517)
+
+- Implement estimateBalancedTxBody and estimateOrCalculateBalancedTxBody
+  (feature, compatible)
+  [PR 511](https://github.com/IntersectMBO/cardano-api/pull/511)
+
+- Add support for CC keys in the SomeAddressVerificationKey type
+  (feature, breaking)
+  [PR 514](https://github.com/IntersectMBO/cardano-api/pull/514)
+
 ## 8.44.0.0
 
 - - Add `getReferenceInputsSizeForTxIds` function.
--- a/cardano-api/cardano-api.cabal
+++ b/cardano-api/cardano-api.cabal
@@ -1,7 +1,7 @@
 cabal-version: 3.4
 
 name:                   cardano-api
-version:                8.44.0.0
+version:                8.45.0.0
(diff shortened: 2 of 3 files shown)


cardano-api/CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
...
+
+- Implement estimateBalancedTxBody and estimateOrCalculateBalancedTxBody
+  (feature, compatible)
+  [PR 511](https://github.com/IntersectMBO/cardano-api/pull/511)
+
+- Add support for CC keys in the SomeAddressVerificationKey type
+  (feature, breaking)
+  [PR 514](https://github.com/IntersectMBO/cardano-api/pull/514)
+
+- - Add `getReferenceInputsSizeForTxIds` function.
```
carbolymer, line 13: This shouldn't be in changelog as it was released previously. Is it a bug in the changelog script?

carbolymer (changes requested):
Redundant entry in changelog

cardano-api/CHANGELOG.md
smelc: Good catch, this probably comes from this interleaving: 

image

carbolymer: Linearized view of GH commits is ass.
image
Yeah. Seems that the script looked at `428fa68e4` or `296a61656` but they aren't children of 8.44 tag `530d44384` so they shouldn't be taken into account.

smelc: This is caused by `git log`'s behavior:

```
       List commits that are reachable by following the parent links from the given commit(s), but exclude commits
       that are reachable from the one(s) given with a ^ in front of them.

       A special notation "<commit1>..<commit2>" can be used as a short-hand for "^<commit1> <commit2>".
```

So in this case the script executes `git log --merges --oneline "cardano-api-8.44.0.0..HEAD"` i.e.:

```
git log --merges --oneline ^cardano-api-8.44.0.0 HEAD
```

And indeed, `428fa68e4` is a parent of `HEAD` that is not a parent of `cardano-api-8.44.0.0`, so it gets listed. What we want is the list of commits between the two commits we give, not git log's behavior of using the parent link.

smelc: I believe we need to add [--ancestry-path](https://git-scm.com/docs/git-log#Documentation/git-log.txt---ancestry-pathltcommitgt) to our `git log` invokation.

smelc:
PR updating API in CLI: https://github.com/IntersectMBO/cardano-cli/pull/720

PR #555, opened 2024-06-14
evaluateTransactionExecutionUnitsShelley: return logs

smelc:
# Changelog

```yaml
- description: |
    evaluateTransactionExecutionUnitsShelley: return logs that are useful for debugging
  type:
```

# Context

This was asked by a user [here on Slack](https://input-output-rnd.slack.com/archives/G011N23CEAE/p1717012833898899) and it makes sense, since existing callers can just ignore the new logs that are being returned; taking advantage of Haskell's lazyness.

# How to trust this PR

It's returning more data

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/Fees.hs
+++ b/cardano-api/internal/Cardano/Api/Fees.hs
@@ -84,7 +84,7 @@ import qualified Ouroboros.Consensus.HardFork.History as Consensus
 import qualified PlutusLedgerApi.V1 as Plutus
 
 import           Control.Monad (forM_)
-import           Data.Bifunctor (bimap, first)
+import           Data.Bifunctor (bimap, first, second)
 import           Data.ByteString.Short (ShortByteString)
 import           Data.Foldable (toList)
 import           Data.Function ((&))
@@ -97,6 +97,7 @@ import           Data.Ratio
 import           Data.Set (Set)
 import qualified Data.Set as Set
 import qualified Data.Text as Text
+import           Data.Text (Text)
 import           Lens.Micro ((.~), (^.))
 
 {- HLINT ignore "Redundant return" -}
@@ -618,7 +619,7 @@ evaluateTransactionExecutionUnits :: forall era. ()
   -> UTxO era
   -> TxBody era
   -> Either (TransactionValidityError era)
-            (Map ScriptWitnessIndex (Either ScriptExecutionError ExecutionUnits))
+            (Map ScriptWitnessIndex (Either ScriptExecutionError ([Text], ExecutionUnits)))
 evaluateTransactionExecutionUnits era systemstart epochInfo pp utxo txbody =
     case makeSignedTransaction' era [] txbody of
       ShelleyTx sbe tx' -> evaluateTransactionExecutionUnitsShelley sbe systemstart epochInfo pp utxo tx'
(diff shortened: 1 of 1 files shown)


cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -618,7 +618,7 @@ evaluateTransactionExecutionUnits :: forall era. ()
   -> UTxO era
   -> TxBody era
   -> Either (TransactionValidityError era)
-            (Map ScriptWitnessIndex (Either ScriptExecutionError ExecutionUnits))
+            (Map ScriptWitnessIndex (Either ScriptExecutionError ([Text.Text], ExecutionUnits)))
```
palas, line 621: I would import the constructor like this: `import Data.Text (Text)` and then just write `Text` everywhere instead of `Text.Text`. It is already done like this with `Map`.

I am also aware that, before your PR, there was already a case where it was done like `Text.Text` too, so you are following the pattern either way.

palas (approved):
You're right, it looks like `evalTxExUnits` just uses `evalTxExUnitsWithLogs` anyway.
I just wrote a stylistic suggestion.

cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -631,11 +631,11 @@ evaluateTransactionExecutionUnitsShelley :: forall era. ()
   -> UTxO era
   -> L.Tx (ShelleyLedgerEra era)
   -> Either (TransactionValidityError era)
-            (Map ScriptWitnessIndex (Either ScriptExecutionError ExecutionUnits))
+            (Map ScriptWitnessIndex (Either ScriptExecutionError ([Text.Text], ExecutionUnits)))
```
Jimbo4350, line 634: A type annotation isn't strictly necessary but it wouldn't hurt here.

Jimbo4350 (changes requested):
Nice :+1: . I agree with @palas's comment

cardano-api/internal/Cardano/Api/Fees.hs
smelc: @palas> done the change :+1:

cardano-api/internal/Cardano/Api/Fees.hs
smelc: @Jimbo4350> do you mean something like:

```hs
type TransactionExecutionUnits = Map ScriptWitnessIndex (Either ScriptExecutionError ([Text], ExecutionUnits))
```

and using `TransactionExecutionUnits` in `evaluateTransactionExecutionUnits`?

Jimbo4350 (approved):

cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -618,7 +619,7 @@ evaluateTransactionExecutionUnits :: forall era. ()
   -> UTxO era
   -> TxBody era
   -> Either (TransactionValidityError era)
-            (Map ScriptWitnessIndex (Either ScriptExecutionError ExecutionUnits))
+            (Map ScriptWitnessIndex (Either ScriptExecutionError ([Text], ExecutionUnits)))
```
Jimbo4350, line 622: A type synonym for `[Text]` would aid in readability

smelc: @Jimbo4350> added type synonym `EvalTxExecutionUnitsLog`.

cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -648,12 +653,12 @@ evaluateTransactionExecutionUnitsShelley sbe systemstart epochInfo (LedgerProtoc
       :: Alonzo.AlonzoEraScript (ShelleyLedgerEra era)
       => AlonzoEraOnwards era
       -> Map (L.PlutusPurpose L.AsIx (ShelleyLedgerEra era))
-             (Either (L.TransactionScriptFailure (ShelleyLedgerEra era)) Alonzo.ExUnits)
-      -> Map ScriptWitnessIndex (Either ScriptExecutionError ExecutionUnits)
+             (Either (L.TransactionScriptFailure (ShelleyLedgerEra era)) ([Text.Text], Alonzo.ExUnits))
+      -> Map ScriptWitnessIndex (Either ScriptExecutionError ([Text.Text], ExecutionUnits))
```
Jimbo4350, line 657: Here as well.

Jimbo4350 (approved):

cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -648,12 +653,12 @@ evaluateTransactionExecutionUnitsShelley sbe systemstart epochInfo (LedgerProtoc
...
     fromLedgerScriptExUnitsMap aOnwards exmap =
       Map.fromList
         [ (toScriptIndex aOnwards rdmrptr,
-           bimap (fromAlonzoScriptExecutionError aOnwards) fromAlonzoExUnits exunitsOrFailure)
+           bimap (fromAlonzoScriptExecutionError aOnwards) (second fromAlonzoExUnits) exunitsOrFailure)
```
Jimbo4350, line 661: You can actually use `fmap` here instead of `second`.

smelc: I rather not generalize this code that is involved already. I think `second` better conveys the intent :+1:

PR #557, opened 2024-06-19
queryStateForBalancedTx: return current treasury value

smelc:
# Changelog

```yaml
- description: |
    Make the query used by the CLI's `transaction build` return the current treasury value, so that command to do treasury donation doesn't require the user to pass it. Corresponding CLI PR: https://github.com/IntersectMBO/cardano-cli/pull/778
  type:
```

# Context

Required for `transaction build` not to perform a new call to retrieve the treasury's current value in https://github.com/IntersectMBO/cardano-cli/pull/778

# How to trust this PR

It returns additional data. Existing data is unchanged.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/Convenience/Query.hs
+++ b/cardano-api/internal/Cardano/Api/Convenience/Query.hs
@@ -1,4 +1,8 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE PartialTypeSignatures #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeFamilies #-}
 
@@ -6,6 +10,7 @@
 --
 module Cardano.Api.Convenience.Query (
     QueryConvenienceError(..),
+    TxCurrentTreasuryValue(..),
     determineEra,
     -- * Simplest query related
     executeQueryCardanoMode,
@@ -17,8 +22,10 @@ module Cardano.Api.Convenience.Query (
 
 import           Cardano.Api.Address
 import           Cardano.Api.Certificate
+import           Cardano.Api.Eon.ConwayEraOnwards
 import           Cardano.Api.Eon.ShelleyBasedEra
 import           Cardano.Api.Eras
+import           Cardano.Api.Feature (Featured (..))
 import           Cardano.Api.IO
 import           Cardano.Api.IPC
 import           Cardano.Api.IPC.Monad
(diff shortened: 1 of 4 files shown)


cardano-api/internal/Cardano/Api/Convenience/Query.hs
```diff
@@ -69,6 +72,9 @@ renderQueryConvenienceError (QceUnsupportedNtcVersion (UnsupportedNtcVersionErro
   "This query requires at least " <> textShow minNtcVersion <> " but the node negotiated " <> textShow ntcVersion <> ".\n" <>
   "Later node versions support later protocol versions (but development protocol versions are not enabled in the node by default)."
 
+newtype TxCurrentTreasuryValue = TxCurrentTreasuryValue { unTxCurrentTreasuryValue :: L.Coin }
+  deriving Show
```
carbolymer, line 76: I think this could be more readable
```suggestion
  deriving newtype Show
```

cardano-api/internal/Cardano/Api/Query/Expr.hs
```diff
@@ -259,3 +261,9 @@ queryStakeVoteDelegatees :: ()
 queryStakeVoteDelegatees era stakeCredentials = do
   let sbe = conwayEraOnwardsToShelleyBasedEra era
   queryExpr $ QueryInEra $ QueryInShelleyBasedEra sbe $ QueryStakeVoteDelegatees stakeCredentials
+
+queryAccountState :: ()
+  => ShelleyBasedEra era
```
carbolymer, line 266: Are you sure about this witness? If it's in `>= v16`, shouldn't there be `ConwayEraOnwards era`?

carbolymer (approved):

cardano-api/internal/Cardano/Api/Query/Expr.hs
carbolymer: This should `>=` Conway afaik.

carbolymer (changes requested):
https://github.com/IntersectMBO/cardano-api/pull/557/files#r1654608157

Jimbo4350 (approved):
LGTM but you need to make the change suggested by @carbolymer

cardano-api/internal/Cardano/Api/Query/Expr.hs
smelc: @carbolymer, @Jimbo4350> So then I should return a `Maybe TxTreasuryValue` in `queryStateForBalancedTx`?

And not include it if the era is before Conway in `transaction build` in CLI?

carbolymer: That would make most sense. Or use `Maybe (Featured ConwayEraOnwards era TxTreasuryValue)` to signal when treasury is present.

smelc: @carbolymer> did the change in the third commit `queryStateForBalancedTx: return Maybe (Featured ConwayEraOnwards era TxTreasuryValue` :+1:

carbolymer (approved):

PR #559, opened 2024-06-19
Small improvements to queries

smelc:
# Changelog

```yaml
- description: |
    Small improvements to queries
  type:
```

# Context

* Spin-off of https://github.com/IntersectMBO/cardano-api/pull/557
* Goal is to make `fromConsensusQueryResultShelleyBased` more maintainable.

# How to trust this PR

* The first commit doesn't change the runtime
* The second commit was carefully done mechanically, but please reread it carefully
* It's more deletions than additions :medal_military: 

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/Query.hs
+++ b/cardano-api/internal/Cardano/Api/Query.hs
@@ -831,6 +831,9 @@ fromConsensusQueryResult (QueryInEra (QueryInShelleyBasedEra ShelleyBasedEraConw
                  r'
       _ -> fromConsensusQueryResultMismatch
 
+-- This function is written like this so that we have exhaustive pattern checking
+-- on the @QueryInShelleyBasedEra era result@ value. Don't change the top-level
+-- @case sbeQuery of ...@!
 fromConsensusQueryResultShelleyBased
   :: forall era ledgerera protocol result result'.
      HasCallStack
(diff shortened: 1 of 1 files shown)


cardano-api/internal/Cardano/Api/Convenience/Query.hs
```diff
@@ -69,22 +71,26 @@ renderQueryConvenienceError (QceUnsupportedNtcVersion (UnsupportedNtcVersionErro
   "This query requires at least " <> textShow minNtcVersion <> " but the node negotiated " <> textShow ntcVersion <> ".\n" <>
   "Later node versions support later protocol versions (but development protocol versions are not enabled in the node by default)."
 
+-- | Data returned by 'queryStateForBalancedTx'. Parameterized on the type of 'eraHistory',
+-- because sometimes we don't need it.
+data StateForBalancedTx history era = StateForBalancedTx
```
Jimbo4350, line 76: I don't think this refactor is necessary. It's better to explicitly have the parameters needed rather than introduce a polymorphic variable is sometimes `()` or `EraHistory`. This is also introducing more churn in the cli.

cardano-api/internal/Cardano/Api/Query.hs
```diff
@@ -842,133 +845,123 @@ fromConsensusQueryResultShelleyBased
   -> Consensus.BlockQuery (Consensus.ShelleyBlock protocol ledgerera) result'
   -> result'
   -> result
```
Jimbo4350, line 847: :+1:

Jimbo4350 (commented):
LGTM but I disagree with `StateForBalancedTx`. It's not necessary and IMO it takes away from code readability.

cardano-api/internal/Cardano/Api/Convenience/Query.hs
smelc: The thing is, I'm adding another member to this record in this follow-up PR: https://github.com/IntersectMBO/cardano-api/pull/557

If we don't add this record, this follow-up PR will make `queryStateForBalancedTx` return yet another value, which I think is not really nice.

carbolymer: I'm also unsure about this new datatype. But I think my gripe is parameterisation by `history`. I think having a single type parameter and returning a product `(EraHistory, StateForBalancedTx era)` where needed would be cleaner.

smelc: Alright, let me scratch the new datatype then :+1:

carbolymer (approved):

PR #564, opened 2024-06-24
Fix reading of Plutus V2 cost models with different lengths in AlonzoGenesis in different eras

carbolymer:
# Changelog

```yaml
- description: |
    Fix reading of Plutus V2 cost models with different lengths in AlonzoGenesis in different eras
  type:
```

# Context

Part of the fix for:
- https://github.com/IntersectMBO/cardano-cli/issues/801

Plutus V2 cost model for Babbage and for Conway have a different number of parameters. Ledger's JSON decoder for a map format of the cost model is unaware about this fact and just tries to validate for all parameters and load all 185 parameters no matter the era. This in order makes a problem for CBOR decoder, which is era-aware and will fail for a cost model with 185 parameters in Babbage.

This PR adds a function for how we decoding the cost model from JSON, which makes an adjustment in loaded parameters count depending on an era.


cardano-api/test/cardano-api-test/Test/Cardano/Api/Genesis.hs
```diff
@@ -0,0 +1,65 @@
...
+  costModelValues <-
+    fmap L.getCostModelParams
+    . H.nothingFail
+    . M.lookup L.PlutusV2
+    . L.costModelsValid
+    $ L.agCostModels genesis
+  let last10CostModelValues = reverse . take 10 $ reverse costModelValues
+
+  H.noteShow_ costModelValues
+  last10CostModelValues === replicate 10 maxBound
```
carbolymer, line 36: - [ ] add test for length

cardano-api/internal/Cardano/Api/Genesis.hs
```diff
@@ -195,8 +219,75 @@ conwayGenesisDefaults = ConwayGenesis { cgUpgradePParams = defaultUpgradeConwayP
...
+
+      obj@(A.Object _) -> do
+        -- decode cost model into a map first
+        costModel :: Map V2.ParamName Int64 <- modifyError ("Decoding cost model object: " <> ) $ fromJsonE obj
+        setCostModelDefaultValues
+          . A.toJSON -- convert to an array representation of Int64 values
+          . fmap snd
+          . sortOn fst -- ensure proper order of params in the list
+          . toList
+          . (`M.union` costModelDefaultValues) -- add default values of missing params
```
carbolymer, line 244: * [ ] use optionalCostModelDefaultValues instead here
* [ ] do the extra validation for missing fields before converting to an array

palas:
FYI: Because we now enforce the use of `cabal-gild` to format cabal files and, in order to reduce the amount of hassle, I have cherry-picked the appropriate commit into your PR. This will make the `cabal-gild` CI check pass for the PR. The commit should be discarded automatically as soon as you either rebase or merge the PR (since it is already in the `main` branch). Feel free to discard it if you want, but that will make the new required action to fail until you rebase.

palas:
FYI: I have rebased your branch because we have done changes to the formatting. I have made a copy of the unrebased branch here: [backup/mgalazyn/fix/allow-reading-plutus-v2-175-param-costmodel](https://github.com/IntersectMBO/cardano-api/tree/backup/mgalazyn/fix/allow-reading-plutus-v2-175-param-costmodel)

Jimbo4350 (changes requested):
~~I think there is a significantly simpler way to approach this.~~

~~Parameterize `cardano-api`'s  `alonzoGenesisDefaults` on the era. If conway era is specified, add in the additional cost model parameters. If not don't add them. This should result in minimal changes. This will generate the appropriate cost model when using `create-testnet-data` and we should't have any issues.~~

Jimbo4350 (approved):
LGTM however this code will disappear when we hardfork to Conway. In future we should avoid this situation completely by using an array in the genesis file and having a mechanism to pretty print the cost models (makes debugging easier). 

When supplying an array, ledger calls `mkCostModel` which always retains the original values supplied. Plutus's `mkEvaluationContext` will then automatically adjust the cost model (either filling in missing values or truncating them).

Our pretty printing function can potentially do some validation and suggest to the user if they are missing parameters from their particular cost model. However I want to avoid us manually doing the validation as much as possible, to the point of requesting ledger/plutus to expose additional functions if needed.

Can you confirm with @mgmeier  and @mkoura that this fixes their issue?

carbolymer:
>Parameterize cardano-api's alonzoGenesisDefaults on the era. If conway era is specified, add in the additional cost model parameters. If not don't add them. This should result in minimal changes.

The problem occurs when we provide alonzo genesis specification from a file. We're not using `alonzoGenesisDefaults` there, so it won't solve this.

>When supplying an array, ledger calls mkCostModel which always retains the original values supplied. Plutus's mkEvaluationContext will then automatically adjust the cost model (either filling in missing values or truncating them).

That's correct. The whole purpose of this fix was to support flexible map format.

>Our pretty printing function can potentially do some validation and suggest to the user if they are missing parameters from their particular cost model. 

I don't understand which pretty printing function are you referring to.

Jimbo4350:
> The problem occurs when we provide alonzo genesis specification from a file. We're not using alonzoGenesisDefaults there, so it won't solve this.

Benchmarking is using `create-testnet-data` which will use `alonzoGenesisDefaults` if no Alonzo genesis file is created. This would fix their problem. @mkoura uses an array so they are excluded. I don't think anyone else is affected because they're not launching Babbage testnets with V2 cost models in the Alonzo genesis file so this is really for the benchmarking team.

> I don't understand which pretty printing function are you referring to.

We would have to implement it.

> The whole purpose of this fix was to support flexible map format.

Yes but this requires us to maintain code that we should not be maintaining. Benchmarking wants to be able to view the cost model as a json map. We can give them that whilst avoiding the additional gymnastics around decoding the cost model from a json map. In the future its possible for this situation to arise again and we can (and should) avoid it.

carbolymer:
>Yes but this requires us to maintain code that we should not be maintaining

I completely agree. I'll gladly remove this code when we won't need to support this edge case anymore.

>Benchmarking wants to be able to view the cost model as a json map. 

I understood that the use case is to be able to define and inspect cost model in JSON and not just only view it. I guess we could provide separate translation layer for them doing pretty much what's done in this decoding function. It still means that we have to fix ledger bugs in `cardano-api` either way.

palas:
FYI: I have rebased your branch because we have done changes to the formatting. I have made a copy of the unrebased branch here: https://github.com/IntersectMBO/cardano-api/tree/backup2/mgalazyn/fix/allow-reading-plutus-v2-175-param-costmodel

PR #578, opened 2024-07-03
Add cabal-gild and update hook

palas:
# Changelog

```yaml
- description: |
    Added cabal-gild to dev shell, to haskell-style-lint hook script, and created CI action to check it.
  type:
    - maintenance
```

# Context

We don't want to have to worry about the formatting of `.cabal` files, and we want it to be consistent. This PR tries to address this.

# How to trust this PR

Probably check that the hook works for you. Also that you are happy with the changes and configuration, and that the CI passes.

I've also checked that it doesn't pass when the cabal file is modified and unformatted. See:
 - Discarded commit: https://github.com/IntersectMBO/cardano-api/pull/578/commits/8cbc6d764acdee10df02ff28f632e0b2a7bce017
 - Failing action corresponding to the commit: [Job 27020041405](https://github.com/IntersectMBO/cardano-api/actions/runs/9785988330/job/27020041405?pr=578)

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/.github/workflows/check-cabal-gild.yml
+++ b/.github/workflows/check-cabal-gild.yml
@@ -1,0 +1,43 @@
+name: Check cabal-gild format
+
+on:
+  merge_group:
+  pull_request:
+
+jobs:
+  check-cabal-gild:
+    runs-on: ubuntu-latest
+
+    env:
+      CARDANO_GUILD_VERSION: "1.3.1.2"
+
+    steps:
+    - name: Download cardano-gild
+      run: |
+        cardano_gild_path="$(mktemp -d)"
+        version="${{env.CARDANO_GUILD_VERSION}}"
+        curl -sL \
+          "https://github.com/tfausak/cabal-gild/releases/download/$version/cabal-gild-$version-linux-x64.tar.gz" \
+          | tar -C "$cardano_gild_path" -xz
+
+        echo "PATH=$cardano_gild_path:$PATH" >> "$GITHUB_ENV"
+
+    - uses: actions/checkout@v4
+
+    - name: Run cardano-gild over all modified files
+      run: |
+        rc="0"
+        
+        for file in $(git ls-files "*.cabal")
+        do
+          echo "cabal-gild --mode=check --input=$file"
+          if ! cabal-gild --mode=check --input="$file"
+          then
+            echo "💣 $file is badly formatted. Fix it with:"
+            echo "cabal-gild --mode=format --io=$file"
+            rc="1"
+          fi
+        done
+        
+        exit $rc
+
--- a/cardano-api/cardano-api.cabal
+++ b/cardano-api/cardano-api.cabal
@@ -1,409 +1,411 @@
 cabal-version: 3.4
+name: cardano-api
+version: 9.0.0.0
+synopsis: The cardano API
+description: The cardano API.
+category:
+  Cardano,
+  API,
 
-name:                   cardano-api
-version:                9.0.0.0
-synopsis:               The cardano API
-description:            The cardano API.
-category:               Cardano,
-                        API,
-copyright:              2020-2023 Input Output Global Inc (IOG).
-author:                 IOHK
-maintainer:             operations@iohk.io
-license:                Apache-2.0
-license-files:          LICENSE
-                        NOTICE
-build-type:             Simple
-extra-doc-files:        README.md, CHANGELOG.md
+copyright: 2020-2023 Input Output Global Inc (IOG).
+author: IOHK
+maintainer: operations@iohk.io
+license: Apache-2.0
+license-files:
+  LICENSE
+  NOTICE
+
+build-type: Simple
+extra-doc-files:
+  CHANGELOG.md
+  README.md
 
 common project-config
-  default-language:     Haskell2010
-  default-extensions:   OverloadedStrings
-  build-depends:        base >= 4.14 && < 4.20
-
-  ghc-options:          -Wall
-                        -Wcompat
-                        -Wincomplete-record-updates
-                        -Wincomplete-uni-patterns
-                        -Wno-unticked-promoted-constructors
-                        -Wpartial-fields
-                        -Wredundant-constraints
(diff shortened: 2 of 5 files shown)


.github/workflows/check-cabal-gild.yml
```diff
@@ -0,0 +1,53 @@
...
+
+    - name: Run cardano-gild over all modified files
+      run: |
+        git add .
+        git stash
+        git fetch origin ${{ github.base_ref }} --unshallow
+        for x in $(git diff --name-only --diff-filter=ACMR origin/${{ github.base_ref }}..HEAD "*.cabal" | tr '\n' ' '); do
+          cabal-gild -i "$x" -o "$x"
+        done
+        git --no-pager diff --exit-code
```
smelc, line 52:
```suggestion
        rc="0"
        
        for file in $(git ls-files "*.cabal")
        do
          echo "cabal-gild --mode=check --input=$file"
          if ! cabal-gild --mode=check --input="$file"
          then
            echo "💣 $file is badly formatted. Fix it with:"
            echo "cabal-gild --mode=format --io=$file"
            rc="1"
          fi
        done
        
        exit $rc
```

Rationale:

1. Less git magic: either to understand, and so to maintain
2. Provides the command to fix the problem, instead of showing the diff; so easier to put into action

.github/workflows/check-cabal-gild.yml
```diff
@@ -0,0 +1,53 @@
...
+      fail-fast: false
+
+    env:
+      CARDANO_GUILD_VERSION: "1.3.1.2"
+
+    steps:
+    - name: Download cardano-gild
+      run: |
+        cardano_gild_path="$(mktemp -d)"
+        version="${{env.CARDANO_GUILD_VERSION}}"
```
carbolymer, line 35: If we have cabal gild in nix shell, I think it would be better to use the one from nix. We wouldn't have to keep the two versions (gha and nix shell) in sync.

cardano-api/cardano-api.cabal
```diff
@@ -1,409 +1,417 @@
...
 
 common maybe-Win32
   if os(windows)
-    build-depends:      Win32
+    build-depends: Win32
 
 common text
-  if impl(ghc == 8.10.7) && os(darwin) && arch(aarch64)
-    build-depends: text >= 1.2.5.0
+  if impl(ghc == 8.10.7)&& os(darwin)&& arch(aarch64)
```
carbolymer, line 49: huh this one's weird with `&&` shifted to the left. But I guess we can live with that.

cardano-api/cardano-api.cabal
```diff
@@ -1,409 +1,417 @@
...
-                      , transformers
-                      , transformers-except ^>= 0.1.3
-                      , typed-protocols ^>= 0.1.1
-                      , unordered-containers >= 0.2.11
-                      , vector
-                      , yaml
+  exposed-modules:
+    -- TODO: Eliminate in the future when
+    -- we create wrapper types for the ledger types
+    -- in this module
```
carbolymer, line 64: Oh no this is wrong. This comment was over the orphans module.
1. This comment is outdated - should be removed.
2. How does cabal-gild handle such comments? Was this a fluke, or those comments will always be moved to the top?
3. What about comments in the same line after the module - what happens with them? e.g.
  ```
  Cardano.Api.Address -- something something
  ```

cardano-api/cardano-api.cabal
```diff
@@ -1,409 +1,417 @@
...
-                      , cryptonite
-                      , deepseq
-                      , memory
-                      , nothunks
-                      , ouroboros-network-protocols
-                      , typed-protocols ^>= 0.1.1
+  exposed-modules:
+    -- TODO: Eliminate Cardano.Api.Ledger when
+    -- cardano-api only depends on modules
+    -- exposed by cardano-api-ledger
```
carbolymer, line 239: Again, comment in the wrong place. I'm not sure if this one is up to date. I think you should consult with @Jimbo4350 this one.

carbolymer (changes requested):
Blocking on misaligned comments in the cabal file.

.github/workflows/check-cabal-gild.yml
```diff
@@ -0,0 +1,53 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.type }}-${{ startsWith(github.ref, 'refs/heads/gh-readonly-queue/') && github.ref || github.event.pull_request.number || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  check-cabal-gild:
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
```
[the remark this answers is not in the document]
palas: No, I just copied it from the stylish-haskell workflow. I will delete it then

.github/workflows/check-cabal-gild.yml
palas: Nice. Yes, it is two files so, no need to optimize.

cardano-api/cardano-api.cabal
palas: You are right. I will try that. And I will check how `cabal-fmt` does it. Otherwise we can qualify the comment, even it is not ideal

palas: Nice, the inline comments worked!

cardano-api/cardano-api.cabal
palas: `cabal-fmt` looks nicer:
```
if ((impl(ghc ==8.10.7) && os(osx)) && arch(aarch64))
```

cardano-api/cardano-api.cabal
palas: `cabal-fmt` puts them in the end instead, same issue

cardano-api/cardano-api.cabal
Jimbo4350: We can remove the comment.

Jimbo4350 (approved):

PR #582, opened 2024-07-05
Set Fourmolu as formatter

palas:
# Changelog

```yaml
- description: |
    Added Fourmolu to dev shell, to haskell-style-lint hook script, and modified stylish-haskell action to use Fourmolu instead
  type:
```

# Context

We don't want to have to worry about formatting `.hs` files, and we want to be consistent. This PR tries to address this by establishing and enforcing a strong formatter (Formoulu).

# How to trust this PR

Probably check that the hook works for you. Also that you are happy with the changes and configuration, and that the CI passes.

# Checklist

- [x] Self-reviewed the diff


Jimbo4350 (approved):

fourmolu.yaml
```diff
@@ -0,0 +1,17 @@
+indentation: 2
+column-limit: 100
```
carbolymer, line 2: In 2024 I think we could go with a little bit higher number like 120.

fourmolu.yaml
```diff
@@ -0,0 +1,17 @@
...
+function-arrows: leading
+comma-style: leading
+import-export-style: leading
+indent-wheres: true
```
carbolymer, line 6: I think I'd prefer this to be false. We don't compromise readability and have more horizontal space for functions' bodies in where

fourmolu.yaml
```diff
@@ -0,0 +1,17 @@
...
+record-brace-space: true
```
carbolymer, line 7: I think `false` would be better here. Consider this snippet now:
```haskell
  (==)
    (AddressInEra ByronAddressInAnyEra addr1)
    (AddressInEra ByronAddressInAnyEra addr2) = addr1 == addr2
  (==)
    (AddressInEra ShelleyAddressInEra {} addr1)
    (AddressInEra ShelleyAddressInEra {} addr2) = addr1 == addr2
```
braces look a bit confusing here - they're there because we don't want to match on `ShelleyAddressInEra` fields, but the spacing suggests something else.

fourmolu.yaml
```diff
@@ -0,0 +1,17 @@
...
+newlines-between-decls: 1
+haddock-style: single-line
+haddock-style-module:
+let-style: auto
+indent-wheres: false
```
carbolymer, line 12: duplicated entry

fourmolu.yaml
```diff
@@ -0,0 +1,17 @@
...
+in-style: right-align
+unicode: never
+respectful: false
+fixities: []
+single-constraint-parens: never
```
carbolymer, line 17: I think we could add 
```yaml
single-deriving-parens: false
```

carbolymer, line 17: 1. Would it be possible to port import grouping settings from stylish we had before? I found them useful to have a separate import groups for cardano universe, hackage packages and local modules.
2. What do you think about enabling https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/import_qualified_post.html globally? I find it a bit annoying to jump left and right with my eyes while scanning imports. Fourmolu supports this extension OOTB.

carbolymer (commented):
Nice work Pablo! I have a few suggestions for [fourmolu settings](https://fourmolu.github.io/config/) we could use.

fourmolu.yaml
palas: Oh, I hadn't look at the extensions. Definitely, the lack of space for the qualified bit was bothering me. I think there is also a way of grouping local vs external, the advantage of that is that we wouldn't have to update the rules, but I guess that wouldn't be able to tell what are Cardano universe ones

palas: Sounds good to me

palas: Ah, wait, that is a GHC extension, not a fourmolu extension. I think it wouldn't be difficult to make a small perl script that would fix that for us and we can run it after fourmolu. Let me give that a try

palas: Or we can just run stylish-haskell after fourmolu 🤔

palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/583

palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/583

fourmolu.yaml
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/583

fourmolu.yaml
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/583

fourmolu.yaml
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/583

fourmolu.yaml
carbolymer: That's fine - I'm not insisting on that. We can go with 100. 👌🏻

PR #603, opened 2024-07-30
Remove CPP extension from `Cardano.Api.Fees`.  Use `IsList(toList,fromList)` instead of specialised functions.

carbolymer:
# Changelog

```yaml
- description: |
    Remove CPP extension from Cardano.Api.Fees and reformat
    Use `IsList(toList,fromList)` instead of specialised functions
  type:
```

# Context
This PR can be reviewed on commit-by-commit basis.

Removed `CPP` extension from `Cardano.Api.Fees` on redundant pattern matching.

Replaced specialized `toList` and `fromList` with `IsList(toList, fromList)` functions. This reduces mental overhead when writing code - you don't have to wonder which collection are you converting to/from anymore.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/Fees.hs
+++ b/cardano-api/internal/Cardano/Api/Fees.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE CPP #-}
 {-# LANGUAGE ConstraintKinds #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
@@ -9,7 +8,6 @@
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
-{-# OPTIONS_GHC -Wno-unticked-promoted-constructors #-}
 
 -- | Fee calculation
 module Cardano.Api.Fees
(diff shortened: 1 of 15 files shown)


cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -1296,12 +1293,9 @@ calcReturnAndTotalCollateral retColSup fee pp' TxInsCollateral{} txReturnCollate
         L.Coin amt = totalCollateralLovelace * 100 - requiredCollateral
         returnCollateral = L.rationalToCoinViaFloor $ amt % 100
     case (txReturnCollateral, txTotalCollateral) of
-#if MIN_VERSION_base(4,16,0)
-#else
       -- For ghc-9.2, this pattern match is redundant, but ghc-8.10 will complain if its missing.
       (rc@TxReturnCollateral{}, tc@TxTotalCollateral{}) ->
```
Jimbo4350, line 1300: This is not a redundant pattern match. If the user specifies return and total collateral we don't do any calculations and return their values.

Jimbo4350 (changes requested):
I think removing that case will break things in the cli.

cardano-api/internal/Cardano/Api/Fees.hs
carbolymer: That's what's happening here. We're not doing any calculations and just returning the values, since everything in `let` block is done lazily.

Jimbo4350: Apologies, misread the diff!

Jimbo4350 (approved):

Issue #607, opened 2024-08-06
[FEATURE] - Challenging Breaking change `Use the ledger's Coin instead of Lovelace`in `8.40.0` 


nhenin:
**Internal/External** : *Internal*   
**Context** : I am challenging this commit : https://github.com/IntersectMBO/cardano-api/commit/46c1e605244fb3d01deda3451a0b56763c7c9370 causing a breaking change into `8.40.0` 
**Where**
```haskell
newtype Lovelace = Lovelace Integer
  deriving stock (Eq, Ord, Show)
  deriving newtype (Enum, Real, Integral, Num, ToJSON, FromJSON, ToCBOR, FromCBOR)
```
**has been replaced by**
```haskell
-- | The amount of value held by a transaction output.
newtype Coin = Coin {unCoin :: Integer}
  deriving
    ( Eq
    , Ord
    , Enum
    , NoThunks
    , Generic
    , ToJSON
    , FromJSON
    , NFData
    )
  deriving (Show) via Quiet Coin
  deriving (Semigroup, Monoid, Group, Abelian) via Sum Integer
  deriving newtype (PartialOrd, ToCBOR, EncCBOR, HeapWords)
````
imo, It is misleading for following reasons :
1. `New Invariant` : It is a Positive Quantity as it is `-- | The amount of value held by a transaction output`. Before you could think it was only Lovelace Quantities.
2. `Naming` : `Coin` is ~ to `Token` so you would expect `CurrencySymbol + Name + Quantity`, here is a definition of Token vs Coin 
```
Coins vs Tokens: Key Differences
The primary difference between coins and tokens lies in their structure and purpose. Coins operate on their own
 blockchain and primarily function as a medium of exchange. They are akin to digital forms of value exchange and are 
often used to pay for transactions within their respective blockchain network.

Tokens, however, operate on existing blockchain networks and aim to offer a wider range of functionalities. They are 
often associated with a specific project or protocol within the blockchain ecosystem and are used to access certain 
features of that project.

In terms of creation, tokens are easier to create than coins. Creating a coin requires building a new blockchain, which 
requires time and expertise. Tokens, however, can be created on an existing blockchain, making the process simpler 
and more accessible.
```
If we use the term `Coin` it needs more explanation. I didn't find any mention of this breaking change in the change log.
[...]

PR #611, opened 2024-08-08
Adapt user-facing functions' names to Lovelace->Coin renaming

smelc:
# Changelog

```yaml
- description: |
    Deprecated some functions named `*Lovelace*`. Use the same ones with the `Coin` name instead (as advised in deprecation stanza).
  type:
```

# Context

As noticed by https://github.com/IntersectMBO/cardano-api/issues/607, when we removed the `Lovelace` type (and replaced it with the ledger's `Coin`), we did not adapt function names accordingly. This PR hence deprecates the `*Lovelace*` functions and advice to use the same ones with the `Coin` name instead.

Fixes https://github.com/IntersectMBO/cardano-api/issues/607

# How to trust this PR

* It's only renamings
* There are no `*Lovelace*` user-facing names anymore that are not deprecated
* The new `*Coin*` functions are correctly exported

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/Value.hs
+++ b/cardano-api/internal/Cardano/Api/Value.hs
@@ -27,10 +27,15 @@ module Cardano.Api.Value
   , calcMinimumDeposit
 
     -- ** Ada \/ L.Coin specifically
+  , quantityToCoin
   , quantityToLovelace
+  , coinToQuantity
   , lovelaceToQuantity
+  , selectCoin
   , selectLovelace
+  , coinToValue
   , lovelaceToValue
+  , valueToCoin
   , valueToLovelace
 
     -- ** Alternative nested representation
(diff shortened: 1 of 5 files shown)


carbolymer (approved):
Good idea!

cardano-api/internal/Cardano/Api/Value.hs
```diff
@@ -243,22 +256,31 @@ negateLedgerValue sbe v =
 filterValue :: (AssetId -> Bool) -> Value -> Value
 filterValue p (Value m) = Value (Map.filterWithKey (\k _v -> p k) m)
 
+{-# DEPRECATED selectLovelace "Use selectCoin instead." #-}
 selectLovelace :: Value -> L.Coin
-selectLovelace = quantityToLovelace . flip selectAsset AdaAssetId
+selectLovelace = selectCoin
```
Jimbo4350, line 261: Why? `Coin` is lovelace which is the common vernacular!

Jimbo4350 (commented):
I think this PR needs to be reverted. Get in touch with @CarlosLopezDeLara because people don't refer to lovelace as coin. The type is `Coin` but it represents lovelace!

nhenin:
Thx @smelc :-)

PR #629, opened 2024-09-05
Enable multiple cores in haskell GHA

carbolymer:
# Changelog

```yaml
- description: |
    Enable multiple cores in haskell GHA
  type:
```

# Context

There's a slight improvement in build times. See the screenshot below: left is before, right is after the change.

screenshot-20250107_100550

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -120,7 +120,7 @@ jobs:
     # this should be a no-op, but if the cache key was not found we need to build stuff so we can
     # cache it for the next step.
     - name: Install dependencies
-      run: cabal build all --enable-tests --only-dependencies -j --ghc-option=-j4
+      run: cabal build all --enable-tests --only-dependencies
 
     # Always store the cabal cache.
     - name: Cache Cabal store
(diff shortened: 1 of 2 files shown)


.github/workflows/haskell.yml
```diff
@@ -134,14 +134,14 @@ jobs:
 
     # Now we build.
     - name: Build all
-      run: cabal build all --enable-tests
+      run: cabal build all --enable-tests -j --ghc-option=-j4
```
smelc, line 137:
```suggestion
      run: cabal build all --enable-tests -j --ghc-option=-j4 # Using 4 cores, because that's the number of processors of GitHub runners: https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories
```

.github/workflows/haskell.yml
```diff
@@ -134,14 +134,14 @@ jobs:
...
 
     - name: Run tests
       env:
         TMPDIR: ${{ runner.temp }}
         TMP: ${{ runner.temp }}
         KEEP_WORKSPACE: 1
-      run: cabal test all --enable-tests --test-show-details=direct -j1
+      run: cabal test all --enable-tests --test-show-details=direct -j
```
smelc, line 144: With `-j` alone, `cabal` will find out the number of cores to use on its own?

smelc (approved):

.github/workflows/haskell.yml
carbolymer: Yes
```
 -j[NUM], --jobs[=NUM]          Run NUM jobs simultaneously (or '$ncpus' if no
                                NUM is given).
```

.github/workflows/haskell.yml
carbolymer: I think this -j4 is redundant.

PR #634, opened 2024-09-11
Add new types to CDDL list and test forward compatibility of `deserialiseTxLedgerCddl`

palas:
# Changelog

```yaml
- description: |
    Added new types to CDDL and added test for forward compatibility of `deserialiseTxLedgerCddl`
  type:
  - test           # fixes/modifies tests
```

# Context

There was an issue with transactions created using the non-deprecated function `serialiseToTextEnvelope` as alternative to `serialiseTxLedgerCddl`. Because the implementation in `cardano-cli` still doesn't accept it.

This PR adds a test ensuring the `deserialiseTxLedgerCddl` function is compatible with `serialiseToTextEnvelope`.

It also adds the new types to the list of `cddlTypeToEra` conversion function, this is necessary for the corresponding fix in `cardano-cli`: https://github.com/IntersectMBO/cardano-cli/pull/892

# How to trust this PR

The change is small. Probably just make sure you agree with the direction of the change.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-api/cardano-api.cabal
+++ b/cardano-api/cardano-api.cabal
@@ -338,6 +338,7 @@ test-suite cardano-api-test
     tasty,
     tasty-hedgehog,
     tasty-quickcheck,
+    text,
     time,
 
   other-modules:
--- a/cardano-api/internal/Cardano/Api/SerialiseLedgerCddl.hs
+++ b/cardano-api/internal/Cardano/Api/SerialiseLedgerCddl.hs
@@ -12,6 +12,7 @@
 module Cardano.Api.SerialiseLedgerCddl
   ( TextEnvelopeCddlError (..)
   , FromSomeTypeCDDL (..)
+  , cddlTypeToEra
 
     -- * Reading one of several transaction or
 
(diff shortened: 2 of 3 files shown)


Jimbo4350 (approved):
LGTM! We need to remove all the type strings that are not defined in a `HasTextEnvelope` instance.

cardano-api/test/cardano-api-test/Test/Cardano/Api/Typed/CBOR.hs
```diff
@@ -32,6 +33,18 @@ import           Test.Tasty.Hedgehog (testProperty)
 -- TODO: Need to add PaymentExtendedKey roundtrip tests however
 -- we can't derive an Eq instance for Crypto.HD.XPrv
 
+prop_forward_compatibility_txbody_CBOR :: Property
```
carbolymer, line 36: This test does the same as the one above, however `prop_roundtrip_txbody_CBOR` uses a deprecated function. I guess it would be nice to add a comment with that information.

@Jimbo4350 when should we remove already deprecated API functions.

carbolymer (approved):

cardano-api/test/cardano-api-test/Test/Cardano/Api/Typed/CBOR.hs
palas: Added comment here: https://github.com/IntersectMBO/cardano-api/pull/634/commits/b33f0afdef355c9ce20f8491ef2885f76726c893

Issue #600, opened 2024-07-29
[FR] - Capability to add additional data in a body content


mmontin:
**Internal/External**
*External*  

**Area**
*Plutus*

**Describe the feature you'd like**

At the moment, it does not seem possible to pass additional data in the form of a correspondance between a datum hash and a datum in a transaction body, in addition to the one collected from the script witnesses in inputs, and the outputs of the body. 

However, when referencing a utxo which happens to have a hashed datum, this is needed.

Indeed, in function `mkShelleyTransactionBody`  we can see that the `HashableScriptData` are derived from the body itself, from the two locations mentioned above, as follows:

```
    scriptdata :: [HashableScriptData]
    scriptdata =
      [d | TxOut _ _ (TxOutDatumInTx _ d) _ <- txOuts]
        ++ [ d
           | ( _
              , AnyScriptWitness
                  ( PlutusScriptWitness
                      _
                      _
                      _
                      (ScriptDatumForTxIn (Just d))
                      _
                      _
                    )
              ) <-
              witnesses
           ]
```

Then, the verification hash of the body is computed from these data using `convPParamsToScriptIntegrityHash`, and is wrapped as well in the Shelley transaction body. As a consequence, there seem to be no builtin way to have additional data in the resulting `Shelley` transaction body.

_My feature request is to have such a way of including additional data in the transaction body._

This could take several forms, such as:
* an additional field in the `TxBodyContent` structure
* an additional parameter in the `mkShelleyTransactionBody` function

**Describe alternatives you've considered**

For now, I have used the following work-around:
* generating the transaction body using `mkShelleyTransactionBody`
* deconstructing the generated transaction body
* manually adding the necessary data in the respective field of the transaction
* manually re-generating the script integrity hash using the new data map with `convPParamsToScriptIntegrityHash`
* wrapping back those changed elements into a new Shelley transaction body.

This does allow the ledger to generate the proper script context but it does not feel like the proper way to achieve this result.

smelc:
cc @Jimbo4350 @carbolymer

mmontin:
Following a discussion I had on Discord, it was confirmed that this feature is not currently present. I will work on a contribution to integrate it.

mmontin:
Just realized that the issue is most likely the same when consuming inputs from pkh with a hashed datum. All indicate that there would currently be no way to provide the associated datum in the body.

smelc:
Closed by https://github.com/IntersectMBO/cardano-api/pull/640

PR #637, opened 2024-09-13
Extra `HashableScriptData` in `TxBody` for PK inputs and reference inputs with hashed datums.

mmontin:
This PR aims to solve issue #600 by allowing users to provide instances of `HashableScriptData` for hashed datums located:
* in reference inputs
* in regular inputs owned by private keys

For now, these data can only be provided:
* in outputs
* in regular inputs owned by scripts

This contribution intends to be as little invasive as possible, and to respect the current way of handling the creation of the data map provided in the script context. The map is currently built by browsing through the relevant parts of the body content and retrieving the relevant data there. Thus, the idea here is not to create a new field in the body content, but instead to allow user to add additional `HashableScriptData` at two addition locations in the body content.

### DIFF
--- a/cardano-api/internal/Cardano/Api/Script.hs
+++ b/cardano-api/internal/Cardano/Api/Script.hs
@@ -54,11 +54,13 @@ module Cardano.Api.Script
   , WitCtxStake
   , WitCtx (..)
   , ScriptWitness (..)
+  , PrivateKeyWitness (..)
   , Witness (..)
   , KeyWitnessInCtx (..)
   , ScriptWitnessInCtx (..)
   , IsScriptWitnessInCtx (..)
   , ScriptDatum (..)
+  , KeyDatum (..)
   , ScriptRedeemer
   , scriptWitnessScript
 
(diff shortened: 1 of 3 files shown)


cardano-api/internal/Cardano/Api/Script.hs
```diff
@@ -722,6 +724,15 @@ data ScriptWitness witctx era where
 
 deriving instance Show (ScriptWitness witctx era)
 
+data PrivateKeyWitness witctx era where
+  PrivateKeyWitness
+    :: KeyDatum witctx
+    -> PrivateKeyWitness witctx era
```
smelc, line 730: Can you make this not a GADT? I know there are lots of them floating around but we are actually moving away from using GADTs when not needed.

cardano-api/internal/Cardano/Api/Script.hs
```diff
@@ -773,6 +784,14 @@ deriving instance Eq (ScriptDatum witctx)
 
 deriving instance Show (ScriptDatum witctx)
 
+data KeyDatum witctx where
+  KeyDatumForTxIn :: Maybe HashableScriptData -> KeyDatum WitCtxTxIn
+  NoKeyDatumForStake :: KeyDatum WitCtxStake
```
smelc, line 789: Can you make this not a GADT? I know there are lots of them floating around but we are actually moving away from using GADTs when not needed.

Jimbo4350 (changes requested):
Sorry this is not the cleanest approach to add this functionality. Myself and my team will handle this. Thank you for the attempt!

cardano-api/internal/Cardano/Api/Script.hs
mmontin: Yes i can. I basically just mimicked how other types around were defined. But indeed GADTs are not necessary there.

mmontin:
@Jimbo4350 This was just the very beginning and I was in need for feedback but that's fine by me if you handle it.

Jimbo4350:
@mmontin https://github.com/IntersectMBO/cardano-api/pull/640

PR #642, opened 2024-09-23
Remove redundant ScriptHash in ReferenceScript datatypes

### LINKED ISSUE #606: Remove redundant `Maybe ScriptHash` from `PReferenceScript`
Hello, I have been asked a question "what is the script hash used for in `PReferenceScript` and when can you set it to `Nothing`"?

Currently the type `PlutusScriptOrReferenceInput` looks like this:
```
data PlutusScriptOrReferenceInput lang
  = PScript (PlutusScript lang)
  | -- | Needed to construct the redeemer pointer map
    -- in the case of minting reference scripts where we don't
    -- have direct access to the script
    PReferenceScript
      TxIn
      (Maybe ScriptHash)
```

It seems that, despite the comment, the script hash is nowhere used; I cloned the repo and removed it, adjusted the pattern match in `scriptWitnessScript`, and everything else just compiled. 

The question is, can we remove the field? I'd gladly prepare the PR if we can.

Thanks!

Swordlash:
# Changelog

```yaml
- description: |
    Remove redundant ScriptHash in ReferenceScript datatypes
  type:
  - breaking       
  - refactoring
```

# Context

Sometime ago I opened an issue #606 but I didn't get any response, so I've made a PR myself.
One of the members of my team asked why is the `Maybe ScriptHash` needed in the `PReferenceScript` and after some digging it seems it's completely redundant and not used anywhere (with the comment being obsolete). Therefore I removed it and everything "just compiled".

Fixes #606 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/Script.hs
+++ b/cardano-api/internal/Cardano/Api/Script.hs
@@ -681,17 +681,12 @@ data WitCtx witctx where
 -- or to mint tokens. This datatype encapsulates this concept.
 data PlutusScriptOrReferenceInput lang
   = PScript (PlutusScript lang)
-  | -- | Needed to construct the redeemer pointer map
-    -- in the case of minting reference scripts where we don't
-    -- have direct access to the script
-    PReferenceScript
-      TxIn
-      (Maybe ScriptHash)
+  | PReferenceScript TxIn 
   deriving (Eq, Show)
 
 data SimpleScriptOrReferenceInput lang
   = SScript SimpleScript
-  | SReferenceScript TxIn (Maybe ScriptHash)
+  | SReferenceScript TxIn
   deriving (Eq, Show)
 
 -- | A /use/ of a script within a transaction body to witness that something is
@@ -791,9 +786,9 @@ scriptWitnessScript (SimpleScriptWitness SimpleScriptInConway (SScript script)) 
   Just $ ScriptInEra SimpleScriptInConway (SimpleScript script)
 scriptWitnessScript (PlutusScriptWitness langInEra version (PScript script) _ _ _) =
   Just $ ScriptInEra langInEra (PlutusScript version script)
-scriptWitnessScript (SimpleScriptWitness _ (SReferenceScript _ _)) =
+scriptWitnessScript (SimpleScriptWitness _ (SReferenceScript _)) =
   Nothing
-scriptWitnessScript (PlutusScriptWitness _ _ (PReferenceScript _ _) _ _ _) =
+scriptWitnessScript (PlutusScriptWitness _ _ (PReferenceScript _) _ _ _) =
   Nothing
 
 -- ----------------------------------------------------------------------------
--- a/cardano-api/internal/Cardano/Api/Script.hs
+++ b/cardano-api/internal/Cardano/Api/Script.hs
@@ -681,7 +681,7 @@ data WitCtx witctx where
 -- or to mint tokens. This datatype encapsulates this concept.
 data PlutusScriptOrReferenceInput lang
   = PScript (PlutusScript lang)
-  | PReferenceScript TxIn 
+  | PReferenceScript TxIn
   deriving (Eq, Show)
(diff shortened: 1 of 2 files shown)


smelc:
@Swordlash> can you fix the formatting of Haskell files? You need to run:

```shell
# order matters: fourmolu first, then stylish-haskell
fourmolu -i <modified files>
stylish-haskell -i <modified files>
```

Swordlash:
Formatting should be good now. Redundant whitespace it seems.

carbolymer:
@Swordlash can you squash the commits?
The change looks good for me, but I'll leave final approval to @Jimbo4350.

cardano-api/internal/Cardano/Api/Script.hs
carbolymer, line 1: Those are used in
https://github.com/IntersectMBO/cardano-cli/blob/a157c492a4664b0ff4216e1c54a87c04adb0067b/cardano-cli/src/Cardano/CLI/EraBased/Run/Transaction.hs#L1381 to gather minting witnesses here: https://github.com/IntersectMBO/cardano-cli/blob/a157c492a4664b0ff4216e1c54a87c04adb0067b/cardano-cli/src/Cardano/CLI/EraBased/Run/Transaction.hs#L1364

carbolymer (changes requested):
It's used in cardano-cli in handling of minting scripts.

cardano-api/internal/Cardano/Api/Script.hs
```diff
@@ -681,17 +681,12 @@ data WitCtx witctx where
 -- or to mint tokens. This datatype encapsulates this concept.
 data PlutusScriptOrReferenceInput lang
   = PScript (PlutusScript lang)
-  | -- | Needed to construct the redeemer pointer map
-    -- in the case of minting reference scripts where we don't
-    -- have direct access to the script
-    PReferenceScript
-      TxIn
-      (Maybe ScriptHash)
```
carbolymer, line 689: The purpose of this field should be documented better.

cardano-api/internal/Cardano/Api/Script.hs
Swordlash: This is some very unfortunate coupling, isn't it? Having a library keep redundant information (and confuse users) for the sake of one of downstream tools. Maybe we could refactor it out in `cardano-cli`?

Swordlash: Of course we can argue that if the user wants to reference a script then he probably knows the hash. If the above is not possible I would at least not make it `Maybe`, for the sake of API cleanliness.

Swordlash: @carbolymer I did the refactor, see the following: https://github.com/Swordlash/cardano-cli/commit/b5c76a012a0daa5d83dcce738dd25a1810b7b802

User provides the `Maybe PolicyId` himself anyway, so `cardano-cli` just temporarily stores this info in the `PReferenceScript` to be unpacked later. We can change the signature of the appropriate function to return a pair, so that we avoid storing the information in the datatype.

Swordlash: @carbolymer did you have a chance to take a look at the above? Tl;dr `cardano-cli` uses this field as a temporary container for user provided data and it can be safely refactored out.

carbolymer: Dzięki. Looks sensible. Would you mind opening a PR to cardano-cli with those changes?

Swordlash: Sure thing

Swordlash: @carbolymer here it is: https://github.com/IntersectMBO/cardano-cli/pull/918

carbolymer:
Superseded by 
- https://github.com/IntersectMBO/cardano-api/pull/663

Issue #608, opened 2024-08-06
[FEATURE] - Documentation and Integration Clarification 


nhenin:
**Internal/External** : *Internal* 
## Context  
The current repository lacks documentation, making it unclear what direction Intersect is taking with it.

When the repository is updated, projects that depend on it face difficulties upgrading due to the following reasons, among others:
  - `cardano-node` bypasses the dependencies `cardano-api` is supposed to abstract. 
  -  Similar concepts but from different core domains are mixed : e.g : `Ledger.Coin` and `Cardano.API.Lovelace` (that has been removed in 8.40 and not explained in the changelog > see #607 )
 
Here is the current documentation of the Readme :
 
![image](https://github.com/user-attachments/assets/9db6f86e-0885-4745-9904-29bd5784d78b)

I am requesting the addition of comprehensive documentation to assist projects relying on this dependency in progressing with their own development.

## What is the "vision" of this repository?

The repository mentions the **integration of the [ledger](https://github.com/IntersectMBO/cardano-ledger), [consensus](https://github.com/IntersectMBO/ouroboros-consensus), and [networking](https://github.com/IntersectMBO/ouroboros-network/tree/master/ouroboros-network) repositories**.

Is `cardano-api` intended to be the interface for handling Cardano-related tasks? If so, why do some projects use `cardano-api` but also directly access the repositories behind the API?

I was expecting to see something more cohesive, like this:
![image](https://github.com/user-attachments/assets/8cdd7129-2af5-4955-a378-d61c2bf59897)

But I can see that most of the repositories using this API are bypassing it : 
### Cardano Node and Cardano Submit API
![image](https://github.com/user-attachments/assets/79be0ee7-4522-4936-b424-6bc36c6fffc8)
### Cardano CLI
![image](https://github.com/user-attachments/assets/793998b9-15e2-4bd8-abe9-f3d6a3eae078)

These issues impact projects built on top of these libraries. I believe it starts with a clarification on the project vision. What is `cardano-api` aimed to be ?

#### Repository Vision Detailed

The README should clearly outline the vision of the repository, including its purpose and long-term goals. This section should also clarify the intended role of `cardano-api` in handling Cardano-related tasks, addressing why some projects might use both `cardano-api` and the underlying repositories directly.

The current `cardano-node` documentation links should be removed , it is irrelevant here.

#### High-Level Features and Examples of Code

The README should include:
- A detailed description of the features the repository aims to provide.
[...]

PR #656, opened 2024-10-17
Expose remaining functions and types from `cardano-ledger-*` required by `cardano-cli`

palas:
# Changelog

```yaml
- description: |
    Expose transitive dependencies from `cardano-ledger` needed by `cardano-cli`
    Breaking change: `Cardano.Api.Ledger` no longer exports Byron's `Tx` but it is now exported in `Cardano.Api.Byron`
  type:
```

# Context

As part of the work targeted at addressing issues pointed out in https://github.com/IntersectMBO/cardano-api/issues/608, this PR aims to provide `cardano-cli` with everything it needs from ledger.

# How to trust this PR

Make sure it is just a refactoring.

Look at it in conjunction with: https://github.com/IntersectMBO/cardano-cli/pull/941

Take advantage that there are one commit for `shelley`, one for `core`, and one for `api` (same for the corresponding PR in cardano-cli)

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/ReexposeLedger.hs
+++ b/cardano-api/internal/Cardano/Api/ReexposeLedger.hs
@@ -46,24 +46,27 @@ module Cardano.Api.ReexposeLedger
   , Coin (..)
   , EraPParams (..)
   , Era (..)
-  , ExUnits (..)
+  , EraTxOut
   , Network (..)
   , PoolCert (..)
   , PParams (..)
   , PParamsUpdate
+  , TxId (..)
+  , TxIn (..)
   , Value
   , addDeltaCoin
   , castSafeHash
   , toDeltaCoin
   , toEraCBOR
   , fromEraCBOR
+  , ppMinFeeAL
   , ppMinUTxOValueL
-  , unData
   -- Conway
   , Anchor (..)
   , Delegatee (..)
   , DRep (..)
   , DRepState (..)
+  , ConwayPlutusPurpose (..)
   , ConwayTxCert (..)
   , ConwayDelegCert (..)
   , ConwayEraTxCert (..)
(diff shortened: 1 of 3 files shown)


cardano-api/src/Cardano/Api/Byron.hs
```diff
@@ -154,14 +155,14 @@ module Cardano.Api.Byron
   )
 where
 
-import           Cardano.Api hiding (Address, Certificate, Lovelace, NetworkMagic, TxIn, TxOut,
-                   UTxO (..))
+import           Cardano.Api hiding (Address, Certificate, Lovelace, NetworkMagic, Tx (..), TxIn,
```
Jimbo4350, line 158: Not for this PR but we shouldn't be importing `Cardano.Api` here. We need to import the types from their relevant internal modules.

cardano-api/internal/Cardano/Api/ReexposeLedger.hs
```diff
@@ -88,7 +94,6 @@ module Cardano.Api.ReexposeLedger
   -- Byron
   , Annotated (..)
   , byronProtVer
-  , Byron.Tx (..)
```
Jimbo4350, line 91: Why is this removed?

Jimbo4350 (approved):

cardano-api/internal/Cardano/Api/ReexposeLedger.hs
palas: Because it clashes with the `Tx` from `Cardano.Ledger.Alonzo.Core` that I had to re-export for removing `cardano-ledger-api` dependency. On the other hand, it seemed to make sense because of what you said that Byron stuff should be in `Cardano.Api.Byron`. And [here](https://github.com/IntersectMBO/cardano-cli/pull/941/commits/c35228407cd7de8685ca693f2d153dedf618ed1e#diff-87ab989fc0540162119ac2e0e294e56a86fdb9abe26cfdc5a0b27e19849ba5b3R129), for example, it was the only thing that wasn't being imported from `Cardano.Api.Byron`, so it seems more consistent.

PR #660, opened 2024-10-22
Switch from cryptonite library to crypton

erikd:
The former library is deprecated in favor of the latter which is a drop in replacement.

Some of cardano-api's dependencies might still depend on the former but they will all be fixed eventually.

# Changelog

```yaml
- description: |
    Switch from cryptonite library to crypton
  type:
    - compatible     # the API has not changed
```

# Context

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/cardano-api.cabal
+++ b/cardano-api/cardano-api.cabal
@@ -252,7 +252,7 @@ library
     cardano-crypto,
     cardano-crypto-class ^>=2.1.2,
     cardano-ledger-byron,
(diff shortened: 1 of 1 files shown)


carbolymer (approved):

erikd:
This can't be done/fixed until the lower level libs (`cardano-crypto`, `plutus`, `ledger` etc) have been updated.

Closing this.

PR #664, opened 2024-10-25
Add function to extract anchor data from certificate

palas:
# Changelog

```yaml
- description: |
    - Add function to extract anchor data from a certificate
  type:
  - feature
```

# Context

This PR is part of addressing [this review comment](https://github.com/IntersectMBO/cardano-cli/pull/951#discussion_r1815183050). It moves the extracting of `anchor data` into `cardano-api` together with the rest of code that inspects certificates for easier maintenance.

## Related PRs

- [cardano-cli] https://github.com/IntersectMBO/cardano-cli/pull/951
- [cardano-node] https://github.com/IntersectMBO/cardano-node/pull/6023

# How to trust this PR

It is quite analogous to `filterUnRegDRepCreds`, except it does some conversion from `PoolMetadata` to `AnchorData` which are analogous types.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/Certificate.hs
+++ b/cardano-api/internal/Cardano/Api/Certificate.hs
@@ -743,7 +743,7 @@ instance Error AnchorDataFromCertificateError where
   prettyError (InvalidPoolMetadataHashError url hash) =
     "Invalid pool metadata hash for URL " <> fromString (show url) <> ": " <> fromString (show hash)
 
--- | Get anchor data hash from a certificate. A return value of `Nothing`
+-- | Get anchor data url and hash from a certificate. A return value of `Nothing`
 -- means that the certificate does not contain anchor data.
 getAnchorDataFromCertificate
   :: Certificate era
--- a/cardano-api/internal/Cardano/Api/Certificate.hs
+++ b/cardano-api/internal/Cardano/Api/Certificate.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GADTs #-}
+{-# LANGUAGE InstanceSigs #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE ScopedTypeVariables #-}
@@ -56,6 +57,10 @@ module Cardano.Api.Certificate
   , Ledger.MIRPot (..)
   , selectStakeCredentialWitness
 
+    -- * Anchor data
+  , AnchorDataFromCertificateError (..)
+  , getAnchorDataFromCertificate
+
     -- * Internal conversion functions
   , toShelleyCertificate
   , fromShelleyCertificate
@@ -77,10 +82,12 @@ import           Cardano.Api.Eon.ConwayEraOnwards
 import           Cardano.Api.Eon.ShelleyBasedEra
 import           Cardano.Api.Eon.ShelleyToBabbageEra
 import           Cardano.Api.Eras
+import           Cardano.Api.Error (Error (..))
 import           Cardano.Api.Governance.Actions.VotingProcedure
 import           Cardano.Api.HasTypeProxy
 import           Cardano.Api.Keys.Praos
 import           Cardano.Api.Keys.Shelley
+import           Cardano.Api.Pretty (Doc)
 import           Cardano.Api.ReexposeLedger (EraCrypto, StandardCrypto)
 import qualified Cardano.Api.ReexposeLedger as Ledger
 import           Cardano.Api.Script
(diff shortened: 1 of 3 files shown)


Jimbo4350 (changes requested):
Let's stick with `Either` please. I don't want a mixture of `MonadError` and `Either`.

cardano-api/internal/Cardano/Api/Certificate.hs
```diff
@@ -724,3 +733,63 @@ fromShelleyPoolParams
...
+data AnchorDataFromCertificateException
+  = InvalidPoolMetadataHash Ledger.Url ByteString
+  deriving (Eq, Show)
+
+instance Exception AnchorDataFromCertificateException where
+  displayException :: AnchorDataFromCertificateException -> String
+  displayException (InvalidPoolMetadataHash url hash) =
+    "Invalid pool metadata hash for URL " <> show url <> ": " <> show hash
+
+-- | Get anchor data hash from a certificate
```
carbolymer, line 746: Can you explain the returned value in the haddock, I mean when the `Nothing` is returned?

cardano-api/internal/Cardano/Api/Certificate.hs
```diff
@@ -724,3 +733,63 @@ fromShelleyPoolParams
...
+getAnchorDataFromCertificate
+  :: MonadError AnchorDataFromCertificateException m
+  => Certificate era
+  -> m (Maybe (Ledger.Anchor StandardCrypto))
+getAnchorDataFromCertificate =
+  \case
+    ShelleyRelatedCertificate _ shelleyCert ->
+      case shelleyCert of
+        Ledger.ShelleyTxCertDelegCert shelleyDelegCert ->
```
carbolymer, line 755: Can you use pattern synonyms, instead of the constructors here?

palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/664/commits/acd50a481b8fd3dd6d236ecbe5017a51c98e9131

cardano-api/internal/Cardano/Api/Certificate.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/664/commits/acd50a481b8fd3dd6d236ecbe5017a51c98e9131

cardano-api/internal/Cardano/Api/Certificate.hs
```diff
@@ -724,3 +735,73 @@ fromShelleyPoolParams
     fromShelleyDnsName =
       Text.encodeUtf8
         . Ledger.dnsToText
+
+data AnchorDataFromCertificateException
+  = InvalidPoolMetadataHash Ledger.Url ByteString
```
Jimbo4350, line 740: I disagree with the name `AnchorDataFromCertificateException`. This is not really an exceptional circumstance.

carbolymer: I agree. I think it should have an `Error` suffix. Also `Exception` instance isn't needed - we're not throwing it anywhere. I think we should only provide `Error` instance (with `prettyError`).

cardano-api/internal/Cardano/Api/Certificate.hs
```diff
@@ -724,3 +735,73 @@ fromShelleyPoolParams
...
+  -> Either AnchorDataFromCertificateException (Maybe (Ledger.Anchor StandardCrypto))
+getAnchorDataFromCertificate c =
+  case c of
+    ShelleyRelatedCertificate stbe scert ->
+      shelleyToBabbageEraConstraints stbe $ getAnchorDataFromShelleyCertificate scert
+    ConwayCertificate ceo ccert ->
+      conwayEraOnwardsConstraints ceo $ getAnchorDataFromConwayCertificate ccert
+ where
+  getAnchorDataFromShelleyCertificate
+    :: (Ledger.ProtVerAtMost era 8, Ledger.ShelleyEraTxCert era)
```
carbolymer, line 761:
```haskell
    :: (Ledger.ProtVerAtMost era 8, Ledger.ShelleyEraTxCert era)
```
A small nitpick: using a hardcoded `ProtVer` here is a leaking abstraction for me. I'd remove type signatures from `getAnchorDataFrom{Shelley,Conway}Certificate` altogether, because the functions are quite obvious in what they do.

Feel free to use your current version of the code if you like it better.

cardano-api/internal/Cardano/Api/Certificate.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/664/commits/8c699b11232e32292900bcf5fe2b5111f30ec26b

cardano-api/internal/Cardano/Api/Certificate.hs
palas: Unfortunately it won't compile without a signature. But you are right, it is a leaking abstraction, so I will inline it. There should be a type that is abstract enough, but I couldn't find it

palas: Addressed here: https://github.com/IntersectMBO/cardano-api/commit/240393bf8da6b0a315d7b6be275c934eabf64636

Jimbo4350 (approved):

PR #666, opened 2024-10-30
Parameterize Value parser on role of the Value being parsed: transaction output or minting policy

smelc:
# Changelog

```yaml
- description: |
    Parameterize Value parser on role of the Value being parsed: transaction output or minting policy
  type:
```

# Context

* API part of fixing https://github.com/IntersectMBO/cardano-cli/issues/904 in the CLI
* CLI PR using this: https://github.com/IntersectMBO/cardano-cli/issues/904

# How to trust this PR

1. This PR adds more checks
2. Existing tests still pass (and there are a number of them!)

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/ValueParser.hs
+++ b/cardano-api/internal/Cardano/Api/ValueParser.hs
@@ -4,6 +4,7 @@ module Cardano.Api.ValueParser
   ( parseValue
   , assetName
   , policyId
+  , ValueRole (..)
   )
 where
 
@@ -13,6 +14,7 @@ import           Cardano.Api.Utils (failEitherWith)
 import           Cardano.Api.Value
 
 import           Control.Applicative (many, some, (<|>))
+import           Control.Monad (unless, when)
 import qualified Data.ByteString.Char8 as BSC
 import qualified Data.Char as Char
 import           Data.Functor (void, ($>))
@@ -26,9 +28,36 @@ import           Text.Parsec.Expr (Assoc (..), Operator (..), buildExpressionPar
 import           Text.Parsec.String (Parser)
 import           Text.ParserCombinators.Parsec.Combinator (many1)
 
--- | Parse a 'Value' from its string representation.
-parseValue :: Parser Value
-parseValue = evalValueExpr <$> parseValueExpr
+-- | The role for which a 'Value' is being parsed.
+data ValueRole
+  = -- | The value is used as a UTxO or transaction output.
+    RoleUTxO
+  | -- | The value is used as a minting policy.
+    RoleMint
+  deriving (Eq, Show, Enum, Bounded)
+
+-- | Parse a 'Value' from its string representation. The @role@ argument for which purpose
+-- the value is being parsed. This is used to enforce additional constraints on the value.
+-- Why do we parse a general value and check for additional constraints you may ask?
+-- Because we can't rule out the negation operator
+-- for transaction outputs: some users have negative values in additions, with the addition's total
+-- summing up to a positive value. So forbidding negations altogether is too restrictive.
+parseValue :: ValueRole -> Parser Value
+parseValue role = do
+  valueExpr <- parseValueExpr
+  let value = evalValueExpr valueExpr
+  case role of
+    RoleUTxO -> do
+      unless (allPositive value) $
+        fail $
+          "Value must be positive in UTxO (or transaction output): " <> show value
+      return value
+    RoleMint -> do
+      let (Coin lovelace) = selectLovelace value
+      when (lovelace /= 0) $
+        fail $
+          "Lovelace must be zero in minting value: " <> show value
+      return value
 
 -- | Evaluate a 'ValueExpr' and construct a 'Value'.
 evalValueExpr :: ValueExpr -> Value
@@ -170,6 +199,6 @@ assetId =
   assetIdNoAssetName :: PolicyId -> Parser AssetId
   assetIdNoAssetName polId = pure (AssetId polId "")
 
--- | Quantity (word64) parser.
+-- | Quantity (word64) parser. Only accepts positive quantities.
 quantity :: Parser Quantity
 quantity = fmap Quantity word64
--- a/cardano-api/internal/Cardano/Api/ValueParser.hs
+++ b/cardano-api/internal/Cardano/Api/ValueParser.hs
@@ -2,6 +2,9 @@
 
 module Cardano.Api.ValueParser
   ( parseValue
+  , parseTxOutMultiAssetValue
+  , parseMintingMultiAssetValue
+  , parseUTxOValue
   , assetName
   , policyId
   , ValueRole (..)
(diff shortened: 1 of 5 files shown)


cardano-api/internal/Cardano/Api/ValueParser.hs
```diff
@@ -25,10 +26,32 @@ import           Text.Parsec.Char (alphaNum, char, digit, hexDigit, space, space
...
+
+data ValueRole
+  = RoleTxOut
+  | RoleMint
+  deriving (Eq, Show, Enum, Bounded)
 
 -- | Parse a 'Value' from its string representation.
-parseValue :: Parser Value
-parseValue = evalValueExpr <$> parseValueExpr
+-- Why do we parse a general value and check for additional constraints you may ask?
```
carbolymer, line 37: Can you add a comment to `ValueRole` with explanation of the difference?

cardano-api/internal/Cardano/Api/ValueParser.hs
```diff
@@ -25,10 +26,32 @@ import           Text.Parsec.Char (alphaNum, char, digit, hexDigit, space, space
 import           Text.Parsec.Expr (Assoc (..), Operator (..), buildExpressionParser)
 import           Text.Parsec.String (Parser)
 import           Text.ParserCombinators.Parsec.Combinator (many1)
+import Control.Monad (when, unless)
+
+data ValueRole
+  = RoleTxOut
```
carbolymer, line 32: I'm also wondering about this usage site: https://github.com/IntersectMBO/cardano-cli/blob/447fa87001663f942a8f79252876b355fa1f0a04/cardano-cli/src/Cardano/CLI/EraBased/Options/Common.hs#L2046

It's a parser of a whole utxo value, so technically it's not Tx out value. It should work with `RoleTxOut` case of the parser, but the name should be different I think.

cardano-api/internal/Cardano/Api/ValueParser.hs
smelc: Done :+1:

cardano-api/internal/Cardano/Api/ValueParser.hs
smelc: @carbolymer> What do you prefer:

1. Rename `RoleTxOut` into `RoleTxOutOrUtxo`
2. Add a `RoleUtxo`

I'm tempted by option 2., in which case we could check that the quantity is positive and the only `AssetId` is ada?

cardano-api/internal/Cardano/Api/ValueParser.hs
```diff
@@ -4,6 +4,7 @@ module Cardano.Api.ValueParser
   ( parseValue
   , assetName
   , policyId
```
carbolymer, line 6: Can you rename those to indicate that they're parsers? They're too generic and are causing name conflicts frequently.

cardano-api/internal/Cardano/Api/ValueParser.hs
carbolymer, line 1: In line 67, there's a case match:
```haskell
ValueExprMultiAsset polId aName quant -> [(AssetId polId aName, quant)]
```
could you provide a parser
```haskell
Parser (PolicyId, AssetName, Quantity)
```
?

(It could be also `Parser (PolicyId, AssetName, Quantity)` - whatever is more convenient here).

That would be very useful for parsing values for minting.

This can be done in a follow up PR.

carbolymer: If you don't plan to work on it, I can implement that after you finish your work.

cardano-api/internal/Cardano/Api/ValueParser.hs
carbolymer: I'm liking the second option. 
> the only AssetId is ada?
UTXO can contain also non-ada, right?

cardano-api/internal/Cardano/Api/ValueParser.hs
smelc: @carbolymer> I want to have @Jimbo4350's feedback before amending this PR.

Jimbo4350: I was thinking of a change like this: [https://github.com/IntersectMBO/cardano-api/compare/master...jordan/delete-me-value-parser-test](https://github.com/IntersectMBO/cardano-api/compare/master...jordan/delete-me-value-parser-test)

```
ghci> parse parseMintingMultiAssetValue "" "-5 51936f3c98a04b6609aa9b5c832ba1182cf43a58e534fcc05db09d69.4D696C6C6172436F696E"
Right (valueFromList [(AssetId "51936f3c98a04b6609aa9b5c832ba1182cf43a58e534fcc05db09d69" "MillarCoin",-5)])
ghci> parse  parseTxOutMultiAssetValue "" "-5 51936f3c98a04b6609aa9b5c832ba1182cf43a58e534fcc05db09d69.4D696C6C6172436F696E"
Left (line 1, column 1):
unexpected "-"
expecting multi-asset value expression
```

cardano-api/internal/Cardano/Api/ValueParser.hs
smelc: @carbolymer> yes I checked and UTxO can contain non-ada. So I added the new `RoleUTxO` but keps the same checks as `RoleTxOut`.

carbolymer: I'm not sure that duplicated role makes sense here. You will always be able to put things which are in UTXO into TX output.

I think I would just stick with `RoleUTxO`, since all tx outputs will be "unspent".

cardano-api/internal/Cardano/Api/ValueParser.hs
carbolymer: My main problem with the parser is it's return type. `Value` is too broad for minting - you shouldn't be able to keep ADA there.

cardano-api/internal/Cardano/Api/ValueParser.hs
smelc: @carbolymer> Only kept `RoleUTxO` :+1:

carbolymer (approved):

cardano-api/internal/Cardano/Api/ValueParser.hs
carbolymer: Regarding, negative values, I think we still need them for burning tokens. https://developers.cardano.org/docs/native-tokens/minting/#burning-token

cardano-api/internal/Cardano/Api/ValueParser.hs
```diff
@@ -26,9 +31,43 @@ import           Text.Parsec.Expr (Assoc (..), Operator (..), buildExpressionPar
...
+-- for transaction outputs: some users have negative values in additions, with the addition's total
+-- summing up to a positive value. So forbidding negations altogether is too restrictive.
+parseValue :: ValueRole -> Parser Value
+parseValue role = do
+  valueExpr <- parseValueExpr
+  let value = evalValueExpr valueExpr
+  case role of
+    RoleUTxO -> do
+      unless (allPositive value) $
+        fail "Value must be positive in an UTxO (or transaction output)"
```
Jimbo4350, line 55: I would return the value as well in the error messages.

Jimbo4350 (approved):

cardano-api/internal/Cardano/Api/ValueParser.hs
smelc: Showed the value :+1: 

```diff
@@ -52,12 +52,12 @@ parseValue role = do
   case role of
     RoleUTxO -> do
       unless (allPositive value) $
-        fail "Value must be positive in an UTxO (or transaction output)"
+        fail $ "Value must be positive in UTxO (or transaction output): " <> show value
       return value
     RoleMint -> do
       let (Coin lovelace) = selectLovelace value
       when (lovelace /= 0) $
-        fail "Lovelace must be zero in a minting value"
+        fail $ "Lovelace must be zero in minting value: " <> show value
       return value
```

Issue #673, opened 2024-11-05
Rename functions in ValueParser.hs, to indicate in the name they are parsers


smelc:
Can you rename those to indicate that they're parsers? They're too generic and are causing name conflicts frequently.

_Originally posted by @carbolymer in https://github.com/IntersectMBO/cardano-api/pull/666#discussion_r1828100824_

PR #674, opened 2024-11-06
ValueParser: rename publicly exposed function names to indicate they are parsers

smelc:
# Changelog

```yaml
- description: |
    ValueParser: rename publicly exposed function names to indicate they are parsers
    
    To adapt: prefix old function name by `parse` and everything should compile again.
  type:
```

# Context

* Fixes https://github.com/IntersectMBO/cardano-api/issues/673
* Most functions in `ValueParser` that have a `-> Parser _` return type had nothing in their name indicating that they do so. Those functions' name now start with `parse`.

# How to trust this PR

It's only renamings

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/ValueParser.hs
+++ b/cardano-api/internal/Cardano/Api/ValueParser.hs
@@ -5,9 +5,9 @@ module Cardano.Api.ValueParser
   , parseTxOutMultiAssetValue
   , parseMintingMultiAssetValue
   , parseUTxOValue
-  , assetName
-  , policyId
-  , ValueRole (..)
+  , parseAssetName
+  , parsePolicyId
+  , ParserValueRole (..)
   )
 where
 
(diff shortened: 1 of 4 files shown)


smelc:
The third commit is optional, since it only concerns names that are not exposed.

cardano-api/internal/Cardano/Api/ValueParser.hs
```diff
@@ -5,8 +5,8 @@ module Cardano.Api.ValueParser
   , parseTxOutMultiAssetValue
   , parseMintingMultiAssetValue
   , parseUTxOValue
-  , assetName
-  , policyId
+  , parseAssetName
+  , parsePolicyId
   , ValueRole (..)
```
carbolymer, line 10: Because it's only used in `ValueParser` and exported from this module, maybe it should be more specific as well?
```suggestion
  , ParserValueRole (..)
```
Otherwise we could move it near `Value` definition to make it more widely used. But I have doubts about its utility outside of parsing.

Jimbo4350 (approved):

PR #678, opened 2024-11-12
Add mnemonic support

palas:
# Changelog

```yaml
- description: |
    Added support for generating mnemonics and for deriving payment and stake keys from mnemonics.
  type:
  - feature
```

# Context

Depends on release of `cardano-addresses` (https://github.com/IntersectMBO/cardano-addresses/pull/279)
Allows https://github.com/IntersectMBO/cardano-cli/pull/975.

# How to trust this PR

I would look at the tests. The hard-coded tests have been checked against existing wallets.
I have also tested that the derivation works by doing a prototype command in the `cardano-cli`.
Make sure you like the changes to the API. I have tried to keep them minimal.
Probably the most controversial bit are the instances I added, but they are not exposed. The reason I parametrised the index type is that for some types of keys that I haven't added yet (because I am waiting for a `cardano-addresses` release) the payment is not necessary, so I plan to set it to `()` in those.
Also make sure the documentation is good enough.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -13,7 +13,7 @@ repository cardano-haskell-packages
 -- See CONTRIBUTING for information about these, including some Nix commands
 -- you need to run if you change them
 index-state:
-  , hackage.haskell.org 2025-03-03T01:12:46Z
+  , hackage.haskell.org 2025-03-17T12:01:29Z
   , cardano-haskell-packages 2025-03-03T04:46:42Z
 
 packages:
--- a/cardano-api/cardano-api.cabal
+++ b/cardano-api/cardano-api.cabal
@@ -110,8 +110,11 @@ library
     base16-bytestring >=1.0,
     base58-bytestring,
     base64-bytestring,
+    basement,
     bech32 >=1.1.0,
     bytestring,
+    bytestring-trie,
+    cardano-addresses ^>=4.0.0,
     cardano-binary,
     cardano-crypto,
     cardano-crypto-class ^>=2.1.2,
@@ -139,6 +142,7 @@ library
     directory,
     either,
     errors,
+    extra,
     filepath,
     formatting,
     groups,
@@ -160,7 +164,7 @@ library
     ouroboros-network-framework,
     ouroboros-network-protocols,
     parsec,
-    plutus-ledger-api:{plutus-ledger-api, plutus-ledger-api-testlib} ^>=1.37,
+    plutus-ledger-api:{plutus-ledger-api, plutus-ledger-api-testlib} ^>=1.40,
     prettyprinter,
     prettyprinter-ansi-terminal,
     prettyprinter-configurable ^>=1.36,
@@ -221,6 +225,7 @@ library
     Cardano.Api.Internal.Json
     Cardano.Api.Internal.Keys.Byron
     Cardano.Api.Internal.Keys.Class
+    Cardano.Api.Internal.Keys.Mnemonics
     Cardano.Api.Internal.Keys.Praos
     Cardano.Api.Internal.Keys.Read
     Cardano.Api.Internal.Keys.Shelley
@@ -403,7 +408,7 @@ test-suite cardano-api-golden
     hedgehog-extras ^>=0.7,
     microlens,
     parsec,
-    plutus-core ^>=1.37,
+    plutus-core ^>=1.40,
     plutus-ledger-api,
     tasty,
     tasty-hedgehog,
--- a/cardano-api/src/Cardano/Api.hs
+++ b/cardano-api/src/Cardano/Api.hs
@@ -179,6 +179,22 @@ module Cardano.Api
   , castHash
   , renderSafeHashAsHex
 
+    -- * Mnemonics
+
+    -- | Functions for working with mnemonics
+    -- ** Mnemonics generation
+  , MnemonicSize (..)
+  , generateMnemonic
+
+    -- ** Key derivation from mnemonics
+  , MnemonicToSigningKeyError (..)
+  , signingKeyFromMnemonic
+  , signingKeyFromMnemonicWithPaymentKeyIndex
+
+    -- ** Mnemonic word queries
+  , findMnemonicWordsWithPrefix
+  , autocompleteMnemonicPrefix
+
     -- * Payment addresses
 
     -- | Constructing and inspecting normal payment addresses
@@ -1119,6 +1135,7 @@ import Cardano.Api.Internal.IPC.Monad
 import Cardano.Api.Internal.InMode
 import Cardano.Api.Internal.Keys.Byron
 import Cardano.Api.Internal.Keys.Class
+import Cardano.Api.Internal.Keys.Mnemonics
 import Cardano.Api.Internal.Keys.Read
 import Cardano.Api.Internal.Keys.Shelley
 import Cardano.Api.Internal.LedgerState
--- a/cardano-api/src/Cardano/Api/Internal/Keys/Mnemonics.hs
+++ b/cardano-api/src/Cardano/Api/Internal/Keys/Mnemonics.hs
@@ -1,0 +1,360 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE InstanceSigs #-}
+{-# LANGUAGE TupleSections #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+
+module Cardano.Api.Internal.Keys.Mnemonics
+  ( MnemonicSize (..)
+  , generateMnemonic
+  , MnemonicToSigningKeyError (..)
+  , signingKeyFromMnemonic
+  , signingKeyFromMnemonicWithPaymentKeyIndex
+  , findMnemonicWordsWithPrefix
+  , autocompleteMnemonicPrefix
+  )
+where
+
+import Cardano.Api.Internal.Error (Error (..))
+import Cardano.Api.Internal.Keys.Class (Key (..))
+import Cardano.Api.Internal.Keys.Shelley
+  ( AsType
+  , CommitteeColdExtendedKey
+  , CommitteeHotExtendedKey
+  , DRepExtendedKey
+  , PaymentExtendedKey
+  , SigningKey (..)
+  , StakeExtendedKey
+  )
+
+import Cardano.Address.Derivation
+  ( Depth (..)
+  , DerivationType (..)
+  , HardDerivation (..)
+  , Index
+  , XPrv
+  , genMasterKeyFromMnemonic
+  , indexFromWord32
+  )
+import Cardano.Address.Style.Shelley
+  ( Role (..)
+  , Shelley (..)
+  , deriveCCColdPrivateKey
+  , deriveCCHotPrivateKey
+  , deriveDRepPrivateKey
+  )
+import Cardano.Crypto.Encoding.BIP39 (Dictionary (dictionaryIndexToWord))
+import Cardano.Mnemonic
+  ( MkSomeMnemonic (mkSomeMnemonic)
+  , MkSomeMnemonicError (..)
+  , SomeMnemonic
+  , entropyToMnemonic
+  , genEntropy
+  , mnemonicToText
+  )
+
+import Control.Monad.IO.Class (MonadIO, liftIO)
+import Data.Bifunctor (first)
+import Data.ByteString qualified as BS
+import Data.Either.Combinators (mapLeft, maybeToRight)
+import Data.Either.Extra (maybeToEither)
+import Data.Foldable (toList)
+import Data.Text (Text)
+import Data.Text qualified as Text
+import Data.Text.Encoding (decodeUtf8, encodeUtf8)
+import Data.Trie (submap)
+import Data.Trie qualified as Trie
+import Data.Trie.Convenience qualified as Trie
+import Data.Word (Word32)
+import Prettyprinter (Doc, Pretty (..))
+
+import Basement.Compat.IsList qualified as Basement
+import Basement.String qualified as Basement
+import Crypto.Encoding.BIP39.English (english)
+
+-- | The size of a mnemonic sentence.
+-- The size is given in the number of words in the sentence.
+-- The allowed sizes are 12, 15, 18, 21, and 24.
+data MnemonicSize
+  = MS12
+  | MS15
+  | MS18
+  | MS21
+  | MS24
+  deriving (Eq, Show)
+
+-- | Generate a mnemonic sentence of the given size.
+generateMnemonic
+  :: MonadIO m
+  => MnemonicSize
+  -- ^ The size of the mnemonic sentence to generate.
+  -- Must be one of 12, 15, 18, 21, or 24.
+  -> m [Text]
+generateMnemonic MS12 = liftIO (mnemonicToText @12 . entropyToMnemonic <$> genEntropy)
+generateMnemonic MS15 = liftIO (mnemonicToText @15 . entropyToMnemonic <$> genEntropy)
+generateMnemonic MS18 = liftIO (mnemonicToText @18 . entropyToMnemonic <$> genEntropy)
+generateMnemonic MS21 = liftIO (mnemonicToText @21 . entropyToMnemonic <$> genEntropy)
+generateMnemonic MS24 = liftIO (mnemonicToText @24 . entropyToMnemonic <$> genEntropy)
+
+-- | Errors that can occur when converting a mnemonic sentence to a signing key
+data MnemonicToSigningKeyError
+  = InvalidMnemonicError String
+  | InvalidAccountNumberError Word32
+  | InvalidPaymentKeyNoError Word32
+  deriving (Eq, Show)
+
+-- For information about address derivation check:
+--  * https://cips.cardano.org/cip/CIP-1852
+--  * https://github.com/uniVocity/cardano-tutorials/blob/master/cardano-addresses.md#understanding-the-hd-wallet-address-format-bip-44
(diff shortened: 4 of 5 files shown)


cardano-api/internal/Cardano/Api/Keys/Mnemonics.hs
```diff
@@ -0,0 +1,134 @@
...
+  | InvalidPaymentKeyNoError Word32
+  deriving (Eq, Show)
+
+instance Error MnemonicToSigningStakeKeyError where
+  prettyError :: MnemonicToSigningStakeKeyError -> Doc ann
+  prettyError (InvalidMnemonicError str) = "Invalid mnemonic sentence: " <> pretty str
+  prettyError (InvalidAccountNumberError accNo) = "Invalid account number: " <> pretty accNo
+  prettyError (InvalidPaymentKeyNoError keyNo) = "Invalid payment key number: " <> pretty keyNo
+
+class ExtendedSigningKeyRole keyrole indexType where
```
carbolymer, line 71: What's the purpose of the second class parameter? When it will be different than `Word32`? Is it depending on the `keyrole`? If yes, there should be a functional dependency `keyrole -> indexType`.

cardano-api/internal/Cardano/Api/Keys/Mnemonics.hs
```diff
@@ -0,0 +1,134 @@
...
+import           Data.Either.Extra (maybeToEither)
+import           Data.Text (Text)
+import           Data.Word (Word32)
+import           Prettyprinter (Doc, Pretty (..))
+
+-- | The size of a mnemonic sentence.
+-- The size is given in the number of words in the sentence.
+-- The allowed sizes are 12, 15, 18, 21, and 24.
+data MnemonicSize
+  = MS_12
```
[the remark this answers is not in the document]
palas: I copied it from `cardano-addresses`, but we can change it. Still, because it is two words, I think the snake case would be `MS12`

cardano-api/internal/Cardano/Api/Keys/Mnemonics.hs
palas: Yes, it will be `()` for other `keyrole`s. I will look into functional dependencies, I am not familiar.

cardano-api/internal/Cardano/Api/Keys/Mnemonics.hs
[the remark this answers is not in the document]
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/678/commits/2d7999526ec592aec4879aa500a1539ed81bd1cf

cardano-api/internal/Cardano/Api/Keys/Mnemonics.hs
palas: Thanks, that is exactly what I needed, it removed the need for the `ExtendedSigningKeyRole` and the type annotation in the test.
Addressed here: https://github.com/IntersectMBO/cardano-api/pull/678/commits/2316942153d6d21f16ca7fdd6b730e6603ae4469

cardano-api/internal/Cardano/Api/Keys/Mnemonics.hs
```diff
@@ -0,0 +1,134 @@
...
+instance Error MnemonicToSigningStakeKeyError where
+  prettyError :: MnemonicToSigningStakeKeyError -> Doc ann
+  prettyError (InvalidMnemonicError str) = "Invalid mnemonic sentence: " <> pretty str
+  prettyError (InvalidAccountNumberError accNo) = "Invalid account number: " <> pretty accNo
+  prettyError (InvalidPaymentKeyNoError keyNo) = "Invalid payment key number: " <> pretty keyNo
+
+class ExtendedSigningKeyRole keyrole indexType where
+  -- | Derive an extended private key of the keyrole from an account extended private key
+  deriveSigningKeyFromAccount
+    :: AsType keyrole -> Shelley 'AccountK XPrv -> indexType -> Either Word32 (SigningKey keyrole)
```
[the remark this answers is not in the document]
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/678/commits/2316942153d6d21f16ca7fdd6b730e6603ae4469

cardano-api/internal/Cardano/Api/Keys/Mnemonics.hs
```diff
@@ -0,0 +1,170 @@
...
+-- In this function we only ask for two indices: the account number and the
+-- payment key number. Each account can have multiple payment keys.
+signingKeyFromMnemonic
+  :: ExtendedSigningKeyRole keyrole indexType
+  => AsType keyrole
+  -- ^ Type of the extended signing key to generate.
+  -> [Text]
+  -- ^ The mnemonic sentence. The length must be one of 12, 15, 18, 21, or 24.
+  -- Each element of the list must be a single word.
+  -> Word32
```
Jimbo4350, line 145: > IMPORTANT: All desktop/mobile wallets around currently support account 0 only. If you create account 1, 2 or whatever the transactions associated with these accounts will not show up

I think we should default to 0 to avoid potential issues wrt this. If users ask for more flexibility we can expose it later.

cardano-api/internal/Cardano/Api/Keys/Mnemonics.hs
```diff
@@ -0,0 +1,170 @@
...
+  -- ^ The account number in the derivation path. First account is 0.
+  -> indexType
```
Jimbo4350, line 147: > 

    IMPORTANT: Wallets following the BIP-44 standard basically look for 20 transactions after the last transaction received. So if you receive a payment on address generated with address_index = 6, your wallet will only display the money received on addresses from 0 to 26.

    If you received payment on address with address_index = 30, the funds will not be displayed to you even though it's on the blockchain. It will only appear once there is a transaction in some address where address_index is between 10 and 29.

    This applies for all current desktop/mobile/hardware wallets, so be aware. The gap limit can be customized on some wallets, but increasing it reduces synchronization performance.


Something else to keep in mind and add to the comments here.

cardano-api/internal/Cardano/Api/Keys/Mnemonics.hs
```diff
@@ -0,0 +1,170 @@
...
+instance ExtendedSigningKeyRole CommitteeHotExtendedKey () where
+  deriveSigningKeyFromAccount
+    :: AsType CommitteeHotExtendedKey
+    -> Shelley 'AccountK XPrv
+    -> ()
+    -> Either Word32 (SigningKey CommitteeHotExtendedKey)
+  deriveSigningKeyFromAccount _ accK _ =
+    return $ CommitteeHotExtendedSigningKey $ getKey $ deriveCCHotPrivateKey accK
+
+-- | Generate a signing key from a mnemonic sentence.
```
Jimbo4350, line 132: Add https://cips.cardano.org/cip/CIP-1852 and https://github.com/uniVocity/cardano-tutorials/blob/master/cardano-addresses.md#understanding-the-hd-wallet-address-format-bip-44 to this comment.

cardano-api/internal/Cardano/Api/Keys/Mnemonics.hs
```diff
@@ -0,0 +1,170 @@
...
+  | InvalidAccountNumberError Word32
+  | InvalidPaymentKeyNoError Word32
+  deriving (Eq, Show)
+
+instance Error MnemonicToSigningStakeKeyError where
+  prettyError :: MnemonicToSigningStakeKeyError -> Doc ann
+  prettyError (InvalidMnemonicError str) = "Invalid mnemonic sentence: " <> pretty str
+  prettyError (InvalidAccountNumberError accNo) = "Invalid account number: " <> pretty accNo
+  prettyError (InvalidPaymentKeyNoError keyNo) = "Invalid payment key number: " <> pretty keyNo
+
```
Jimbo4350, line 71: I would add https://cips.cardano.org/cip/CIP-1852 and https://github.com/uniVocity/cardano-tutorials/blob/master/cardano-addresses.md#understanding-the-hd-wallet-address-format-bip-44 here as well.

cardano-api/internal/Cardano/Api/Keys/Mnemonics.hs
```diff
@@ -0,0 +1,170 @@
...
+    -> Either Word32 (SigningKey StakeExtendedKey)
+  deriveSigningKeyFromAccount _ accK idx = do
+    payKeyIx <- maybeToEither idx $ indexFromWord32 @(Index 'Soft 'PaymentK) idx
+    return $ StakeExtendedSigningKey $ getKey $ deriveAddressPrivateKey accK Stake payKeyIx
+
+instance ExtendedSigningKeyRole DRepExtendedKey () where
+  deriveSigningKeyFromAccount
+    :: AsType DRepExtendedKey
+    -> Shelley 'AccountK XPrv
+    -> ()
```
Jimbo4350, line 109: Add documentation on why there are no indexes for drep and committee keys (even though it is obvious).

Jimbo4350 (commented):
I made a first pass. I want to look at it again before approving.

cardano-api/internal/Cardano/Api/Keys/Mnemonics.hs
```diff
@@ -0,0 +1,155 @@
...
+  -- without a second factor.
+  -> Word32
+  -- ^ The account number in the derivation path. First account is 0.
+  -> Word32
+  -- ^ The payment key number in the derivation path. First key is 0.
+  -> Either MnemonicToSigningStakeKeyError (SigningKey keyrole)
+signingKeyFromMnemonic role mnemonicWords mSecondFactor accNo payKeyNo = do
+  -- Convert raw types to the ones used in the cardano-addresses library
+  someMnemonic <- mapLeft InvalidMnemonicError $ wordsToSomeMnemonic mnemonicWords
+  secondFactorBytes <- toSecondFactor mSecondFactor
```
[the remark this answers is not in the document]
Jimbo4350: ah I see this unnecessarily brings in `IO`.

cardano-api/internal/Cardano/Api/Keys/Mnemonics.hs
Jimbo4350: But open to feedback here @smelc @carbolymer.

cardano-api/internal/Cardano/Api/Keys/Mnemonics.hs
```diff
@@ -0,0 +1,170 @@
...
+  deriveSigningKeyFromAccount
+    :: AsType keyrole
+    -- ^ Type of the extended signing key to generate.
+    -> Shelley 'AccountK XPrv
+    -- ^ The account extended private key from which to derivate the private key for the keyrole.
+    -> indexType
+    -- ^ The payment key number in the derivation path (as 'Word32') if applicable for
+    -- the given key role, otherwise '()'. First key is 0.
+    -> Either Word32 (SigningKey keyrole)
+    -- ^ The derived extended signing key or the 'indexType' if it is 'Word32' and it is invalid.
```
carbolymer, line 83: Actually, I'm thinking you don't need functional dependency here, you could just use associated type family:

```suggestion
class ExtendedSigningKeyRole keyrole where
  type EskIndex keyrole
  -- | Derive an extended private key of the keyrole from an account extended private key
  deriveSigningKeyFromAccount
    :: AsType keyrole
    -- ^ Type of the extended signing key to generate.
    -> Shelley 'AccountK XPrv
    -- ^ The account extended private key from which to derivate the private key for the keyrole.
    -> EskIndex keyrole
    -- ^ The payment key number in the derivation path (as 'Word32') if applicable for
    -- the given key role, otherwise '()'. First key is 0.
    -> Either Word32 (SigningKey keyrole)
    -- ^ The derived extended signing key or the 'indexType' if it is 'Word32' and it is invalid.
```

The result is pretty much the same, but the type class signature is a bit simpler and an instance for example becomes:

```haskell
instance ExtendedSigningKeyRole PaymentExtendedKey where
  type EskIndex PaymentExtendedKey = Word32
  deriveSigningKeyFromAccount
    :: AsType PaymentExtendedKey
    -> Shelley 'AccountK XPrv
    -> Word32
    -> Either Word32 (SigningKey PaymentExtendedKey)
  deriveSigningKeyFromAccount _ accK idx = do
```

The difference is that you're not exposing the key explicitly in the instance head.

In the end both are very similar, so it's up to you.

carbolymer: More comparisons of functional dependencies and type families if you'd like:
* https://gitlab.haskell.org/ghc/ghc/-/wikis/tf-vs-fd
* https://wiki.haskell.org/Functional_dependencies_vs._type_families

cardano-api/internal/Cardano/Api/Keys/Mnemonics.hs
carbolymer: Where does this limitation come from? Is this an assumption of the wallets, or is it a limitation of the algorithm?
If the former, I think `Word32` is fine here. Otherwise we should default to 0.

palas: It seems Eternl does support multiple accounts:
image

cardano-api/internal/Cardano/Api/Keys/Mnemonics.hs
palas: Oh! I did think that having that type as a parameter to `ExtendedSigningKeyRole` was contrived, but didn't know how to fix it. Let me try that

cardano-api/internal/Cardano/Api/Keys/Mnemonics.hs
palas: Added here: https://github.com/IntersectMBO/cardano-api/pull/678/commits/f7f9c655b3391302d45b054af53f3a94aff22f20

cardano-api/internal/Cardano/Api/Keys/Mnemonics.hs
palas: Added here: https://github.com/IntersectMBO/cardano-api/commit/f7f9c655b3391302d45b054af53f3a94aff22f20

cardano-api/internal/Cardano/Api/Keys/Mnemonics.hs
palas: Added here: https://github.com/IntersectMBO/cardano-api/commit/f7f9c655b3391302d45b054af53f3a94aff22f20

cardano-api/internal/Cardano/Api/Keys/Mnemonics.hs
palas: Added here: https://github.com/IntersectMBO/cardano-api/commit/f7f9c655b3391302d45b054af53f3a94aff22f20

cardano-api/internal/Cardano/Api/Keys/Mnemonics.hs
```diff
@@ -0,0 +1,213 @@
...
+--  * https://github.com/uniVocity/cardano-tutorials/blob/master/cardano-addresses.md#understanding-the-hd-wallet-address-format-bip-44
+--  * https://cips.cardano.org/cip/CIP-0105
+instance Error MnemonicToSigningKeyError where
+  prettyError :: MnemonicToSigningKeyError -> Doc ann
+  prettyError (InvalidMnemonicError str) = "Invalid mnemonic sentence: " <> pretty str
+  prettyError (InvalidAccountNumberError accNo) = "Invalid account number: " <> pretty accNo
+  prettyError (InvalidPaymentKeyNoError keyNo) = "Invalid payment key number: " <> pretty keyNo
+
+class ExtendedSigningKeyRole keyrole where
+  type EskrPaymentAddrIndex keyrole
```
carbolymer, line 77: Can you add a haddock to the type?

carbolymer (commented):
LGTM. I'm leaving approval to @Jimbo4350

cardano-api/internal/Cardano/Api/Keys/Mnemonics.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/678/commits/8c899e34ca29224ab7192208e72e69734848a805

cardano-api/internal/Cardano/Api/Keys/Mnemonics.hs
Jimbo4350: Ok let's leave it as is (i.e not defaulting to 0)

Jimbo4350 (approved):
This LGTM however I think it may be better as two different type classes so we can avoid using `()` and an associated type.

What do you think about `class SigningKeyFromRootKey keyrole` and `class SigningKeyFromRootKeyIndexed keyrole`

smelc:
> This LGTM however I think it may be better as two different type classes so we can avoid using () and an associated type.

While I salute the effort to keep one interface that you did @palas, I agree with @Jimbo4350's suggestion here. I think it'll make the API clearer.

Also big kudos on the large amount of documentation written in the new code :clap:

palas:
@Jimbo4350 @carbolymer So we would need two functions right? Another option I considered was having an algebraic data type as a parameter with all the info. Let me know if you prefer that.

palas:
@Jimbo4350 @smelc So we would need two functions right? Another option I considered was having an algebraic data type as a parameter with all the info. Let me know if you prefer that.

Jimbo4350:
> @Jimbo4350 @smelc So we would need two functions right? Another option I considered was having an algebraic data type as a parameter with all the info. Let me know if you prefer that.

I think two separate classes is better here.

PR #679, opened 2024-11-12
Restore `aarch64` support

palas:
# Changelog

```yaml
- description: |
    Restored `aarch64` support
  type:
  - maintenance
```

# Context

`cardano-cli` already has support `aarch64` and it depends on `cardano-api`, so I am not sure why this is disabled.

# How to trust this PR

If the CI passes I would trust it.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -22,8 +22,8 @@
     supportedSystems = [
       "x86_64-linux"
       "x86_64-darwin"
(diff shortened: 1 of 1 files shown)


smelc (approved):
Indeed these platforms have been running smoothly on CLI since a while, so approving :+1:

carbolymer:
@smelc smooth as butter 😆 
image

PR #684, opened 2024-11-20
Integrate query proposals

CarlosLopezDeLara:
# Changelog

```yaml
- description: |
    Integrate queryProposals (GetProposals)
  type:
```

# Context

This query is relatively new in consensus and will be useful for https://github.com/IntersectMBO/cardano-cli/issues/954

# How to trust this PR

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/Query.hs
+++ b/cardano-api/internal/Cardano/Api/Query.hs
@@ -125,6 +125,7 @@ import           Data.Either.Combinators (rightToMaybe)
 import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
 import           Data.Maybe (mapMaybe)
+import           Data.Sequence (Seq)
 import           Data.Set (Set)
 import qualified Data.Set as Set
 import           Data.SOP.Constraint (SListI)
@@ -296,6 +297,9 @@ data QueryInShelleyBasedEra era result where
   QueryStakeVoteDelegatees
     :: Set StakeCredential
     -> QueryInShelleyBasedEra era (Map StakeCredential (Ledger.DRep StandardCrypto))
+  QueryProposals
+    :: Set (L.GovActionId StandardCrypto)
+    -> QueryInShelleyBasedEra era (Seq (L.GovActionState (ShelleyLedgerEra era)))
 
 -- | Mapping for queries in Shelley-based eras returning minimal node-to-client protocol versions. More
 -- information about queries versioning can be found:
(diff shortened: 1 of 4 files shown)


smelc (approved):

cardano-api/internal/Cardano/Api/Query.hs
```diff
@@ -296,6 +297,9 @@ data QueryInShelleyBasedEra era result where
   QueryStakeVoteDelegatees
     :: Set StakeCredential
     -> QueryInShelleyBasedEra era (Map StakeCredential (Ledger.DRep StandardCrypto))
+  QueryProposals
+    :: Set (L.GovActionId StandardCrypto)
+    -> QueryInShelleyBasedEra era (Seq (L.GovActionState (ShelleyLedgerEra era {-StandardCrypto-})))
```
carbolymer, line 302: accidental leftover comment?
```suggestion
    -> QueryInShelleyBasedEra era (Seq (L.GovActionState (ShelleyLedgerEra era)))
```

cardano-api/internal/Cardano/Api/Query/Expr.hs
```diff
@@ -490,3 +492,23 @@ queryAccountState cOnwards =
   queryExpr $
     QueryInEra . QueryInShelleyBasedEra (inject cOnwards :: ShelleyBasedEra era) $
       QueryAccountState
+
+queryProposals
```
carbolymer, line 496: What happens if you pass an empty set here? Do you get all proposals or none? I can't remember what's the behaviour here. 😄 
Can you add a comment describing the behaviour for an empty set?
(we should do that for the other queries too, but that's out of scope of this PR).

carbolymer (approved):
LGTM! Can you test the query on the live node prior to merging?

cardano-api/internal/Cardano/Api/Query/Expr.hs
```diff
@@ -490,3 +492,25 @@ queryAccountState cOnwards =
...
+      point
+      QueryInMode
+      r
+      IO
+      ( Either
+          UnsupportedNtcVersionError
+          (Either EraMismatch (Seq (L.GovActionState (ShelleyLedgerEra era))))
+      )
+queryProposals cOnwards govActionIds = do
+  let sbe :: ShelleyBasedEra era = inject cOnwards
```
Jimbo4350, line 513: We have https://github.com/IntersectMBO/cardano-api/pull/690 which will allow us to remove this type annotation and the existential quantification!

Jimbo4350 (approved):
LGTM!

smelc:
@carbolymer> I think it'd be nicer for the developer experience to merge this before having tested it on `cardano-testnet`, because testing requires building both `cardano-node` and `cardano-cli` locally with SRPs, which is cumbersome. I don't think there's much harm to be done if we merge this. It's not like 1. it can be very wrong and 2. users are going to jump on this feature.

smelc:
Agreed during the team meeting to merge after the release (the one being worked on by @neilmayhew).

smelc:
Putting to draft to make sure we don't merge until we're sure the behavior we observe in https://github.com/IntersectMBO/cardano-node/pull/6045 is the intended one (as [agreed on Slack](https://input-output-rnd.slack.com/archives/C031GFLC3DF/p1732895386605449?thread_ts=1732869358.126819&cid=C031GFLC3DF), we don't want to merge until we're sure our code (and the ledger's code) is correct).

smelc:
We have confirmed that the behavior of the query we observe on `cardano-testnet` is correct, so merging. cc @CarlosLopezDeLara

PR #619, opened 2024-08-21
Typed Protocols: new API

coot:
# Changelog

```yaml
- description: |
    Updated to `typed-protocols-0.3`, `ouroboros-network-0.18`, and new version of `ouroboros-consensus`.
# uncomment types applicable to the change:
  type:
  # - feature        # introduces a new feature
  # - breaking       # the API has changed in a breaking way
  # - compatible     # the API has changed but is non-breaking
  # - optimisation   # measurable performance improvements
  # - refactoring    # QoL changes
  # - bugfix         # fixes a defect
  # - test           # fixes/modifies tests
    - maintenance    # not directly related to the code
  # - release        # related to a new release preparation
  # - documentation  # change in code docs, haddocks...
```

# Context

This PR introduces new API of `typed-protocols`.

# Checklist

- [x] Commit sequence broadly makes sense and commits have useful messages
- [ ] New tests are added if needed and existing tests are updated. See [Running tests](https://github.com/input-output-hk/cardano-node-wiki/wiki/Running-tests) for more details
- [x] Self-reviewed the diff

PR #687, opened 2024-11-21
Integrate in preparation for `cardano-node` release `10.2.0`

neilmayhew:
# Changelog

```yaml
- description: |
    Integrate `typed-protocols`, `ouroboros-network`, `cardano-ledger` and `ouroboros-consensus`
    in preparation for `cardano-node` release `10.2.0`
  type:
   - breaking
   - feature
```

# Context

See IntersectMBO/cardano-node#6040 for overall progress and links to PRs in other repos

# How to trust this PR

Similar changes have been made in `ouroboros-consensus`. Another commit is cherry-picked from #619.

# Checklist

- [x] Self-reviewed the diff


neilmayhew:
This will need to have the srp removed after new consensus packages are released to CHaP

cardano-api/internal/Cardano/Api/Query.hs
```diff
@@ -156,12 +156,12 @@ data QueryInMode result where
...
+    QueryCurrentEra -> NodeToClientV_16
     QueryInEra q -> nodeToClientVersionOf q
-    QueryEraHistory -> NodeToClientV_9
-    QuerySystemStart -> NodeToClientV_9
-    QueryChainBlockNo -> NodeToClientV_10
-    QueryChainPoint -> NodeToClientV_10
+    QueryEraHistory -> NodeToClientV_16
+    QuerySystemStart -> NodeToClientV_16
+    QueryChainBlockNo -> NodeToClientV_16
+    QueryChainPoint -> NodeToClientV_16
```
carbolymer, line 165: Where does this change come from? I believe those queries were introduced in the earlier eras.

This will probably break test networks which start in older eras and then hardfork to Conway.

cardano-api/internal/Cardano/Api/Query.hs
```diff
@@ -306,23 +306,23 @@ data QueryInShelleyBasedEra era result where
 --   * https://ouroboros-network.cardano.intersectmbo.org/ouroboros-network/Ouroboros-Network-NodeToClient.html#t:NodeToClientVersion
 --   * https://ouroboros-consensus.cardano.intersectmbo.org/docs/for-developers/QueryVersioning/#implementation
 instance NodeToClientVersionOf (QueryInShelleyBasedEra era result) where
-  nodeToClientVersionOf QueryEpoch = NodeToClientV_9
```
carbolymer, line 309: Same comment as for `NodeToClientVersionOf (QueryInMode result)`

cardano-api/internal/Cardano/Api/IPC.hs
```diff
@@ -202,15 +204,19 @@ connectToLocalNodeWithVersion
...
-        (Net.localSnocket iomgr)
-        Net.NetworkConnectTracers
-          { Net.nctMuxTracer = nullTracer
-          , Net.nctHandshakeTracer = nullTracer
-          }
-        versionedProtocls
-        (unFile localNodeSocketPath)
+    liftIO $ Net.withIOManager $ \iomgr -> do
+      r <-
+        Net.connectTo
```
carbolymer, line 209: I'm wondering what's the purpose of the new return type of `connectTo`: `IO (Either SomeException a)`. Exceptions can be thrown in IO anyway...

cardano-api/internal/Cardano/Api/Query.hs
```diff
@@ -356,9 +356,9 @@ data QueryUTxOFilter
   deriving (Eq, Show)
 
 instance NodeToClientVersionOf QueryUTxOFilter where
-  nodeToClientVersionOf QueryUTxOWhole = NodeToClientV_9
```
carbolymer, line 359: Same comment as for `NodeToClientVersionOf (QueryInMode result)`

carbolymer (changes requested):
NTC version change looks problematic to me.

Can you clean up commits history please?

cardano-api/internal/Cardano/Api/Query.hs
crocodile-dentist: These changes are due to the fact that NodeToClientV_15 and earlier have been dropped from o-network.

cardano-api/internal/Cardano/Api/IPC.hs
crocodile-dentist: We distinguish between errors raised by a miniprotocol implementation, which are user defined arbitrary code, and general errors such as handshake errors etc. The former is returned in `Left` while the latter has to be caught manually.

cardano-api/internal/Cardano/Api/Query.hs
carbolymer: Thanks. I'll double check if that's a blocker then and get back to you.

neilmayhew:
> Can you clean up commits history please?

@carbolymer What would you like changed? They're in a logical order and are labelled with the part of the integration that they apply to. However, perhaps you'd like to see some of them squashed. If so, could you let me know which ones?

cardano-api/internal/Cardano/Api/Query.hs
carbolymer: Seems it's not. It's just a complication for the client side: https://ouroboros-consensus.cardano.intersectmbo.org/docs/for-developers/QueryVersioning#checks

cardano-api/cardano-api.cabal
```diff
@@ -406,10 +408,10 @@ test-suite cardano-api-golden
     errors,
     filepath,
     hedgehog >=1.1,
-    hedgehog-extras ^>=0.6.1.0,
+    hedgehog-extras >=0.6.1 && <0.6.5.1,
```
carbolymer, line 411: This is not needed anymore.
```suggestion
    hedgehog-extras >=0.6.1
```

carbolymer (approved):

carbolymer:
@neilmayhew Maybe it's too granular. But I don't have a big problem with it, we can keep it that way.

neilmayhew:
> @neilmayhew Maybe it's too granular. But I don't have a big problem with it, we can keep it that way.

It's a matter of taste, I guess. My own style is to be very fine grained, but I need to be reminded once in a while that this doesn't work for everybody :smile:

carbolymer:
Force merging since garnix was stuck and blocked the merge.

PR #701, opened 2024-12-06
Add 'Convert Experimental.Era MaryEraOnwards' instance

smelc:
# Changelog

```yaml
- description: |
    Add 'Convert Experimental.Era MaryEraOnwards' instance
  type:
```

# Context

Add instance required by CLI (see https://github.com/IntersectMBO/cardano-cli/pull/988)

# How to trust this PR

It's only an instance addition, and a legit one.

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/Experimental/Eras.hs
+++ b/cardano-api/internal/Cardano/Api/Experimental/Eras.hs
@@ -34,6 +34,7 @@ where
 
 import           Cardano.Api.Eon.BabbageEraOnwards
 import           Cardano.Api.Eon.Convert
+import           Cardano.Api.Eon.MaryEraOnwards
 import           Cardano.Api.Eon.ShelleyBasedEra (ShelleyBasedEra (..), ShelleyLedgerEra)
 import qualified Cardano.Api.Eras as Api
 import           Cardano.Api.Eras.Core (BabbageEra, ConwayEra, Eon (..))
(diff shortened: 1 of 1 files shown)


cardano-api/internal/Cardano/Api/Experimental/Eras.hs
```diff
@@ -201,6 +202,11 @@ instance Convert Era ShelleyBasedEra where
     BabbageEra -> ShelleyBasedEraBabbage
     ConwayEra -> ShelleyBasedEraConway
 
+instance Convert Era MaryEraOnwards where
```
Jimbo4350, line 205: This is indicative of a deeper `cardano-api` issue. We should not be converting to the "old api".

smelc: The issue is as follows right now:

Here (within the implementation of `transaction build-estimate`): https://github.com/IntersectMBO/cardano-cli/blob/3c0af7a8088b7c885ddc3c5c03734f2a2d5577cb/cardano-cli/src/Cardano/CLI/EraBased/Run/Transaction.hs#L509

We call API's `estimateBalancedTxBody`: https://github.com/IntersectMBO/cardano-api/blob/76fba1ca226e3cd48e7d42f947da0d9aa1b8f117/cardano-api/internal/Cardano/Api/Fees.hs#L191

And the thing is that `transaction build-estimate` takes an experimental witness, whereas `estimateBalancedTxBody` expects a `MaryEraOnwards` witness.

So I'm not sure how you want to fix that?

Jimbo4350 (approved):

cardano-api/internal/Cardano/Api/Experimental/Eras.hs
carbolymer: Why not just `convert @BabbageEraOnwards . convert` instead of adding a new instance?

Otherwise we will en up with `N!` instances for converting between eons. That's a lot of duplicated boilerplate.

smelc: Indeed thank you @carbolymer, I modified the CLI here accordingly, to avoid the additional instance: https://github.com/IntersectMBO/cardano-cli/pull/1004

smelc:
Closing, not needed anymore, as @carbolymer found out :pray:

PR #710, opened 2024-12-13
Add module for UTxO type and operations

locallycompact:
# Changelog

```yaml
- description: |
    Add `Cardano.Api.Tx.UTxO` module for common UTxO operations.
  type:
```

# Context

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/cardano-api.cabal
+++ b/cardano-api/cardano-api.cabal
@@ -94,6 +94,7 @@ library
     Cardano.Api.Ledger.Lens
     Cardano.Api.Network
     Cardano.Api.Shelley
+    Cardano.Api.Tx.UTxO
 
   build-depends:
     FailT,
@@ -244,6 +245,7 @@ library
     Cardano.Api.Internal.StakePoolMetadata
     Cardano.Api.Internal.Tx.Body
     Cardano.Api.Internal.Tx.Compatible
+    Cardano.Api.Internal.Tx.UTxO
     Cardano.Api.Internal.TxIn
     Cardano.Api.Internal.TxMetadata
     Cardano.Api.Internal.Utils
--- a/cardano-api/src/Cardano/Api.hs
+++ b/cardano-api/src/Cardano/Api.hs
@@ -1142,6 +1142,7 @@ import           Cardano.Api.Internal.SerialiseUsing
 import           Cardano.Api.Internal.StakePoolMetadata
 import           Cardano.Api.Internal.Tx.Body
 import           Cardano.Api.Internal.Tx.Sign
+import           Cardano.Api.Internal.Tx.UTxO
 import           Cardano.Api.Internal.TxMetadata
 import           Cardano.Api.Internal.Utils
 import           Cardano.Api.Internal.Value
--- a/cardano-api/src/Cardano/Api/Internal/Convenience/Construction.hs
+++ b/cardano-api/src/Cardano/Api/Internal/Convenience/Construction.hs
@@ -22,6 +22,7 @@ import           Cardano.Api.Internal.ProtocolParameters
 import           Cardano.Api.Internal.Query
 import           Cardano.Api.Internal.Tx.Body
 import           Cardano.Api.Internal.Tx.Sign
+import           Cardano.Api.Internal.Tx.UTxO (UTxO (..))
 import           Cardano.Api.Internal.Utils
 
 import qualified Cardano.Ledger.Api as L
--- a/cardano-api/src/Cardano/Api/Internal/Convenience/Query.hs
+++ b/cardano-api/src/Cardano/Api/Internal/Convenience/Query.hs
@@ -35,6 +35,7 @@ import           Cardano.Api.Internal.ProtocolParameters
 import           Cardano.Api.Internal.Query
 import           Cardano.Api.Internal.Query.Expr
 import           Cardano.Api.Internal.Tx.Body
+import           Cardano.Api.Internal.Tx.UTxO
 import           Cardano.Api.Internal.Utils
 
 import qualified Cardano.Ledger.Api as L
--- a/cardano-api/src/Cardano/Api/Internal/Fees.hs
+++ b/cardano-api/src/Cardano/Api/Internal/Fees.hs
@@ -67,6 +67,7 @@ import           Cardano.Api.Internal.Query
 import           Cardano.Api.Internal.Script
 import           Cardano.Api.Internal.Tx.Body
 import           Cardano.Api.Internal.Tx.Sign
+import           Cardano.Api.Internal.Tx.UTxO
 import           Cardano.Api.Internal.Value
 import qualified Cardano.Api.Ledger.Lens as A
 
--- a/cardano-api/src/Cardano/Api/Internal/Plutus.hs
+++ b/cardano-api/src/Cardano/Api/Internal/Plutus.hs
@@ -14,12 +14,13 @@ import           Cardano.Api.Internal.Eon.AlonzoEraOnwards (AlonzoEraOnwards (..
 import           Cardano.Api.Internal.Eon.Convert (convert)
 import           Cardano.Api.Internal.Eon.ShelleyBasedEra (ShelleyLedgerEra)
 import           Cardano.Api.Internal.Pretty (Pretty (pretty), docToText)
-import           Cardano.Api.Internal.Query (UTxO, toLedgerUTxO)
+import           Cardano.Api.Internal.Query (toLedgerUTxO)
 import qualified Cardano.Api.Internal.ReexposeLedger as L
 import           Cardano.Api.Internal.Script (ScriptHash, fromShelleyScriptHash)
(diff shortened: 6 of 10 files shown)


cardano-api/internal/Cardano/Api/Query/UTxO.hs
```diff
@@ -0,0 +1,88 @@
...
+   where
+    toTxIn :: (Text, Aeson.Value) -> Parser (TxIn, TxOut CtxUTxO era)
+    toTxIn (txinText, txOutVal) = do
+      (,)
+        <$> parseJSON (Aeson.String txinText)
+        <*> parseJSON txOutVal
+
+-- | Create a 'UTxO from a list of 'TxIn' and 'out' pairs.
+fromPairs :: [(TxIn, TxOut CtxUTxO era)] -> UTxO era
+fromPairs = UTxO . Map.fromList
```
carbolymer, line 52: An instance of `IsList` can be provided to avoid the need of having a specialised function.

cardano-api/internal/Cardano/Api/Query/UTxO.hs
```diff
@@ -0,0 +1,88 @@
...
+import qualified Data.Aeson.KeyMap as KeyMap
+import           Data.Aeson.Types (Parser)
+import qualified Data.HashMap.Strict as HashMap
+import qualified Data.List as List
+import           Data.Map (Map)
+import qualified Data.Map as Map
+import           Data.Set (Set)
+import           Data.Text (Text)
+
+newtype UTxO era = UTxO {unUTxO :: Map TxIn (TxOut CtxUTxO era)}
```
carbolymer, line 27: An instance of `MonoTraversable (UTxO era)` would be useful here.

cardano-api/internal/Cardano/Api/Query/UTxO.hs
```diff
@@ -0,0 +1,88 @@
...
+singleton :: (TxIn, TxOut CtxUTxO era) -> UTxO era
+singleton (i, o) = UTxO $ Map.singleton i o
+
+-- | Find an 'out' for a given 'TxIn'.
+resolve :: TxIn -> UTxO era -> Maybe (TxOut CtxUTxO era)
+resolve k = Map.lookup k . unUTxO
+
+-- | Turn a 'UTxO into a list of pairs.
+pairs :: UTxO era -> [(TxIn, TxOut CtxUTxO era)]
+pairs = Map.toList . unUTxO
```
carbolymer, line 64: Redundant, if `IsList (UTxO era)` is provided

carbolymer (commented):
LGTM, @Jimbo4350 thoughts?

locallycompact:
8.10 doesn't support IsList so this is a bit awkward.

cardano-api/internal/Cardano/Api/Query/UTxO.hs
```diff
@@ -0,0 +1,90 @@
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE UndecidableInstances #-}
```
carbolymer, line 6: Why `UndecidableInstances` is needed?

locallycompact: Apparently

```
internal/Cardano/Api/Query/UTxO.hs:29:20: error: [GHC-22979]
    • The type-family application ‘Item (Map TxIn (TxOut CtxUTxO era))’
        is no smaller than the LHS of the family instance ‘Item (UTxO era)’
    • In the instance declaration for ‘IsList (UTxO era)’
    Suggested fix: Perhaps you intended to use UndecidableInstances
   |
29 |   deriving newtype IsList
   |                    ^^^^^^
```

locallycompact: @carbolymer Can we resolve this? Is something weird here?

carbolymer: Sorry, I've missed GH notification about your message. Let's leave it.

cardano-api/internal/Cardano/Api/Query/UTxO.hs
```diff
@@ -0,0 +1,84 @@
...
+-- | Create an empty `UTxO`.
+empty :: UTxO era
+empty = UTxO Map.empty
+
+-- | Create a `UTxO` from a single unspent transaction output.
+singleton :: TxIn -> TxOut CtxUTxO era -> UTxO era
+singleton i o = UTxO $ Map.singleton i o
+
+-- | Same as `lookup`.
+resolve :: TxIn -> UTxO era -> Maybe (TxOut CtxUTxO era)
```
Jimbo4350, line 63: why `resolve` as well?

cardano-api/cardano-api.cabal
```diff
@@ -248,6 +250,7 @@ library
     Cardano.Api.Ledger
     Cardano.Api.Network
     Cardano.Api.Shelley
+    Cardano.Api.UTxO
```
Jimbo4350, line 253: Keep this module hidden and expose the functionality from `Cardano.Api.` We need to do an audit on how we expose things.

locallycompact: No. This module by design is intended to be imported qualified, similar to container modules such as `Data.Map`, as it contains function names which commonly conflict.

cardano-api/src/Cardano/Api/UTxO.hs
```diff
@@ -0,0 +1,3 @@
+module Cardano.Api.UTxO (module Cardano.Api.Query.UTxO) where
```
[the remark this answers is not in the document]
locallycompact: Also see above.

cardano-api/internal/Cardano/Api/Query/UTxO.hs
locallycompact: Yes this could be dropped.

cardano-api/cardano-api.cabal
Jimbo4350: Yes. This PR will remain unmerged until this change is made

cardano-api/src/Cardano/Api/UTxO.hs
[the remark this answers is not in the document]
Jimbo4350: Likewise, this PR will remain unmerged until this change is made.

Jimbo4350 (changes requested):
Two changes need to be made

cardano-api/cardano-api.cabal
locallycompact: OK. Didn't mean to sound presumptuous, I assumed we were misunderstanding each other. If I do understand you correctly, do you mean to expose functions such as `empty` and `lookup` directly through `Cardano.Api`? What I was trying to do here is provide a qualified import that deals with only this container type, because that is clear to the reader, eg `import Cardano.Api.UTxO as UTxO`, and called as `UTxO.empty`, `UTxO.lookup`. Doing `import qualified Cardano.Api as UTxO` is confusing, as it would contain functions that don't concern UTxOs, and `import qualified Cardano.Api as Api` would also be confusing, as `Api.empty` requires context for the reader when reading at the call site. Can you confirm if this is what you meant?

carbolymer: I agree here with @locallycompact . Exposing commonly used function names like `empty` or `singleton` through `Cardano.Api` will be annoying. End users will be forced to qualify `Cardano.Api` imports. Moreover, if we decide to provide another similar container utility the same way in cardano-api, then the names will be in use already, and we'll be forced to use different names e.g. `emptyTxSet`, `lookupTxSet` ...

Jimbo4350: @locallycompact Fair point.

@carbolymer  I'd like to task you with writing an ADR as to how we should expose `cardano-api`'s functionality. As it has been sub optimal for a long time.

Jimbo4350 (approved):

cardano-api/cardano-api.cabal
carbolymer: Or we could pick a middle ground here: export just the `UTxO` type and its instances through `Cardano.Api`, the utility functions should stay in the `UTxO` module exported from `cardano-api`.

locallycompact: @carbolymer I think that's the current behaviour

carbolymer:
Can you place the new UTxO module in `Cardano.Api.Tx.UTxO`? The cabal package is fine. We're shifting modules a bit in https://github.com/IntersectMBO/cardano-api/pull/748/ so it'll better suit there I think.

locallycompact:
I will rebase after that work.

carbolymer (approved):

PR #719, opened 2025-01-09
UTxO-HD targeting main

jasagredo:
# Changelog

```yaml
- description: |
    <insert-changelog-description-here>
  type:
```

# Context

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff


Jimbo4350 (approved):

cardano-api/internal/Cardano/Api/LedgerState.hs
```diff
@@ -2017,12 +2178,42 @@ data AnyNewEpochState where
...
     shelleyBasedEraConstraints sbe $ showsPrec p ledgerNewEpochState
 
+getUTxOValues
+  :: forall era
+   . ShelleyBasedEra era
+  -> Ledger.LedgerTables
+      (Ledger.LedgerState (Consensus.CardanoBlock Consensus.StandardCrypto))
+      Ledger.ValuesMK
+  -> Map TxIn (TxOut CtxUTxO era)
+getUTxOValues sbe tbs =
```
Jimbo4350, line 2197: Is this only retrieving UTxOs in memory? I.e there could be UTxOs on disk that will not be returned by this function?

cardano-api/internal/Cardano/Api/LedgerState.hs
```diff
@@ -2200,7 +2391,7 @@ foldEpochState nodeConfigFilePath socketPath validationMode terminationEpoch ini
                         let !err = Just e
                          in clientIdle_DoneNwithMaybeError n err
                       Right lState -> do
-                        let newEpochState = AnyNewEpochState sbe lState
+                        let newEpochState = AnyNewEpochState sbe lState (clsTables newLedgerState)
```
Jimbo4350, line 2394: This will only return UTxOs in memory correct? I.e there may be UTxOs on disk that will not be returned in `newEpochState`.

jasagredo: `Cardano.Api.LedgerState.LedgerState` is a purely in-memory ledger state. There are no UTxOs on the disk. If we were to have UTxOs on disk, we would have to have a LedgerDB that takes care of managing that disk storage. `foldEpochState` carries a single LedgerState thus the decision of making it in-memory.

.github/workflows/haskell.yml
```diff
@@ -80,6 +80,19 @@ jobs:
...
+
+    - name: Mac install lmdb
+      if: matrix.sys.os == 'macos-latest'
+      run: brew install lmdb
+
+    - name: Windows install lmdb
+      if: matrix.sys.os == 'windows-latest'
+      shell: 'C:/msys64/usr/bin/bash.exe -e {0}'
+      run: /usr/bin/pacman --noconfirm -S mingw-w64-x86_64-lmdb
+
```
carbolymer, line 95: I think this should go into `input-output-hk/actions/base@latest` to avoid having extra steps in multiple repos.

carbolymer (approved):

PR #729, opened 2025-01-16
Remove ProtocolParameters

smelc:
# Changelog

```yaml
- description: |
    Remove the ProtocolParameters type, that has been deprecated for a while
# uncomment types applicable to the change:
  type:
  # - feature        # introduces a new feature
  - breaking       # the API has changed in a breaking way
  # - compatible     # the API has changed but is non-breaking
  # - optimisation   # measurable performance improvements
  # - refactoring    # QoL changes
  # - bugfix         # fixes a defect
  # - test           # fixes/modifies tests
  # - maintenance    # not directly related to the code
  # - release        # related to a new release preparation
  # - documentation  # change in code docs, haddocks...
```

# Context

* `ProtocolParameters` had been deprecated in favor the ledger's `PParams` type. We are now removing `ProtocolParameters` altogether.
* Fixes https://github.com/IntersectMBO/cardano-api/issues/384
* To be merged after https://github.com/IntersectMBO/cardano-node/pull/6131 is merged

# How to trust this PR

* It's deletion only
* Execute `git grep ' ProtocolParameters '` in `cardano-cli` and observe that there are no hits
* Dependency on API's `ProtocolParameters` has been removed from `cardano-node` in https://github.com/IntersectMBO/cardano-node/pull/6131

# Checklist

- [X] Commit sequence broadly makes sense and commits have useful messages
- [X] Self-reviewed the diff

PR #731, opened 2025-01-20
Shelley: export {to,from}AlonzoLanguage (Plutus script language conversions)

smelc:
# Changelog

```yaml
- description: |
    Shelley: export {to,from}AlonzoLanguage (Plutus script language conversions)
  type:
```

# Context

Required to adapt `cardano-node` to https://github.com/IntersectMBO/cardano-api/pull/729

# Checklist

- [X] Self-reviewed the diff


cardano-api/src/Cardano/Api/Shelley.hs
```diff
@@ -137,6 +137,8 @@ module Cardano.Api.Shelley
   , PlutusScript (..)
   , PlutusScriptOrReferenceInput (..)
   , SimpleScriptOrReferenceInput (..)
+  , toAlonzoLanguage
+  , fromAlonzoLanguage
```
carbolymer, line 141: I think we should replace those two function with `Inject` instances.

carbolymer (approved):
LGTM

cardano-api/src/Cardano/Api/Shelley.hs
smelc: @carbolymer> I disagree, I think we should keep `Inject` for conversions between eons; otherwise upon seeing a call to `inject` in the code we won't easily know what it's about.

carbolymer: That's not what is `Inject` for in general. It's used widely in ledger codebase for conversion between isomorphic types.

smelc: @carbolymer> is the ledger a model codebase in terms of ease-of-understanding :wink:?

PR #456, opened 2024-02-15
Implemenet and expose getScriptCredential

Jimbo4350:
# Changelog

```yaml
- description: |
   Implemenet and expose getScriptCredential
# uncomment types applicable to the change:
  type:
  - feature        # introduces a new feature
  - compatible     # the API has changed but is non-breaking

```

# Context

Additional context for the PR goes here. If the PR fixes a particular issue please provide a [link](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword=) to the issue.

# How to trust this PR

Highlight important bits of the PR that will make the review faster. If there are commands the reviewer can run to observe the new behavior, describe them.

# Checklist

- [ ] Commit sequence broadly makes sense and commits have useful messages
- [ ] New tests are added if needed and existing tests are updated. See [Running tests](https://github.com/input-output-hk/cardano-node-wiki/wiki/Running-tests) for more details
- [ ] Self-reviewed the diff

PR #734, opened 2025-01-22
Change a representation of witnesses in transaction's certificates to an ordered map where a certificate is the key

carbolymer:
# Changelog

```yaml
- description: |
    Change a representation of witnesses in transaction's certificates to an ordered map where a certificate is the key.
  type:
```

# Context

This PR changes representation of transaction certificates' witnesses from a list of with stake credentials to a map.

The issue is that there's a relationship `Certificate *--1 StakeCredential` which makes the assigning of witnesses impossible in the cases where we have more than one certificate with the same stake credential. This then breaks witnesses indexing.

See related PRs:
- https://github.com/IntersectMBO/cardano-cli/pull/1028
- https://github.com/IntersectMBO/cardano-node/pull/6082/

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/internal/Cardano/Api/Certificate.hs
+++ b/cardano-api/internal/Cardano/Api/Certificate.hs
@@ -142,6 +142,8 @@ data Certificate era where
 
 deriving instance Eq (Certificate era)
 
+deriving instance Ord (Certificate era)
+
 deriving instance Show (Certificate era)
 
 instance Typeable era => HasTypeProxy (Certificate era) where
--- a/cardano-api/internal/Cardano/Api/Fees.hs
+++ b/cardano-api/internal/Cardano/Api/Fees.hs
@@ -245,10 +245,7 @@ estimateBalancedTxBody
     --  1. Subtract certificate and proposal deposits
     -- from the total available Ada value!
     -- Page 24 Shelley ledger spec
-    let certificates =
-          case txCertificates txbodycontent1 of
-            TxCertificatesNone -> []
-            TxCertificates _ certs _ -> map toShelleyCertificate certs
+    let certificates = convCertificates sbe $ txCertificates txbodycontent1
 
         proposalProcedures :: OSet.OSet (L.ProposalProcedure (ShelleyLedgerEra era))
         proposalProcedures =
@@ -478,8 +475,9 @@ estimateTransactionKeyWitnessCount
             length [() | (_, _, BuildTxWith KeyWitness{}) <- withdrawals]
           _ -> 0
         + case txCertificates of
-          TxCertificates _ _ (BuildTxWith witnesses) ->
-            length [() | (_, KeyWitness{}) <- witnesses]
+          TxCertificates _ credWits ->
+            length
+              [() | (_, BuildTxWith (Just (_, KeyWitness{}))) <- toList credWits]
           _ -> 0
         + case txUpdateProposal of
           TxUpdateProposal _ (UpdateProposal updatePerGenesisKey _) ->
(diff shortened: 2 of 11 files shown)


cardano-api/internal/Cardano/Api/Fees.hs
```diff
@@ -1500,18 +1498,27 @@ substituteExecutionUnits
...
     mapScriptWitnessesCertificates TxCertificatesNone = Right TxCertificatesNone
-    mapScriptWitnessesCertificates txCertificates'@(TxCertificates supported certs _) =
+    mapScriptWitnessesCertificates txCertificates'@(TxCertificates supported _) = do
       let mappedScriptWitnesses
-            :: [(StakeCredential, Either (TxBodyErrorAutoBalance era) (Witness WitCtxStake era))]
+            :: [ ( Certificate era
+                 , Either
+                    (TxBodyErrorAutoBalance era)
+                    ( BuildTxWith
+                        BuildTx
```
Jimbo4350, line 1507: This makes my eyes bleed (I know it's not your fault).

cardano-api/internal/Cardano/Api/Tx/Body.hs
```diff
@@ -1280,38 +1283,56 @@ indexTxWithdrawals (TxWithdrawals _ withdrawals) =
...
   TxCertificates
     :: ShelleyBasedEra era
-    -> [Certificate era]
-    -> BuildTxWith build [(StakeCredential, Witness WitCtxStake era)]
-    -- ^ There can be more than one script witness per stake credential
+    -> OMap
+        (Certificate era)
+        ( BuildTxWith
+            build
+            (Maybe (StakeCredential, Witness WitCtxStake era))
```
Jimbo4350, line 1297: I would seek confirmation that  https://github.com/IntersectMBO/cardano-api/pull/456 is not an issue with this change. I suspect this needs to be a list still.

cardano-api/internal/Cardano/Api/Certificate.hs
```diff
@@ -142,6 +142,8 @@ data Certificate era where
 
 deriving instance Eq (Certificate era)
 
+deriving instance Ord (Certificate era)
```
Jimbo4350, line 145: I would add a `ord_distributive` test for `Certificate` just to be safe.

Jimbo4350 (changes requested):
Confirm we aren't reintroducing a bug and I'll be happy to approve.

cardano-api/internal/Cardano/Api/Tx/Body.hs
carbolymer: I've ran regression tests here: https://github.com/IntersectMBO/cardano-node-tests/actions/runs/12928366127
`cardano-testnet` tests pass: https://github.com/IntersectMBO/cardano-node/pull/6082

Is that enough?

carbolymer: >I suspect this needs to be a list still.
I don't see a reason for that.

cardano-api/internal/Cardano/Api/Certificate.hs
carbolymer: Added

cardano-api/internal/Cardano/Api/Tx/Body.hs
```diff
@@ -1280,38 +1283,56 @@ indexTxWithdrawals (TxWithdrawals _ withdrawals) =
...
-  [ (ScriptWitnessIndexCertificate ix, cert, stakeCred, wit)
-  | (ix, cert) <- zip [0 ..] certs
-  , stakeCred <- maybeToList (selectStakeCredentialWitness cert)
-  , wit <- findAll stakeCred witnesses
+indexTxCertificates (TxCertificates _ certsWits) =
+  [ (ScriptWitnessIndexCertificate ix, cert, stakeCred, witness)
+  | (ix, (cert, BuildTxWith (Just (stakeCred, witness)))) <- zip [0 ..] $ toList certsWits
   ]
- where
-  findAll needle = map snd . filter ((==) needle . fst)
```
Jimbo4350, line 1310: So this is my concern here. Previously you can have multiple stake credentials witnessing a single certificate. Now we have limited that to one. @mkoura do your tests cover this case?

carbolymer: >you can have multiple stake credentials witnessing a single certificate
How is that possible? There's a relationship `Certificate *-1 StakeCredential`, not the other way around.

Jimbo4350: >  Previously you can have multiple stake credentials witnessing a single certificate. 

Poor wording. A script stake credential can require multiple witnesses IIRC.

carbolymer: See PR description:
>The issue is that there's a relationship Certificate *--1 StakeCredential which makes the assigning of witnesses impossible in the cases where we have more than one certificate with the same stake credential. This then breaks witnesses indexing.

You can still have multiple witnesses per stake credential, but only one witness per certificate.

cardano-api/internal/Cardano/Api/Tx/Body.hs
```diff
@@ -1280,38 +1283,56 @@ indexTxWithdrawals (TxWithdrawals _ withdrawals) =
...
+ where
+  getStakeCred (cert, mWit) = do
+    let wit =
+          maybe
+            (KeyWitness KeyWitnessForStakeAddr)
+            (ScriptWitness ScriptWitnessForStakeAddr)
+            mWit
+    ( cert
+      , pure $
+          (,wit) <$> selectStakeCredentialWitness cert
```
Jimbo4350, line 1323: To be aware of.

In the ledger repo
```

-- For both of the functions `getScriptWitnessConwayTxCert` and
-- `getVKeyWitnessConwayTxCert` we preserve the old behavior of not requiring a witness
-- for staking credential registration, but only during the transitional period of Conway
-- era and only for staking credential registration certificates without a deposit. Future
-- eras will require a witness for registration certificates, because the one without a
-- deposit will be removed.
```

Eventually we should be able to remove the `Maybe` in:
```
OMap
        (Certificate era)
        ( BuildTxWith
            build
            (Maybe (StakeCredential, Witness WitCtxStake era))
```

Can you add a comment to the type definition pointing this out?

Jimbo4350 (approved):
Nice work. Takes a little while to page the context back into my head. One minor comment.

cardano-api/internal/Cardano/Api/Tx/Body.hs
carbolymer: Added

PR #746, opened 2025-02-04
Remove deprecated functions and update types and update `serialiseTxLedgerCddl` to use new format

palas:
# Changelog

```yaml
- description: |
    Removed `serialiseTxLedgerCddl` and `deserialiseTxLedgerCddl`, and updated `writeTxFileTextEnvelopeCddl ` to use new format.
  type:
  - breaking
```

# Context

`serialiseTxLedgerCddl` and `deserialiseTxLedgerCddl` have been deprecated for two months, so I am removing them. Just realised `writeTxFileTextEnvelopeCddl` is still using the old format, so I updated that and I will remove the compatibility layer in two months.

# How to trust this PR

Check that what I say in `Context` is true. Also that I haven't removed any still relevant tests.

# Checklist

- [x] Self-reviewed the diff


cardano-api/test/cardano-api-test/Test/Cardano/Api/CBOR.hs
```diff
@@ -41,28 +40,6 @@ import           Test.Tasty.Hedgehog (testProperty)
...
--- compatible with both during the transition.
-prop_forward_compatibility_txbody_CBOR :: Property
-prop_forward_compatibility_txbody_CBOR = H.property $ do
-  AnyShelleyBasedEra era <- H.noteShowM . H.forAll $ Gen.element [minBound .. maxBound]
-  x <- H.forAll $ makeSignedTransaction [] . fst <$> genValidTxBody era
-  shelleyBasedEraConstraints
-    era
-    ( H.tripping
-        x
-        (serialiseToTextEnvelope (Just (TextEnvelopeDescr "Ledger Cddl Format")))
```
Jimbo4350, line 56: These should be replaced with roundtrip tests that use `serialiseToTextEnvelope`  and `deserialiseFromTextEnvelope`. We still care about having coherent instances!

palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/746/commits/fb5de89bc18954e2792acac2355bc815f87e67d8

Jimbo4350 (approved):
So the overall aim is to remove the module `Cardano.Api.Internal.SerialiseLedgerCddl`  in its entirety and have functions like `legacyComparison` used in `Cardano.Api.Internal.SerialiseTextEnvelope` to maintain backwards compatibility wrt deserialization. Can you create an issue for this? We should also require decode only tests upon the final removal.

palas:
Backport: https://github.com/IntersectMBO/cardano-api/tree/remove-deprecated-functions-backport

PR #750, opened 2025-02-11
Split compatible transaction building and signing

carbolymer:
# Changelog

```yaml
- description: |
    Split compatible transaction building into separate building and signing functions.
    Rename `Cardano.Api.Internal.Tx.Compatible` to `Cardano.Api.Internal.Compatible.Tx`.
  type:
```

# Context

`createCompatibleSignedTx` wasn't really signing the transaction but just blindly adding signatures. It's split into two separate functions for building and signing.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/src/Cardano/Api/Internal/Tx/Sign.hs
+++ b/cardano-api/src/Cardano/Api/Internal/Tx/Sign.hs
@@ -2,7 +2,6 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GADTs #-}
-{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE PatternSynonyms #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
@@ -13,8 +12,6 @@
 -- The Shelley ledger uses promoted data kinds which we have to use, but we do
 -- not export any from this API. We also use them unticked as nature intended.
 {-# OPTIONS_GHC -Wno-unticked-promoted-constructors #-}
-
-{- HLINT ignore "Avoid lambda using `infix`" -}
 
 -- | Complete, signed transactions
 module Cardano.Api.Internal.Tx.Sign
@@ -43,8 +40,10 @@ module Cardano.Api.Internal.Tx.Sign
   , makeByronKeyWitness
   , ShelleyWitnessSigningKey (..)
   , makeShelleyKeyWitness
+  , makeShelleyKeyWitness'
   , WitnessNetworkIdOrByronAddress (..)
   , makeShelleyBootstrapWitness
+  , makeShelleyBasedBootstrapWitness
   , makeShelleySignature
   , getShelleyKeyWitnessVerificationKey
   , getTxBodyAndWitnesses
@@ -126,6 +125,12 @@ data Tx era where
     :: ShelleyBasedEra era
     -> L.Tx (ShelleyLedgerEra era)
     -> Tx era
+
+-- | This pattern will be deprecated in the future. We advise against introducing new usage of it.
+pattern Tx :: TxBody era -> [KeyWitness era] -> Tx era
+pattern Tx txbody ws <- (getTxBodyAndWitnesses -> (txbody, ws))
+  where
+    Tx txbody ws = makeSignedTransaction ws txbody
 
 instance Show (InAnyCardanoEra Tx) where
   show (InAnyCardanoEra _ tx) = show tx
@@ -749,12 +754,6 @@ instance IsCardanoEra era => HasTextEnvelope (KeyWitness era) where
 getTxBodyAndWitnesses :: Tx era -> (TxBody era, [KeyWitness era])
 getTxBodyAndWitnesses tx = (getTxBody tx, getTxWitnesses tx)
 
--- | This pattern will be deprecated in the future. We advise against introducing new usage of it.
-pattern Tx :: TxBody era -> [KeyWitness era] -> Tx era
-pattern Tx txbody ws <- (getTxBodyAndWitnesses -> (txbody, ws))
-  where
-    Tx txbody ws = makeSignedTransaction ws txbody
-
 {-# COMPLETE Tx #-}
 
 data ShelleyWitnessSigningKey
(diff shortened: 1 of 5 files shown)


cardano-api/src/Cardano/Api/Internal/Tx/Compatible.hs
```diff
@@ -224,3 +210,77 @@ createCommonTxBody era ins outs txFee' =
             .~ Seq.fromList txOuts'
           & L.feeTxBodyL
             .~ txFee'
+
+-- | Sign a transaction body
+makeSignedTransaction
```
carbolymer, line 215: I'ts a 1-1 copy from `Cardano.Api.Internal.Tx.Sign`. I imagine that we'll be deprecating eras and simplifying `Cardano.Api.Internal.Tx.Sign` so we won't be able to reuse code easily.

cardano-api/src/Cardano/Api/Internal/Tx/Compatible.hs
```diff
@@ -224,3 +210,77 @@ createCommonTxBody era ins outs txFee' =
...
+  :: forall era
+   . [KeyWitness era]
+  -> TxBody era
+  -> Tx era
+  -- ^ a signed transaction
+makeSignedTransaction
+  witnesses
+  ( ShelleyTxBody
```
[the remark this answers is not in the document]
carbolymer: Ok I'll try to rewrite this without using our `Tx` type

cardano-api/src/Cardano/Api/Internal/Tx/Compatible.hs
carbolymer: I meant deprecating `Eon`s. So `Cardano.Api.Internal.Tx.Sign` will only support current and upcoming eras, correct?

Jimbo4350: `Cardano.Api.Internal.Compatible` has to support all Shelley based eras only for the functionality QA needs.

Once we can provide what QA needs, we can consider parameterizing functions in `Cardano.Api.Internal.Compatible` on `CardanoEra era` however the Byron/Shelley split in the code makes things cleaner.

cardano-api/src/Cardano/Api/Internal/Tx/Sign.hs
```diff
@@ -1106,19 +1106,27 @@ makeShelleyKeyWitness
...
-          \wsk ->
-            let sk = toShelleySigningKey wsk
-                vk = getShelleyKeyWitnessVerificationKey sk
-                signature = makeShelleySignature txhash sk
-             in ShelleyKeyWitness sbe $
-                  L.WitVKey vk signature
+makeShelleyKeyWitness sbe (ShelleyTxBody _ txBody _ _ _ _) =
+  makeShelleyKeyWitness' sbe (A.TxBody txBody)
+
+makeShelleyKeyWitness'
```
carbolymer, line 1112: I'm not sure what's the best approach here to naming and the argument for the version of this function working on ledger's tx body, so I went with this.

cardano-api/src/Cardano/Api/Internal/Tx/Compatible.hs
```diff
@@ -164,11 +165,11 @@ createCompatibleSignedTx sbe ins outs witnesses txFee' anyProtocolUpdate anyVote
     :: [(ScriptWitnessIndex, Certificate era, StakeCredential, Witness WitCtxStake era)]
   indexedTxCerts = indexTxCertificates txCertificates'
 
-  allWitnesses
+  setScriptWitnesses
```
Jimbo4350, line 168: This just occurred to me. Does QA need backwards compatibility for script witnesses? If not we can drop it.

cardano-api/src/Cardano/Api/Internal/Tx/Compatible.hs
```diff
@@ -224,3 +210,31 @@ createCommonTxBody era ins outs txFee' =
             .~ Seq.fromList txOuts'
           & L.feeTxBodyL
             .~ txFee'
+
+-- | Add provided witnesses to the transaction
+addWitnesses
```
Jimbo4350, line 215: I would suggest changing the directory name space from: `Cardano.Api.Internal.Tx.Compatible` to `Cardano.Api.Internal.Compatible.Tx`. Then you can add `Cardano.Api.Internal.Compatible.Witness`.

cardano-api/src/Cardano/Api/Internal/Tx/Sign.hs
Jimbo4350: I would suggest changing the directory name space from: `Cardano.Api.Internal.Tx.Compatible` to `Cardano.Api.Internal.Compatible.Tx`. Then you can create `Cardano.Api.Internal.Compatible.Witness` and add the witness related functions there.

cardano-api/src/Cardano/Api/Ledger/Lens.hs
```diff
@@ -83,6 +84,12 @@ strictMaybeL = lens g s
   s :: StrictMaybe a -> Maybe a -> StrictMaybe a
   s _ = maybe SNothing SJust
 
+txToTxBodyL :: ShelleyBasedEra era -> Lens' (L.Tx (ShelleyLedgerEra era)) (TxBody era)
```
Jimbo4350, line 87: Is this called in the cli? If it's not lets not expose this because it encourages the use of `TxBody`.

Jimbo4350 (approved):
LGTM :+1:

cardano-api/src/Cardano/Api/Ledger/Lens.hs
carbolymer: It's used here: https://github.com/IntersectMBO/cardano-cli/pull/1057/files#diff-725abf998f7812014ea390b5cde9b1fe273d7af84e585bc47f37f9052399e7f9R286

Note that it's a `TxBody` from line 74, which is just a newtype over `L.TxBody (ShelleyLedgerEra era)`

cardano-api/src/Cardano/Api/Internal/Tx/Sign.hs
carbolymer: There's only one witness function at the moment, I don't anticipate adding more. I can rename the module, but I don't think we need to split it yet.

cardano-api/src/Cardano/Api/Internal/Tx/Compatible.hs
carbolymer: I think so. This was requested afair: https://github.com/IntersectMBO/cardano-cli/pull/972/files#diff-caae695482c364ec0098429ee19afc0089a97fe85de02cf1f81d3b029ef18109R41

PR #768, opened 2025-03-03
Make it build with ghc 9.12

erikd:
# Changelog

```yaml
- description: Make it build with ghc 9.12
  type:
```

# Context

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc: ["8.10.7", "9.6", "9.10"]
+        ghc: ["8.10.7", "9.6", "9.12"]
         cabal: ["3.14"]
         sys:
           - { os: windows-latest, shell: 'C:/msys64/usr/bin/bash.exe -e {0}' }
--- a/flake.nix
+++ b/flake.nix
@@ -3,17 +3,16 @@
 
   inputs = {
     hackageNix = {
-      # The "for-stakage" part is a workaround haskell.nix and hackage.nix
-      # being out of sync until we drop support for GHC 8.10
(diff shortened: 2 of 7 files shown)


carbolymer:
Nice. There are nix errors here though. Maybe `haskellNix` needs to be bumped for that as well? We have it pinned in the input: https://github.com/IntersectMBO/cardano-api/blob/master/flake.nix#L10

flake.nix
```diff
@@ -99,7 +96,9 @@
               ghcid = "0.8.9";
               cabal-gild = "1.3.1.2";
               fourmolu = "0.18.0.0";
-              haskell-language-server.src = nixpkgs.haskell-nix.sources."hls-2.9";
+              # HLS is corrently broken for GHC 9.8
+              # See https://github.com/haskell/haskell-language-server/issues/4493
+              # haskell-language-server.src = nixpkgs.haskell-nix.sources."hls-2.9";
```
carbolymer, line 101: Would it be possible to build HLS from master, instead?

.github/workflows/haskell.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc: ["8.10.7", "9.6", "9.10"]
+        ghc: ["8.10.7", "9.6", "9.10", "9.12"]
```
Jimbo4350, line 23: Do we need to be building with all of these? We need 8.10.7 for sure but do we need 9.6 and 9.10?

erikd: Dropped `9.10`.

Jimbo4350 (approved):

carbolymer (approved):

PR #771, opened 2025-03-05
Add `HasTextEnvelope` instance for `EraHistory`

palas:
# Changelog

```yaml
- description: |
    Added `HasTextEnvelope` instance for `EraHistory`
  type:
```

# Context

This change in the API is required to support the related PR in `cardano-cli`: https://github.com/IntersectMBO/cardano-cli/pull/1079

# How to trust this PR

I would look at all the changes, which are few. The fact that it type-checks already gives a lot of guarantee. The approach is pretty similar to that of other `HasTextEnvelope` instances. The tests in the related `cardano-cli` PR can offer further assurance.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-api/cardano-api.cabal
+++ b/cardano-api/cardano-api.cabal
@@ -165,6 +165,7 @@ library
     prettyprinter-ansi-terminal,
     prettyprinter-configurable ^>=1.36,
     random,
+    reflection,
     safe-exceptions,
     scientific,
     serialise,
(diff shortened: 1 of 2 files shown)


cardano-api/cardano-api.cabal
```diff
@@ -165,6 +165,7 @@ library
     prettyprinter-ansi-terminal,
     prettyprinter-configurable ^>=1.36,
     random,
+    reflection,
```
carbolymer, line 168: seems we have in the dependency tree already, so it shouldn't influence build times

carbolymer (approved):

Jimbo4350 (commented):
LGTM however where is this being used?

Issue #769, opened 2025-03-04
[FR] - TxMintValue shouldn't allow multiple script witnesses per PolicyId?


eddiemundo:
Seems like the `TxMintValue` constructor was changed from
```
MaryEraOnwards era -> Value -> BuildTxWith ctx (Map PolicyId (ScriptWitness WitCtxMint era))
```
to
```
MaryEraOnwards era -> Map PolicyId [(AssetName, Quantity, BuildTxWith build (ScriptWitness WitCtxMint era))]
```
which is a bit confusing because it suggests each PolicyId can have multiple script witness when AFAIK they can't. I assume that in the case multiple scripts are used for a PolicyId only the first or last one is used...

carbolymer:
Fair point, thanks for reporting that!

PR #663, opened 2024-10-25
Refactor `TxMintValue`

carbolymer:
# Changelog

```yaml
- description: |
    Add `sbeToSimpleScriptLanguageInEra`, `getScriptWitnessScript`, `getScriptWitnessReferenceInput`, `getScriptWitnessReferenceInputOrScript` function
    Refactor `TxMintValue` to better represent minting state
    Propagate `IsPlutusLanguage` constraint to `ScriptLanguage lang`, `AnyPlutusScriptVersion`, `Script lang` and `ScriptWitness witctx era` types.
    Remove `Maybe ScriptHash` from `PReferenceScript` and `SReferenceScript`.

# uncomment types applicable to the change:
  type:
  # - feature        # introduces a new feature
   - breaking       # the API has changed in a breaking way
  # - compatible     # the API has changed but is non-breaking
  # - optimisation   # measurable performance improvements
   - refactoring    # QoL changes
  # - bugfix         # fixes a defect
  # - test           # fixes/modifies tests
  # - maintenance    # not directly related to the code
  # - release        # related to a new release preparation
  # - documentation  # change in code docs, haddocks...
```

# Context

Supersedes:
- https://github.com/IntersectMBO/cardano-api/pull/642

Fixes:
- https://github.com/IntersectMBO/cardano-api/issues/606

# Checklist

- [ ] Commit sequence broadly makes sense and commits have useful messages
- [ ] New tests are added if needed and existing tests are updated. See [Running tests](https://github.com/input-output-hk/cardano-node-wiki/wiki/Running-tests) for more details
- [ ] Self-reviewed the diff

PR #776, opened 2025-03-07
Make 1-1 relationship of witness and policy ID in TxMintValue instead of 1-*

carbolymer:
# Changelog

```yaml
- description: |
    Make 1-1 relationship of witness and policy ID in TxMintValue instead of 1-*
    Remove exports: `parseValue`, `ParserValueRole`
    Add new type `PolicyAssets` representing minted assets within a single PolicyId
    Add `mkTxMintValue` helper function
  type:
```

# Context

Fixes: 
 - https://github.com/IntersectMBO/cardano-api/issues/769

Previous related changes to `TxMintValue`:
 - https://github.com/IntersectMBO/cardano-api/pull/663/

Used in: 
- https://github.com/IntersectMBO/cardano-cli/pull/1085

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/src/Cardano/Api/Internal/Value.hs
+++ b/cardano-api/src/Cardano/Api/Internal/Value.hs
@@ -25,6 +25,10 @@ module Cardano.Api.Internal.Value
   , negateValue
   , negateLedgerValue
   , calcMinimumDeposit
+  , PolicyAssets (..)
+  , policyAssetsToValue
+  , valueToPolicyAssets
+  , multiAssetToPolicyAssets
 
     -- ** Ada \/ L.Coin specifically
   , Lovelace
@@ -43,6 +47,8 @@ module Cardano.Api.Internal.Value
     -- ** Rendering
   , renderValue
   , renderValuePretty
+  , renderMultiAsset
+  , renderMultiAssetPretty
 
     -- * Internal conversion functions
   , toByronLovelace
@@ -50,6 +56,7 @@ module Cardano.Api.Internal.Value
   , fromShelleyDeltaLovelace
   , toMaryValue
   , fromMaryValue
+  , fromMultiAsset
   , fromLedgerValue
   , toLedgerValue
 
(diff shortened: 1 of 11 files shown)


cardano-api/src/Cardano/Api/Internal/Value.hs
```diff
@@ -303,27 +313,86 @@ fromLedgerValue sbe v =
...
+fromMaryAssetName (Mary.AssetName n) = AssetName $ Short.fromShort n
 
 -- | Calculate cost of making a UTxO entry for a given 'Value' and
 -- mininimum UTxO value derived from the 'ProtocolParameters'
 calcMinimumDeposit :: Value -> Lovelace -> Lovelace
 calcMinimumDeposit v =
   Mary.scaledMinDeposit (toMaryValue v)
 
+-- | Map of non-ADA assets with their quantity, for a single policy
+newtype PolicyAssets = PolicyAssets (Map AssetName Quantity)
```
Jimbo4350, line 342: Almost everywhere `PolicyAssets` is used in a `Map` i.e `Map PolicyId PolicyAssets`. Why not make this type:

`newtype PolicyAssets = PolicyAssets (Map PolicyId (Map AssetName Quantity))`?

And if we're doing that, we can just use the ledger type directly `newtype MultiAsset = MultiAsset (Map PolicyID (Map AssetName Integer))`

Jimbo4350 (commented):
One comment

cardano-api/src/Cardano/Api/Internal/Value.hs
carbolymer: I agree. But the main purpose of this type is to use it in TxMintValue where we have one witness per policy ID. So we need some type aggregating assets and quantities. You can't do it in a nice way if you have a type `Map PolicyId (Map AssetName Quantity)` - you'd have to keep two maps and keeps them in sync... cumbersome and error prone.

carbolymer: An alternative would be to just use `Map AssetName Quantity` there, but it's annoying to handle: you'd need to write operations like merging or negating by hand.

Jimbo4350 (approved):

Issue #668, opened 2024-10-31
[FR] - Add extended Pool Operator Signing/Verification Key Types


gitmachtl:
As part of a desaster recovery or a lost hardware device, if you derive stake pool cold keys from mnemonics, you get extended ones.

Currently it looks like that the extended versions of both the secret key and the verification key are missing in the api/cli.

Those would be like:

```json
{
      "type": "StakePoolExtendedSigningKey_ed25519_bip32",
      "description": "Stake Pool Operator Signing Key",
      "cborHex": "588078e35b66cbb23cbc623d81623e34ad157a667cfc4f843649236e7b7fb322b65b6d2c6afe2ebb5923875e776eaac1b21002253bb6b0003cc6a9d8c45ab5dd30995474de0dd906dae02e33cad45bb24a21f1ade6af57457b6ea171985412cd35d070a3adf6f2ecd799fa127e536d2c93c2c1946d11fb92c5608dfb3acef8eb72f1"
}
```
```json
{
      "type": "StakePoolExtendedVerificationKey_ed25519_bip32",
      "description": "Stake Pool Operator Verification Key",
      "cborHex": "58405474de0dd906dae02e33cad45bb24a21f1ade6af57457b6ea171985412cd35d070a3adf6f2ecd799fa127e536d2c93c2c1946d11fb92c5608dfb3acef8eb72f1"
}
```

Those `types` should be added, so we can also use them normally with all pool key operations.

Also we don't have a bech specification yet in CIP-005 for that. The bech prefix `pool_sk` and `pool_vk` are only for normal ones, not extended ones.

smelc:
Maybe for you @palas?

smelc:
cc @CarlosLopezDeLara

gitmachtl:
I think this would be an easy and fast one to implement? Just that we have the pool stuff feature-complete?

gitmachtl:
❤

PR #781, opened 2025-03-18
Add pool operator extended key support

palas:
# Changelog

```yaml
- description: |
    Add support for stake pool extended key
  type:
  - feature
```

# Context

Tries to address https://github.com/IntersectMBO/cardano-api/issues/668. Work in progress

# How to trust this PR

Don't trust it.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/src/Cardano/Api/Internal/Keys/Shelley.hs
+++ b/cardano-api/src/Cardano/Api/Internal/Keys/Shelley.hs
@@ -1,11 +1,14 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DeriveAnyClass #-}
 {-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GADTs #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE InstanceSigs #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TypeFamilies #-}
 -- The Shelley ledger uses promoted data kinds which we have to use, but we do
 -- not export any from this API. We also use them unticked as nature intended.
@@ -37,6 +40,7 @@ module Cardano.Api.Internal.Keys.Shelley
   , VerificationKey (..)
   , SigningKey (..)
   , Hash (..)
+  , AnyStakePoolKey (..)
   )
 where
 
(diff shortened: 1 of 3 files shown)


cardano-api/src/Cardano/Api/Internal/Keys/Shelley.hs
```diff
@@ -411,6 +428,10 @@ instance SerialiseAsRawBytes (Hash StakeKey) where
     maybeToRight (SerialiseAsRawBytesError "Unable to deserialise Hash StakeKey") $
       StakeKeyHash . Shelley.KeyHash <$> Crypto.hashFromBytes bs
 
+instance SerialiseAsBech32 (Hash StakeKey) where
```
Jimbo4350, line 431: How do we disambiguate between a stake key hash that belongs to an address not associated with a stake pool? This says every stake key hash is associated with a pool.

cardano-api/src/Cardano/Api/Internal/Keys/Shelley.hs
```diff
@@ -1658,6 +1679,50 @@ instance CastSigningKeyRole GenesisUTxOKey PaymentKey where
...
+
+instance FromCBOR AnyStakePoolVerificationKey where
+  fromCBOR :: Decoder s AnyStakePoolVerificationKey
+  fromCBOR =
+    peekTokenType >>= \case
+      TypeListLen ->
+        decodeListLenOf 2 >> do
+          tag <- fromCBOR
+          case tag of
+            0 -> AnyStakePoolNormalVerificationKey <$> fromCBOR
```
Jimbo4350, line 1703: Introducing a custom serialization format is not a good idea. We've been down this route before with the transaction body and users complained about expecting the CDDL format instead of our own custom format.  The correct way to go about this is to open a CIP and get buy in from the community first.

cardano-api/src/Cardano/Api/Internal/OperationalCertificate.hs
```diff
@@ -44,14 +44,14 @@ import Data.Word
 data OperationalCertificate
```
Jimbo4350, line 44: What about:

```
data OperationalCertificate a
  = OperationalCertificate
      !(Shelley.OCert StandardCrypto)
      !(VerificationKey a)
```

We can then shift the differentiation between normal and extended keys to the cli and avoid `AnyStakePoolVerificationKey` in `cardano-api`.

cardano-api/src/Cardano/Api/Internal/Keys/Shelley.hs
palas: Hm, so associated with a stake pool would be `Hash StakePoolKey`?
And I guess `stake_vkh` for `Hash StakeKey`?
Let me check where this is used

cardano-api/src/Cardano/Api/Internal/OperationalCertificate.hs
palas: But `cardano-cli` won't be able to know which one it is when deserialising, unless either:
- We write it in the envelope
- We have separate flags for operational certificates depending on whether they have a normal or extended key

Jimbo4350: Look at functions like `deserialiseAnyVerificationKeyTextEnvelope`.

Jimbo4350: And look at my cli PR: https://github.com/IntersectMBO/cardano-cli/pull/1087

Jimbo4350: E.g https://github.com/IntersectMBO/cardano-cli/pull/1087/files#diff-35d52f09b905c62b276e4ba31258c58e53061d2ae9e52b02298e8720c6f0be8cR27

For now I'm putting this in the cli to mature a but but we may end up moving it to `cardano-api` once we are happy with it.

palas: But that depends on the envelope type, right?. So basically we would need to define a new type for operational certificates that have an extended key? Like "OperationalCertificateWithExtendedStakePoolKey"

cardano-api/src/Cardano/Api/Internal/Keys/Shelley.hs
palas: Ah, it is not used, I will just remove it

cardano-api/src/Cardano/Api/Internal/OperationalCertificate.hs
palas: Ah, actually, I think I found a better solution. I think I can just use `castVerificationKey`

cardano-api/src/Cardano/Api/Internal/Keys/Shelley.hs
palas: It may not be necessary, because it seems I can just remove it

Jimbo4350 (changes requested):
Have you created an operational certificate with an extended key and tried to deserialise it from disk?

Jimbo4350 (changes requested):
https://input-output-rnd.slack.com/archives/C031GFLC3DF/p1745342421576939

Jimbo4350 (approved):
LGTM! Squash the commits.

PR #790, opened 2025-03-27
Delegate query version decision to Consensus

jasagredo:
# Changelog

```yaml
- description: |
    - Delegate decision on supported versions for queries to Consensus.
  type:
```

# Context

Consensus already maps the versions to the specific queries. Having this mapping also in `cardano-api` makes it possible for a mismatch to exist, leading to very obscure errors. Moreover, Consensus is also capable of setting upper bounds on versions, which was not possible with the old approach in `cardano-api`.

# How to trust this PR

N/A

# Checklist

- [ ] ~New tests are added if needed and existing tests are updated. See [Running tests](https://github.com/input-output-hk/cardano-node-wiki/wiki/Running-tests) for more details~
- [x] Self-reviewed the diff


jasagredo:
This is set to merge into `aniketd/release-10-3` but just so that the diff is reviewable. I think it can wait until after the release of node 10.3.

cardano-api/src/Cardano/Api/Internal/IPC/Version.hs
```diff
@@ -21,15 +40,73 @@ import Ouroboros.Network.NodeToClient.Version (NodeToClientVersion (..))
...
+      -> Consensus.QueryIfCurrent xs result
+      -> Either UnsupportedNtcError Bool
+    go (Consensus.EraNodeToClientEnabled supportedNtc :* _) (Consensus.QZ q') =
+      Right $ Consensus.querySupportedVersion q' supportedNtc
+    go (Consensus.EraNodeToClientDisabled :* _) Consensus.QZ{} =
+      Left undefined
+    go Nil qq = case qq of {}
+    go (_ :* n) (Consensus.QS q') = go n q'
 
-data UnsupportedNtcVersionError = UnsupportedNtcVersionError !MinNodeToClientVersion !NodeToClientVersion
```
palas, line 34: The only potential issue I see is we lose information about what version is required. But on the other hand, I imagine it is going to make maintainability easier and prevent potential discrepancy errors, so it seems positive overall to me. But I will leave approval to @Jimbo4350

cardano-api/src/Cardano/Api/Internal/IPC/Version.hs
```diff
@@ -21,15 +40,73 @@ import Ouroboros.Network.NodeToClient.Version (NodeToClientVersion (..))
...
+        ( \x ->
+            if not x
+              then Just (UnsupportedNtcVersionError ntc)
+              else Nothing
+        )
+        $ case q of
+          Consensus.GetChainBlockNo -> Right True
+          Consensus.GetSystemStart -> Right True
+          Consensus.GetChainPoint -> Right True
+          Consensus.GetLedgerConfig -> Right True
```
amesgen, line 73: Ah wait, shouldn't there be conditional logic here for `GetLedgerConfig`? Was this missed in
https://github.com/intersectMBO/ouroboros-consensus/commit/b0d625eceb0a30b4eac8139dea7f23df3bf4260d?

jasagredo: Hmmm. I seem to recall I restricted it to QueryVersion3 but I don't know how I can extract that one here in api. Let me check

amesgen: Ah right, this query isn't HFC-specific, so it is guarded by `QueryVersion3` :+1: But indeed, it doesn't seem easily accessible from our API ATM.

jasagredo: It is inconvenient to use indeed. I'm going to try to fix it in Consensus for the next version.

cardano-api/src/Cardano/Api/Internal/IPC/Version.hs
Jimbo4350: @jasagredo is there any way we can get a minimum version from consensus?

Jimbo4350 (approved):
LGTM with the exception of: https://github.com/IntersectMBO/cardano-api/pull/790#discussion_r2017409873

However I think not having to maintain the NTC versions trumps this issue. However it would be great if consensus could expose a minimum required version.

cardano-api/src/Cardano/Api/Internal/IPC/Version.hs
carbolymer: I agree, for diagnostics purpose, it would be nice if the query version could be included in the error.

cardano-api/src/Cardano/Api/Internal/IPC/Version.hs
```diff
@@ -21,15 +41,68 @@ import Ouroboros.Network.NodeToClient.Version (NodeToClientVersion (..))
...
+      (Proxy @(Consensus.CardanoBlock StandardCrypto))
+      Map.!? ntc
 
-type MinNodeToClientVersion = NodeToClientVersion
+  getNTC
+    :: forall result
+     . NP Consensus.EraNodeToClientVersion (Consensus.CardanoShelleyEras StandardCrypto)
+    -> Consensus.QueryIfCurrent (Consensus.CardanoShelleyEras StandardCrypto) result
+    -> Either UnsupportedNtcError ()
+  getNTC = go
```
carbolymer, line 86: I think it does checking rather than just getting
```suggestion
  checkNtc
    :: forall result
     . NP Consensus.EraNodeToClientVersion (Consensus.CardanoShelleyEras StandardCrypto)
    -> Consensus.QueryIfCurrent (Consensus.CardanoShelleyEras StandardCrypto) result
    -> Either UnsupportedNtcError ()
  checkNtc = go
```

carbolymer (approved):

cardano-api/src/Cardano/Api/Internal/IPC/Version.hs
jasagredo: Im creating a new function in consensus `querySupportedVersions` (notice the s at the end) which will return a list of versions in which this query is supported.

cardano-api/src/Cardano/Api/Internal/IPC/Version.hs
jasagredo: This function is gone

cardano-api/src/Cardano/Api/Internal/IPC/Version.hs
jasagredo: The new version of this PR now properly produces a list of versions in which the query is supported.

cardano-api/src/Cardano/Api/Internal/Convenience/Query.hs
```diff
@@ -77,14 +77,15 @@ renderQueryConvenienceError (QueryEraMismatch (EraMismatch ledgerEraName' otherE
     <> " era."
 renderQueryConvenienceError ByronEraNotSupported =
   "Byron era not supported"
-renderQueryConvenienceError (QceUnsupportedNtcVersion (UnsupportedNtcVersionError minNtcVersion ntcVersion)) =
+renderQueryConvenienceError (QceUnsupportedNtcVersion (UnsupportedNtcVersionError ntcVersion supportedVersions)) =
```
carbolymer, line 80:
```suggestion
renderQueryConvenienceError (QceUnsupportedNtcVersion (UnsupportedNtcVersionError ntcVersion supportedVersion)) =
```
shouldn't that be a singular noun?

carbolymer (commented):
LGTM!

cardano-api/src/Cardano/Api/Internal/Convenience/Query.hs
jasagredo: Nope, that is a list of supported versions, because of the fact that there might be disjoint segments in the list of versions.

jasagredo:
I think I will need someone to click the merge button for me. Github says I'm not authorized.

We released a new version of consensus with this change, so that we don't need SRPs in the cabal.project.

Ping @carbolymer @Jimbo4350 @palas

Jimbo4350 (approved):

PR #801, opened 2025-04-03
Improve IPC haddock

palas:
# Changelog

```yaml
- description: |
    Add examples to haddock documentation of `Cardano.Api.Internal.IPC` module. Also add tests with examples.
  type:
  - test
  - documentation
```

# Context

In order to make `cardano-api`, this PR adds step-by-step examples on how to do queries and submit transactions to the node by using the IPC node-to-client protocol with the tools provided by the `Cardano.Api.Internal.IPC` module. The PR also includes a test that does the same things that the examples in the documentation do. The node side of the IPC is implemented as a mock of real interaction with a node connected to preview.

# How to trust this PR

Read documentation and check code quality of the included test.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-api/src/Cardano/Api/Internal/IPC.hs
+++ b/cardano-api/src/Cardano/Api/Internal/IPC.hs
@@ -129,7 +129,7 @@ module Cardano.Api.Internal.IPC
     --   Left Shelley.AFPointNotOnChain -> error "Error, point queried is not on chain!"
     -- @
     --
-    -- 'AFPointToolOld' and 'AFPointNotOnChain' errors should not occur when querying with
+    -- 'AFPointTooOld' and 'AFPointNotOnChain' errors should not occur when querying with
     -- either 'VolatileTip' or 'ImmutableTip'.
 
     -- *** Obtaining the UTXO set


cardano-api/src/Cardano/Api/Internal/IPC.hs
```diff
@@ -12,8 +12,181 @@
...
+    -- ** Constructing connection information
+
+    -- | Independently of whether we want to query the node or submit transactions,
+    -- the first thing we need to do is to gather the connection information.
+    --
+    -- We need three pieces of information:
+    --
+    --    * The number of slots per epoch for the network the node is connected to.
+    --      We can obtain this information by looking for the @epochLength@ key in
+    --      the @shelley-genesis.json@ file that the node is using to connect to the
```
Jimbo4350, line 59: You can also query the `GenesisParameters`. I would be hesitate to put `86_400` as they may one day re-spin the network.

cardano-api/src/Cardano/Api/Internal/IPC.hs
```diff
@@ -12,8 +12,181 @@
...
+    --      network. For the preview network, this is currently @86_400@.
+    --    * In the case we are connecting to a testnet, we also need to find out
+    --      the magic number for the network the node is connected to. This can be
+    --      obtained by looking for the @networkMagic@ key in the @shelley-genesis.json@
+    --      file that the node is using to connect to the network. For the preview
+    --      network, the magic number is currently @2@.
```
Jimbo4350, line 65: Network id, not "magic number". You can also query the `GenesisParameters` for this.

cardano-api/src/Cardano/Api/Internal/IPC.hs
```diff
@@ -12,8 +12,181 @@
...
+    -- | Depending on the type of query we want to perform, we usually need
+    -- to know what era the node is currently in. We can hardcode this
+    -- information by using one of the constructors of the 'ShelleyBasedEra' type.
+    -- But we can also obtain this information from the node, as follows:
+    --
+    -- @
+    -- eEra <- runExceptT $ Api.queryNodeLocalState connectionInfo Network.VolatileTip Api.QueryCurrentEra
+    -- @
+    --
+    -- Here, 'VolatileTip' means we want to get the information out of the most recent node that the
```
Jimbo4350, line 101: out of the most recent block

cardano-api/src/Cardano/Api/Internal/IPC.hs
```diff
@@ -12,8 +12,181 @@
...
+    --       connectionInfo
+    --       Network.VolatileTip
+    --       (Api.QueryInEra (Api.QueryInShelleyBasedEra sbe (Api.QueryUTxO Api.QueryUTxOWhole)))
+    -- @
+    --
+    -- This time, @eUtxo@ has a nested type of @Either AcquiringFailure (Either EraMismatch (UTxO era))@.
+    -- So we can unwrap it as follows:
+    --
+    -- @
+    -- utxo <- case eUtxo of
```
Jimbo4350, line 148: We also have `Cardano.Api.Internal.Convenience.Query`

cardano-api/src/Cardano/Api/Internal/IPC.hs
```diff
@@ -12,8 +12,181 @@
...
+    --       )
+    --   Left Shelley.AFPointTooOld -> error "Error, point queried in the chain is too old!"
+    --   Left Shelley.AFPointNotOnChain -> error "Error, point queried is not on chain!"
+    -- @
+    --
+    -- The obtained @utxo@ is a standard @Map@ of type @Map TxIn (TxOut CtxUTxO era)@.
+
+    -- ** Submitting a transaction
+
+    -- | Let's assume we have a signed transaction of the right era that we want to submit to the node,
```
Jimbo4350, line 166: "of the latest era"

cardano-api/test/cardano-api-test/Test/Cardano/Api/IPC.hs
```diff
@@ -0,0 +1,258 @@
...
+  => FilePath
+  -- ^ Directory to create the socket file in.
+  -> FilePath
+  -- ^ Name of the unix socket file to create
+  -> FilePath
+  -- ^ File to read the raw replay data from.
+  -> (FilePath -> m a)
+  -- ^ Action to run while the server is being mocked.
+  -> m a
+mockNode work fileName rawReplayData action = do
```
Jimbo4350, line 220: We're not testing any real functionality here. I think the haddocks examples are enough. I wouldn't bother with this property test.

Jimbo4350 (commented):
Nice work. Can you ask @olgahryniuk to have a look?

cardano-api/test/cardano-api-test/Test/Cardano/Api/IPC.hs
palas: Yes, the main reason is that I had to write the code to make sure the examples I wrote in the Haddock work, so I thought that it would be better this than trashing the code. If we want to update it in the future we don't have to rewrite everything. And it has two advantages:
- If the syntax for the API commands changes, we will know that we have to update the tutorial and how we have to change it (because of type errors)
- If we manage to set up doctest, we could actually use this as executable documentation, otherwise it won't give the expected results.

cardano-api/src/Cardano/Api/Internal/IPC.hs
```diff
@@ -12,8 +12,181 @@
 {-# OPTIONS_GHC -Wno-unticked-promoted-constructors #-}
 
 -- | Node IPC protocols
+--
+-- This module provides the client side of the node-to-client IPC protocol
+-- used to communicate with a local Cardano node. This can be used to
+-- query the node for information, to submit transactions, and even to
+-- get historical information aobut the chain by using the
```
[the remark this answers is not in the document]
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/commit/370579586500a844f50d61a97bca257ced762546

cardano-api/src/Cardano/Api/Internal/IPC.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/commit/370579586500a844f50d61a97bca257ced762546

cardano-api/src/Cardano/Api/Internal/IPC.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/commit/370579586500a844f50d61a97bca257ced762546

cardano-api/src/Cardano/Api/Internal/IPC.hs
palas: I can write "at the time of writing, it is @86_400@, but it can change."

palas: Addressed here: https://github.com/IntersectMBO/cardano-api/commit/9ba65cd705f4215bee754ebee78427fe04a7ee2a

cardano-api/src/Cardano/Api/Internal/IPC.hs
palas: I added a note: https://github.com/IntersectMBO/cardano-api/commit/9ba65cd705f4215bee754ebee78427fe04a7ee2a

cardano-api/src/Cardano/Api/Internal/IPC.hs
```diff
@@ -12,8 +12,187 @@
...
+    --      node is connected to. It can be obtained by inspecting the @epochLength@
+    --      key in the @shelley-genesis.json@ file used by the node. On the preview
+    --      network, the value is, at the time of writing, @86_400@, but it can
+    --      change. This value and other genesis parameters can also be obtained
+    --      using the 'QueryGenesisParameters' query.
+    --    * __Network identifier__. When connecting to a testnet, the network identifier
+    --      is also required. It can be obtained by looking for the @networkId@
+    --      obtained by looking up the @networkMagic@ key in the @shelley-genesis.json@
+    --      file that the node is uses to connect to the network. For the preview
+    --      network, the current identifier is @2@.
```
olgahryniuk, line 67:
```suggestion
    --    * __Network identifier__. When connecting to a testnet, the network identifier
    --      is also required. It can be obtained by looking for the @networkId@
    --      obtained by looking up the @networkMagic@ key in the @shelley-genesis.json@
    --      file that the node uses to connect to the network. For the preview
    --      network, the current identifier is @2@.
```

olgahryniuk (approved):

cardano-api/test/cardano-api-test/Test/Cardano/Api/IPC.hs
```diff
@@ -0,0 +1,284 @@
...
+    -- Server set-up (resource acquisition)
+    ( do
+        socketFile <- H.noteTempFile work fileName
+        sock <- H.evalIO $ do
+          sock <- socket AF_UNIX Stream defaultProtocol
+          bind sock (SockAddrUnix socketFile)
+          listen sock 1
+          return sock
+        threadId <-
+          forkFinally
```
carbolymer, line 258: Why do you need to fork thread here? I believe you could achieve the result with just `bracket`

cardano-api/test/cardano-api-test/Test/Cardano/Api/IPC.hs
```diff
@@ -0,0 +1,284 @@
...
+            & Api.setTxFee txFee
+    eSigningKey <-
+      H.evalIO $
+        Api.readFileTextEnvelope
+          (Api.AsSigningKey Api.AsPaymentKey)
+          "test/cardano-api-test/files/input/ipc-mock/key.skey"
+
+    let signingKey = case eSigningKey of
+          Right sk -> sk
+          Left err -> error $ "Error deserialising signing key: " ++ show err
```
carbolymer, line 169:
```suggestion
          Left err -> fail $ "Error deserialising signing key: " ++ show err
```

cardano-api/test/cardano-api-test/Test/Cardano/Api/IPC.hs
```diff
@@ -0,0 +1,284 @@
...
+        )
+
+    -- We unwrap the error and make sure it is not Byron
+    Api.AnyShelleyBasedEra sbe :: Api.AnyShelleyBasedEra <- case eEra of
+      Right (Api.AnyCardanoEra era) ->
+        Api.caseByronOrShelleyBasedEra
+          (fail "We are in Byron era")
+          (return . Api.AnyShelleyBasedEra)
+          era
+      Left Shelley.AFPointTooOld -> fail "Error, point queried in the chain is too old!"
```
carbolymer, line 76: I think `fail` does not give you the place where failure has occured

cardano-api/test/cardano-api-test/Test/Cardano/Api/IPC.hs
carbolymer: I'm agreeing here with @Jimbo4350. I'm not sure that those advantages are that big.

I think a better approach would be to mock the whole network layer in the `queryNodeLocalState` to make it in-memory only, so that you could set arbitrary haskell values on the other end. Currently the dependency on the networking code is hardcoded there (which shouldn't be).

But it's a much bigger task definitely out of scope for this PR.

Jimbo4350: I still think this is overkill for now. If you can get some sort of mocking functionality from the network team then go for it.

Jimbo4350 (commented):
LGTM! One comment regarding the removal of the mocking test.

cardano-api/src/Cardano/Api/Internal/IPC.hs
```diff
@@ -12,8 +12,187 @@
...
+    -- is not running Byron:
+    --
+    -- @
+    -- Api.AnyShelleyBasedEra sbe :: Api.AnyShelleyBasedEra <- case eEra of
+    --   Right (Api.AnyCardanoEra era) ->
+    --     Api.caseByronOrShelleyBasedEra
+    --       (error "Error, we are in Byron era")
+    --       (return . Api.AnyShelleyBasedEra)
+    --       era
+    --   Left Shelley.AFPointTooOld -> error "Error, point queried in the chain is too old!"
```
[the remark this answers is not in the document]
palas: What do you mean? Without the `Shelley.`? I don't understand

cardano-api/src/Cardano/Api/Internal/IPC.hs
```diff
@@ -12,8 +12,187 @@
 {-# OPTIONS_GHC -Wno-unticked-promoted-constructors #-}
 
 -- | Node IPC protocols
+--
+-- This module provides the client side of the node-to-client interprocess
+-- communication (IPC) for interacting with a local Cardano node. It supports
+-- querying the node for information, submitting transactions, and retrieving
```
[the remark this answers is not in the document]
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/801/commits/c16bb159613e95ac3d478690f517c3f3133a30c4

cardano-api/src/Cardano/Api/Internal/IPC.hs
```diff
@@ -12,8 +12,187 @@
...
+    --           ++ " but it was "
+    --           ++ show ledgerEraName
+    --       )
+    --   Left Shelley.AFPointTooOld -> error "Error, point queried in the chain is too old!"
+    --   Left Shelley.AFPointNotOnChain -> error "Error, point queried is not on chain!"
+    -- @
+    --
+    -- Alternatively, to avoid nested result types, you can use convenience
+    -- functions and types from "Cardano.Api.Internal.Convenience.Query".
+    -- The obtained @utxo@ variable is a standard @Map@ of type @Map TxIn (TxOut CtxUTxO era)@.
```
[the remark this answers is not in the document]
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/801/commits/a32237d2a110f50ebbdd1f8897d11d200b498796

cardano-api/src/Cardano/Api/Internal/IPC.hs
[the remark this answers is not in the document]
Jimbo4350: image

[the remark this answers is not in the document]
Jimbo4350: Therre is a spelling mistake "AFPointToolOld"

cardano-api/test/cardano-api-test/Test/Cardano/Api/IPC.hs
palas: Now it is no longer applicable

cardano-api/src/Cardano/Api/Internal/IPC.hs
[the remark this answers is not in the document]
palas: Ah, I see it now, it was a couple of line below.
Addressed here: https://github.com/IntersectMBO/cardano-api/pull/801/commits/d6eda1567e112009be24b42532f2349df2928bdc

Jimbo4350 (approved):

Issue #803, opened 2025-04-04
[FR] - Extra `ScriptData` for reference inputs with hashed datums


mmontin:
**Internal/External**
*External*

**Area**
*Plutus*

**Describe the feature you'd like**

In july 2024 I posted an [issue](https://github.com/IntersectMBO/cardano-api/issues/600) to report that cardano-api was not providing a way to add extra script data to a transaction body that does not correspond to an output being paid with a datum hash. 

Later on, this feature was implemented in [this PR](https://github.com/IntersectMBO/cardano-api/pull/640).

However, it was very quickly removed in [this commit](https://github.com/IntersectMBO/cardano-api/commit/22417a0f7fce857324dd01389afbeb9fee948cac).

Recently, I dove deeper into the ledger rules and realized that my original feature request was only half relevant. It turns out that the ledger rules forbid to add a resolved datum for a PK input with a datum hash, which I was not aware off (and which I don't know the reason for). I am assuming this is the reason why this new feature was quickly removed after being introduced following my issue.

However, I say half relevant, because the ledger does allow to provide extra resolved datums for reference inputs with hashed datums, which, following the removal of this feature, is still not possible to achieve from within cardano-api, as far as I can see. Feel free to let me know if either this feature is already present elsewhere, or if it is a deliberate choice not to provide it.

I wish this feature would be brought back, even if solely dedicated to reference inputs. Whenever a reference input would be defined in a transaction body, users could be given the opportunity to bind a `ScriptData` to that reference input, similarly to how they can be attached to outputs. It does not necessarily need to be a full new field within a body, which, I agree, was a reasonable solution if those extra datums could come from different places. 

**Describe alternatives you've considered**

I am currently relying on the hack I described in my original issue, where i have to manually deconstruct the generated `ShelleyTxBody`, add the script data and re-bundle it after recomputing the script integrity hash. In case anybody is interested, this hack is performed [here](https://github.com/tweag/cooked-validators/blob/9cb80810d982c9eccd3f7710a996d20f944a95ec/src/Cooked/MockChain/GenerateTx/Body.hs#L86).

Jimbo4350:
> It turns out that the ledger rules forbid to add a resolved datum for a PK input with a datum hash, which I was not aware off (and which I don't know the reason for). I am assuming this is the reason why this new feature was quickly removed after being introduced following my issue.

Yes you are correct. I should have explained in the PR description. 

[...]

PR #814, opened 2025-04-15
Allow providing of actual datum for reference inputs

carbolymer:
# Changelog

```yaml
- description: |
    Allow providing of actual datum for reference inputs in `TxInsReference`.
  type:
```

# Context

This PR adds a method of adding actual datum for the datum hashes present in reference inputs.

Tested in: 
- https://github.com/IntersectMBO/cardano-node/pull/6174

Integrated in CLI:
- https://github.com/IntersectMBO/cardano-cli/pull/1169

Closes https://github.com/IntersectMBO/cardano-api/issues/803

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/src/Cardano/Api/Internal/Tx/Body.hs
+++ b/cardano-api/src/Cardano/Api/Internal/Tx/Body.hs
@@ -301,6 +301,8 @@ module Cardano.Api.Internal.Tx.Body
     -- ** Other transaction body types
   , TxInsCollateral (..)
   , TxInsReference (..)
+  , TxInsReferenceDatums
+  , getReferenceInputDatumMap
   , TxReturnCollateral (..)
   , TxTotalCollateral (..)
   , TxFee (..)
@@ -570,16 +572,31 @@ deriving instance Eq (TxTotalCollateral era)
 
 deriving instance Show (TxTotalCollateral era)
 
-data TxInsReference era where
-  TxInsReferenceNone :: TxInsReference era
+data TxInsReference build era where
+  TxInsReferenceNone :: TxInsReference build era
   TxInsReference
     :: BabbageEraOnwards era
     -> [TxIn]
-    -> TxInsReference era
+    -- ^ A list of reference inputs
+    -> TxInsReferenceDatums build
+    -- ^ A set of datums, whose hashes are referenced in UTXO of reference inputs. Those datums will be inserted
+    -- to the datum map available to the scripts. Note that inserting a datum with hash not present in the reference
+    -- input will result in an error on transaction submission.
+    -> TxInsReference build era
 
-deriving instance Eq (TxInsReference era)
+deriving instance Eq (TxInsReference build era)
 
-deriving instance Show (TxInsReference era)
+deriving instance Show (TxInsReference build era)
+
+-- | The actual datums, referenced by hash in the transaction reference inputs.
+type TxInsReferenceDatums build = BuildTxWith build (Set HashableScriptData)
+
+getReferenceInputDatumMap
+  :: TxInsReferenceDatums build
+  -> Map (Hash ScriptData) HashableScriptData
+getReferenceInputDatumMap = \case
+  ViewTx -> mempty
+  BuildTxWith datumSet -> fromList $ map (\h -> (hashScriptDataBytes h, h)) $ toList datumSet
 
 -- ----------------------------------------------------------------------------
 -- Transaction fees
@@ -984,7 +1001,7 @@ data TxBodyContent build era
   = TxBodyContent
   { txIns :: TxIns build era
   , txInsCollateral :: TxInsCollateral era
-  , txInsReference :: TxInsReference era
+  , txInsReference :: TxInsReference build era
   , txOuts :: [TxOut CtxTx era]
   , txTotalCollateral :: TxTotalCollateral era
   , txReturnCollateral :: TxReturnCollateral CtxTx era
@@ -1075,25 +1092,36 @@ addTxInCollateral
   :: IsAlonzoBasedEra era => TxIn -> TxBodyContent build era -> TxBodyContent build era
 addTxInCollateral txInCollateral = addTxInsCollateral [txInCollateral]
 
-setTxInsReference :: TxInsReference era -> TxBodyContent build era -> TxBodyContent build era
+setTxInsReference :: TxInsReference build era -> TxBodyContent build era -> TxBodyContent build era
 setTxInsReference v txBodyContent = txBodyContent{txInsReference = v}
 
 modTxInsReference
-  :: (TxInsReference era -> TxInsReference era) -> TxBodyContent build era -> TxBodyContent build era
+  :: (TxInsReference build era -> TxInsReference build era)
+  -> TxBodyContent build era
+  -> TxBodyContent build era
 modTxInsReference f txBodyContent = txBodyContent{txInsReference = f (txInsReference txBodyContent)}
 
 addTxInsReference
-  :: IsBabbageBasedEra era => [TxIn] -> TxBodyContent build era -> TxBodyContent build era
-addTxInsReference txInsReference =
-  modTxInsReference
-    ( \case
-        TxInsReferenceNone -> TxInsReference babbageBasedEra txInsReference
-        TxInsReference era xs -> TxInsReference era (xs <> txInsReference)
-    )
+  :: Applicative (BuildTxWith build)
+  => IsBabbageBasedEra era
+  => [TxIn]
+  -> Set HashableScriptData
+  -> TxBodyContent build era
+  -> TxBodyContent build era
+addTxInsReference txInsReference scriptData =
+  modTxInsReference $
+    \case
+      TxInsReferenceNone -> TxInsReference babbageBasedEra txInsReference (pure scriptData)
+      TxInsReference era xs bScriptData' -> TxInsReference era (xs <> txInsReference) ((<> scriptData) <$> bScriptData')
 
 addTxInReference
-  :: IsBabbageBasedEra era => TxIn -> TxBodyContent build era -> TxBodyContent build era
-addTxInReference txInReference = addTxInsReference [txInReference]
+  :: Applicative (BuildTxWith build)
+  => IsBabbageBasedEra era
+  => TxIn
+  -> Maybe HashableScriptData
+  -> TxBodyContent build era
+  -> TxBodyContent build era
+addTxInReference txInReference mDatum = addTxInsReference [txInReference] . fromList $ maybeToList mDatum
 
 setTxOuts :: [TxOut CtxTx era] -> TxBodyContent build era -> TxBodyContent build era
 setTxOuts v txBodyContent = txBodyContent{txOuts = v}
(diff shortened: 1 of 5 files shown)


cardano-api/src/Cardano/Api/Internal/Tx/Body.hs
```diff
@@ -1742,11 +1761,11 @@ fromLedgerTxInsCollateral sbe body =
     sbe
 
 fromLedgerTxInsReference
-  :: ShelleyBasedEra era -> Ledger.TxBody (ShelleyLedgerEra era) -> TxInsReference era
+  :: ShelleyBasedEra era -> Ledger.TxBody (ShelleyLedgerEra era) -> TxInsReference ViewTx era
 fromLedgerTxInsReference sbe txBody =
   caseShelleyToAlonzoOrBabbageEraOnwards
     (const TxInsReferenceNone)
-    (\w -> TxInsReference w $ map fromShelleyTxIn . toList $ txBody ^. L.referenceInputsTxBodyL)
+    (\w -> TxInsReference w (map fromShelleyTxIn . toList $ txBody ^. L.referenceInputsTxBodyL) ViewTx)
```
Jimbo4350, line 1768: Ah yes, there is now way to figure out which datum hashes are relevant.

cardano-api/src/Cardano/Api/Internal/Tx/Body.hs
```diff
@@ -3053,17 +3081,30 @@ collectTxBodyScriptWitnessRequirements
 
 getSupplementalDatums
   :: AlonzoEraOnwards era
+  -> TxInsReference BuildTx era
```
Jimbo4350, line 3084: These technically wouldn't be supplemental datums as those correspond to tx outputs.

carbolymer:
@mmontin This should implement #803 . What do you think about the design here?

cardano-api/src/Cardano/Api/Internal/Tx/Body.hs
```diff
@@ -570,16 +572,20 @@ deriving instance Eq (TxTotalCollateral era)
...
-data TxInsReference era where
-  TxInsReferenceNone :: TxInsReference era
+data TxInsReference build era where
+  TxInsReferenceNone :: TxInsReference build era
   TxInsReference
     :: BabbageEraOnwards era
     -> [TxIn]
-    -> TxInsReference era
+    -- ^ A list of reference inputs
+    -> BuildTxWith build (Set HashableScriptData)
```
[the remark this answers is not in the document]
Jimbo4350: I like this: `type TxInsReferenceActualDatums build = BuildTxWith build (Set HashableScriptData)` 

It's more concise and we have the notion of empty via the `Set` type. Go ahead with the change and I'll have a look after.

Jimbo4350 (approved):
LGTM :+1: . Get input from the ledger team regarding naming and we can finalize that aspect.

cardano-api/src/Cardano/Api/Internal/Tx/Output.hs
```diff
@@ -14,30 +14,43 @@
 {-# LANGUAGE TypeOperators #-}
 
 module Cardano.Api.Internal.Tx.Output
```
mmontin, line 16: Changes in this file seem to be unrelated to the global PR. I don't mind it too much personally but I don't know what the practice are for cardano-api. I imagine you do ;)

cardano-api/test/cardano-api-test/Test/Cardano/Api/Transaction/Autobalance.hs
```diff
@@ -294,7 +294,7 @@ prop_make_transaction_body_autobalance_return_correct_fee_for_multi_asset = H.pr
         Nothing
 
   scriptWitReqsWithAsset <-
-    H.evalEither $ collectTxBodyScriptWitnessRequirements aeo balancedContentWithTxoutAsset
+    H.evalEither $ collectTxBodyScriptWitnessRequirements aeo utxos balancedContentWithTxoutAsset
```
mmontin, line 297: I guess this could be `empty` instead of `utxos` as it is a singleton that was built for other purposes. I think this would be more understandable and not make it look as if it is relevant here.

cardano-api/test/cardano-api-test/Test/Cardano/Api/Transaction/Autobalance.hs
```diff
@@ -324,7 +324,7 @@ prop_make_transaction_body_autobalance_return_correct_fee_for_multi_asset = H.pr
         Nothing
 
   scriptWitReqsBalanced <-
-    H.evalEither $ collectTxBodyScriptWitnessRequirements aeo balancedContent
+    H.evalEither $ collectTxBodyScriptWitnessRequirements aeo utxos balancedContent
```
mmontin, line 327: same as above

cardano-api/test/cardano-api-test/Test/Cardano/Api/Transaction/Autobalance.hs
```diff
@@ -453,7 +453,7 @@ prop_make_transaction_body_autobalance_multi_asset_collateral = H.propertyOnce $
...
```
mmontin, line 456: same

mmontin (commented):
Thanks a lot for implementing this ! I believe this feature, while somewhat a corner case, is very important to have.

In terms of the actual implementation, I have two general comments / remarks on top of the small comments I left here and there

1. I wonder why these additional datums are given as a set directly inside `TxInsReference` instead of input by input. This breaks the link as to which input is supposed to bear which datum hash. In other words, why not couple each `TxIn` with a `Maybe HashableScriptData` in a structure (let's call it `RefInputWithDatum`) and have `TxInsReference` take a list of those? 

2. About the need of providing a `UTxO` here and there. I understand the need for this structure. It allows to ensure that the provided datums do indeed correspond to the resolved variants of hashed datums in reference inputs. However, this will be caught later on in the pipeline anyway as the ledger with throw an error when they do not. So is this necessary to ensure this here? Admittedly, I don't have a full picture as to which properties need to be satisfied by `createTransactionBody` and are expected to hold by the caller. However, it makes the implementation more involved, and also forces the caller to build this UTxO and understand that it corresponds to the corner case of providing resolved datums to reference inputs. Additionally, since the current implementation seems to just ignore datums that do not correspond to hashed datums in ref inputs, there is no clear way to bypass this requirement for testing purposes. And I believe this is dangerous that this process is silent. This makes me wonder if there are other places in the code base where silent discards are performed.

cardano-api/src/Cardano/Api/Internal/Tx/Output.hs
carbolymer: Ah yes, those are unrelated. I've addressed one remark from other PR. I'll make a separate commit with that.

carbolymer:
@mmontin Good remarks thanks!

ad 2. I agree. Passing UTXO around mostly gets in the way, without any real benefit here. I think we could skip it, ditch the filtering and do the validation at transaction submission step (by ledger).

ad 1. Assuming we remove the necessity of passing UTXO: tying `TxIns` with datum values would only make sense if we have UTXOs of reference inputs in the scope, and we would validate that respective datum is at the right input. I agree that there's a link there, but I'm not sure there's a value in that information: this link gets "lost" when the transaction gets submitted to the ledger. So the constructor `TxInsReference` would potentially look similar to this:
```haskell
 TxInsReference
    :: BabbageEraOnwards era
    -> [( TxIn
        , BuildTx build (Maybe HashableScriptData)
        )
       ]
    -> TxInsReference build era
```
which would complicate handling of the type a bit: the values would have to be `zip`ped and `unzip`ed when constructing and deconstructing the type.

mmontin:
About location of datums in the body. Your point makes sense, it would be unwise to complicate the type for no valid reason. And to be honest, the current version where the datum are grouped into a set would be suitable enough for my use case and overall satisfactory, in my opinion. However, I see 2 reasons why the other option also makes sense, based on my current understanding of cardano-api and my current biases as to how I use it in `cooked-validators`:

1. **Consistency**: Currently, the second use case for supplemental datums (when paying outputs with hashed datums) is handled locally, within the `TxOut` itself. Technically, `TxOutSupplementalDatum` is the same as `TxOutDatumHash` with the exception that the full datum will be given to the transaction. This means that `TxOutSupplementalDatum` could technically be removed and, instead, the `txOuts` field of the `TxBodyContent` could be typed:
```
TxOuts
   :: BabbageEraOnwards era -- not sure about Babbage, but not so relevant for the example
   -> [TxOut ctx era]
   -> Set (BuildTx build HashableScriptData)
   -> TxOuts build era
```
I am definitely _not_ advocating for such a change, just pointing out that the choice was made to have these supplemental datums be handled within each `TxOut` instead of grouped outside. Admittedly, this makes a lot more sense here, because the `TxOut`s naturally contain a datum, as opposed to `TxIn` which are merely ids. That being said, handling supplemental datums similarly for the second possible use case would make sense and likely be more familiar to future users relying on this new feature. 

2. **Locality**: This one is more related to my personal bias, but I like when the information is as close as possible to where it belongs. Granted, this will be lost later on in the pipeline, but I see cardano-api transaction bodies as an opportunity to layout pieces of information in a more local way as the transaction produced by the ledger and given to the scripts. Having these supplemental datums defined locally would _feel_ just right (for me at least) and this is the place where I would naturally search for this information. That being said, having the datum a bit higher up does not hurt and they could also be found quite easily there.

I feel like this second point is not quite as strong as I thought when I decided to write it, because it's mostly just based on a feeling, and I leave it to you to take it into account or not. Regardless, I hope these points are of help in making the final decision.

Jimbo4350:
> 1. I wonder why these additional datums are given as a set directly inside `TxInsReference` instead of input by input. This breaks the link as to which input is supposed to bear which datum hash. In other words, why not couple each `TxIn` with a `Maybe HashableScriptData` in a structure (let's call it `RefInputWithDatum`) and have `TxInsReference` take a list of those?

What would be the purpose? You would lose the connection as soon as you construct a valid transaction.

>  However, this will be caught later on in the pipeline anyway as the ledger with throw an error when they do not.

I agree with this regarding removing the `UTxO era` parameter. cc: @carbolymer

Jimbo4350:
>  just pointing out that the choice was made to have these supplemental datums be handled within each TxOut instead of grouped outside.

I see your point but the `TxOut` type was designed this way so that consumers of cardano-api could easily differentiate between the different datums that exist at outputs.

I wouldn't include the "resolved datums'  in the `TxOut` definition because it is involved with the creation of datums/datum hashes (or not) at outputs. The "resolved datums" feature is about including already existing datums in the datum map.

cardano-api/test/cardano-api-test/Test/Cardano/Api/Transaction/Autobalance.hs
carbolymer: Removed `utxos` parameter.

cardano-api/test/cardano-api-test/Test/Cardano/Api/Transaction/Autobalance.hs
carbolymer: Removed `utxos` parameter.

cardano-api/test/cardano-api-test/Test/Cardano/Api/Transaction/Autobalance.hs
carbolymer: Removed `utxos` argument.

cardano-api/src/Cardano/Api/Internal/Tx/Body.hs
```diff
@@ -1075,25 +1095,35 @@ addTxInCollateral
...
+      TxInsReference era xs bScriptData' -> TxInsReference era (xs <> txInsReference) ((<> scriptData) <$> bScriptData')
 
 addTxInReference
-  :: IsBabbageBasedEra era => TxIn -> TxBodyContent build era -> TxBodyContent build era
-addTxInReference txInReference = addTxInsReference [txInReference]
+  :: Applicative (BuildTxWith build)
+  => IsBabbageBasedEra era
+  => TxIn
+  -> TxBodyContent build era
+  -> TxBodyContent build era
```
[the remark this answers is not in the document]
carbolymer: Good point, added, thanks.

PR #816, opened 2025-04-16
Improve autobalancing errors when change has no lovelace

carbolymer:
# Changelog

```yaml
- description: |
    Improve autobalancing errors when change has no lovelace
    `checkMinUTxOValue` has its arguments flipped.
  type:
```

# Context

Previously for txout with 0 lovelace and some non-ada assets the returned error was about negative balance, which doesn't make sense.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/src/Cardano/Api/Internal/Fees.hs
+++ b/cardano-api/src/Cardano/Api/Internal/Fees.hs
@@ -645,13 +645,16 @@ estimateBalancedTxBody
         balance =
           evaluateTransactionBalance sbe pparams poolids stakeDelegDeposits drepDelegDeposits fakeUTxO txbody2
         balanceTxOut = TxOut changeaddr balance TxOutDatumNone ReferenceScriptNone
-    -- check if the balance is positive or negative
-    -- in one case we can produce change, in the other the inputs are insufficient
-    first TxFeeEstimationBalanceError $ balanceCheck sbe pparams balanceTxOut
 
     -- Step 6. Check all txouts have the min required UTxO value
     forM_ (txOuts txbodycontent1) $
-      \txout -> first TxFeeEstimationBalanceError $ checkMinUTxOValue sbe txout pparams
+      \txout -> first TxFeeEstimationBalanceError $ checkMinUTxOValue sbe pparams txout
+
+    -- check if the balance is positive or negative
+    -- in one case we can produce change, in the other the inputs are insufficient
+    finalTxOuts <-
+      first TxFeeEstimationBalanceError $
+        checkAndIncludeChange sbe pparams balanceTxOut (txOuts txbodycontent1)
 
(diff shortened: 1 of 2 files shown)


cardano-api/src/Cardano/Api/Internal/Fees.hs
```diff
@@ -1543,10 +1549,16 @@ balanceCheck sbe bpparams txout@(TxOut _ balance _ _) = do
   let outValue@(L.MaryValue coin multiAsset) = toMaryValue $ txOutValueToValue balance
       isPositiveValue = L.pointwise (>) outValue mempty
   if
-    | L.isZero outValue -> pure () -- empty TxOut
+    | L.isZero outValue -> pure () -- empty TxOut - ok, it's removed at the end
+    | L.isZero coin -> -- no ADA, just non-ADA assets
+        Left $
+          TxBodyErrorAdaBalanceTooSmall
```
Jimbo4350, line 1555: You may want to give  @mkoura a heads up if he is depending on the error output for this failure case.

Jimbo4350 (approved):
Nice :+1:

cardano-api/src/Cardano/Api/Internal/Fees.hs
```diff
@@ -1510,26 +1506,36 @@ makeTransactionBodyAutoBalance
...
+  -> [TxOut CtxTx era]
+  -> Either (TxBodyErrorAutoBalance era) [TxOut CtxTx era]
+checkAndIncludeChange sbe pp change@(TxOut _ balance _ _) rest = do
+  let outValue = toMaryValue $ txOutValueToValue balance
+  if L.isZero outValue
+    then pure rest
+    else do
+      -- check if the balance is positive or negative
+      -- in one case we can produce change, in the other the inputs are insufficient
+      balanceCheck sbe pp change
```
palas, line 1527: I would put this check outside the `if`, even though I understand it is equivalent (since the `balanceCheck` already has an exception for this). But the code would be slightly simpler

palas (approved):
I have limited understanding of this code. But I am pretty confident it is all refactoring except for the addition of the `TxBodyErrorAdaBalanceTooSmall` error, which I don't understand why it is there, but I imagine it is correct.

I just added a suggestion that I think can improve readability a little bit

cardano-api/src/Cardano/Api/Internal/Fees.hs
carbolymer: Good idea. I think I'll do a single check for the txout emptiness in the balance check only i.e. I'll change the output type to:
```haskell
... -> Either (TxBodyErrorAutoBalance era) IsEmpty

data IsEmpty = Empty | NonEmpty
```

PR #828, opened 2025-04-24
Remove `BabbageEra` from experimental `Era` type

palas:
# Changelog

```yaml
- description: |
    Removed Babbage era from `Experimental` API together with `babbageEraOnwardsToEra` function.
  type:
  - breaking
```

# Context

This PR addresses https://github.com/IntersectMBO/cardano-api/issues/827. It is part of an effort to propagate the experimental API through `cardano-cli`. Because Babbage era is deprecated we are removing it from the experimental API. 

# How to trust this PR

Check just the right things are removed and the updates make sense. The fact that it typechecks already gives a lot of assurance in this particular case.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-api/src/Cardano/Api/Internal/Experimental/Tx.hs
+++ b/cardano-api/src/Cardano/Api/Internal/Experimental/Tx.hs
@@ -138,7 +138,6 @@ import Cardano.Api.Internal.Tx.Sign
 import Cardano.Crypto.Hash qualified as Hash
 import Cardano.Ledger.Alonzo.TxBody qualified as L
 import Cardano.Ledger.Api qualified as L
-import Cardano.Ledger.Babbage qualified as Ledger
 import Cardano.Ledger.Conway qualified as Ledger
 import Cardano.Ledger.Conway.TxBody qualified as L
 import Cardano.Ledger.Core qualified as Ledger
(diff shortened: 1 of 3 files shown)


cardano-api/src/Cardano/Api/Internal/Experimental/Tx.hs
```diff
@@ -207,51 +205,67 @@ makeUnsignedTx era bc = obtainCommonConstraints era $ do
...
-          & L.feeTxBodyL .~ fee
-          & L.vldtTxBodyL . L.invalidBeforeL .~ convValidityLowerBound (txValidityLowerBound bc)
-          & L.vldtTxBodyL . L.invalidHereAfterL .~ convValidityUpperBound sbe (txValidityUpperBound bc)
-          & L.reqSignerHashesTxBodyL .~ setReqSignerHashes
-          & L.scriptIntegrityHashTxBodyL .~ scriptIntegrityHash
-          & L.withdrawalsTxBodyL .~ withdrawals
-          & L.certsTxBodyL .~ certs
-          & L.mintTxBodyL .~ setMint
-          & L.auxDataHashTxBodyL .~ maybe SNothing (SJust . Ledger.hashTxAuxData) txAuxData
+          & L.inputsTxBodyL
```
carbolymer, line 208: Can you undo this formatting change? This is less readable

palas: Definitely, I don't know why vscode did that, it must have the wrong fourmolu version

palas: I've addressed it by amending the commit

carbolymer (approved):

palas:
> Nice, how does this look in cardano-cli?

It just affects two lines, where there was a `case` over the era. Very small changes

PR #829, opened 2025-04-25
Fix autobalancing when there's no change. Add property test.

carbolymer:
# Changelog

```yaml
- description: |
    Fix an autobalancing error needlessly thrown when there is no change in the transaction. Add property test for autobalancing.
  type:
   - test           # fixes/modifies tests
```

# Context

n/a

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/src/Cardano/Api/Internal/Fees.hs
+++ b/cardano-api/src/Cardano/Api/Internal/Fees.hs
@@ -384,7 +384,6 @@ import Cardano.Ledger.Plutus.Language qualified as Plutus
 import Cardano.Ledger.Val qualified as L
 import Ouroboros.Consensus.HardFork.History qualified as Consensus
 
-import Control.Monad
 import Data.Bifunctor (bimap, first, second)
 import Data.Bitraversable (bitraverse)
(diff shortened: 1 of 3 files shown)


cardano-api/src/Cardano/Api/Internal/Fees.hs
```diff
@@ -1371,7 +1371,9 @@ makeTransactionBodyAutoBalance
               TxOutDatumNone
               ReferenceScriptNone
 
-      _ <- balanceCheck sbe pp initialChangeTxOut
+      -- Initial change is only used for execution units evaluation, so we don't require minimum UTXO requirement
+      -- to be satisfied at this point
+      _ <- balanceCheck sbe pp SkipMinUtxoCheck initialChangeTxOut
```
Jimbo4350, line 1376: Instead of modifying `balanceCheck` why not check the balance at the end? This seems hacky.

Jimbo4350 (changes requested):

Jimbo4350 (approved):

Jimbo4350:
Nice :+1:

PR #834, opened 2025-05-02
Add genBlockHeader and related functions to `Test.Gen.Cardano.Api.Typed`

locallycompact:
# Changelog

```yaml
- description: |
    Add `genBlockHeader` and related functions to `Test.Gen.Cardano.Api.Typed`.
  type:
```

# Context

I have ported these directly from hydra. Let me know if you would change anything. https://github.com/cardano-scaling/hydra/issues/1987

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/gen/Test/Gen/Cardano/Api/Typed.hs
+++ b/cardano-api/gen/Test/Gen/Cardano/Api/Typed.hs
@@ -137,6 +137,10 @@ module Test.Gen.Cardano.Api.Typed
   , genMintWitnessable
   , genPlutusScriptWitness
   , genIndexedPlutusScriptWitness
+  , genBlockNo
+  , genBlockHeader
+  , genBlockHeaderAt
+  , genBlockHeaderHash
   )
 where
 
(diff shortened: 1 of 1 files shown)


cardano-api/gen/Test/Gen/Cardano/Api/Typed.hs
```diff
@@ -1468,3 +1473,29 @@ genScriptWitnessForMint sbe = do
...
+genBlockHeaderAt :: SlotNo -> Gen BlockHeader
+genBlockHeaderAt slotNo = BlockHeader slotNo <$> genBlockHeaderHash <*> geBlockNo
+
+-- | Generate a random block header hash.
+genBlockHeaderHash :: Gen (Hash BlockHeader)
+genBlockHeaderHash =
+  unsafeBlockHeaderHashFromBytes . BS.pack <$> Gen.list (Range.singleton 32) Q.arbitrary
+ where
+  unsafeBlockHeaderHashFromBytes bytes =
+    case deserialiseFromRawBytes (proxyToAsType Proxy) bytes of
```
carbolymer, line 1494: Is there a way to make this function non-partial?

ch1bo: It only fails should the hash size of block headers would change (currently `32` bytes).

Ideally, we would have a generator for the [`Hash` type of `cardano-base`](https://github.com/IntersectMBO/cardano-base/blob/master/cardano-crypto-class/src/Cardano/Crypto/Hash/Class.hs#L112) which would limit the unsafe construction to that module (there are a few more partial functions in there)

Jimbo4350 (approved):

cardano-api/gen/Test/Gen/Cardano/Api/Typed.hs
carbolymer: >There were other partials in this file so I didn't consider it.
We aren't proud of that, but the goal is to not add unnecessary partial functions.

>It only fails should the hash size of block headers would change (currently 32 bytes).
Thanks Sebastian for clarifying that.

@locallycompact could you put that information into the comment?

carbolymer (approved):

locallycompact:
I think this is good now.

cardano-api/gen/Test/Gen/Cardano/Api/Typed.hs
```diff
@@ -1488,3 +1492,31 @@ genScriptWitnessForMint sbe = do
...
+genBlockHeader :: Gen BlockHeader
+genBlockHeader = genSlotNo >>= genBlockHeaderAt
+
+-- | Generate a random block header with completely random hash, but at a
+-- certain slot.
+genBlockHeaderAt :: SlotNo -> Gen BlockHeader
+genBlockHeaderAt slotNo = BlockHeader slotNo <$> genBlockHeaderHash <*> genBlockNo
+
+-- | Generate a random block header hash.
+-- | This will error if the hash size of block headers changes (currently 32 bytes).
```
carbolymer, line 1509:
```suggestion
-- This will error if the hash size of block headers changes (currently 32 bytes).
```

locallycompact:
Signed

PR #842, opened 2025-05-15
Remove partial instances

carbolymer:
# Changelog

```yaml
- description: |
    Remove partial `IsString` instances. Use `deserialiseFromRawBytesHex` or specialised parser instead.
    Make `SerialiseAsBech32` class use `HumanReadablePart` instead of `Text` for Bech32 prefix in `bech32PrefixFor` and `bech32PrefixesPermitted` functions.
  type:
```

# Context

`IsString` instances generated from `UsingRawBytesHex` were partial and could fail with runtime exception leading to hard to track errors. We were using `fromString` on types using those instances in `cardano-cli` and `cardano-testnet` already. This PR removes those and adds `parsec` parsers instead.

Because `IsString` was used in tests, it's replaced by specialised functions in tests only for example:
```haskell
mkCredential :: HasCallStack => Text -> L.Credential k
mkCredential = errorFail @String . L.parseCredential

mkTxIn :: HasCallStack => Text -> TxIn
mkTxIn = either error id . P.runParser parseTxIn
```
[ledger's `parseCredential`](https://github.com/IntersectMBO/cardano-ledger/blob/607a7fdad352eb72041bb79f37bc1cf389432b1d/libs/cardano-ledger-core/src/Cardano/Ledger/Credential.hs#L143)

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/cardano-api.cabal
+++ b/cardano-api/cardano-api.cabal
@@ -100,6 +100,7 @@ library
     Cardano.Api.Ledger
     Cardano.Api.Ledger.Lens
     Cardano.Api.Network
+    Cardano.Api.Parser.Text
     Cardano.Api.Shelley
     Cardano.Api.Tx.UTxO
 
@@ -322,6 +323,7 @@ test-suite cardano-api-test
   main-is: cardano-api-test.hs
   type: exitcode-stdio-1.0
   build-depends:
+    FailT,
     QuickCheck,
     aeson >=1.5.6.0,
     base16-bytestring,
@@ -420,7 +422,6 @@ test-suite cardano-api-golden
     hedgehog >=1.1,
     hedgehog-extras ^>=0.7,
     microlens,
-    parsec,
     plutus-core ^>=1.45,
     plutus-ledger-api,
     tasty,
--- a/cardano-api/src/Cardano/Api/Internal/Keys/Shelley.hs
+++ b/cardano-api/src/Cardano/Api/Internal/Keys/Shelley.hs
@@ -45,6 +45,7 @@ module Cardano.Api.Internal.Keys.Shelley
   , anyStakePoolVerificationKeyHash
   , AnyStakePoolSigningKey (..)
   , anyStakePoolSigningKeyToVerificationKey
+  , parseHexHash
   )
 where
 
@@ -59,6 +60,7 @@ import Cardano.Api.Internal.SerialiseJSON
 import Cardano.Api.Internal.SerialiseRaw
 import Cardano.Api.Internal.SerialiseTextEnvelope
 import Cardano.Api.Internal.SerialiseUsing
+import Cardano.Api.Parser.Text qualified as P
 
 import Cardano.Crypto.DSIGN qualified as DSIGN
 import Cardano.Crypto.DSIGN.Class qualified as Crypto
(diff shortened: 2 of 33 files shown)


newhoggy (approved):

cardano-api/src/Cardano/Api/Internal/SerialiseBech32.hs
```diff
@@ -29,29 +30,20 @@ import Data.List qualified as List
 import Data.Set (Set)
 import Data.Text (Text)
 import GHC.Exts (IsList (..))
+import GHC.Stack
 
 class (HasTypeProxy a, SerialiseAsRawBytes a) => SerialiseAsBech32 a where
   -- | The human readable prefix to use when encoding this value to Bech32.
-  bech32PrefixFor :: a -> Text
+  bech32PrefixFor :: a -> Bech32.HumanReadablePart
```
Jimbo4350, line 38: Be sure to include this breaking change in the PR description so it is obvious when generating the change logs in the next release.

cardano-api/cardano-api.cabal
```diff
@@ -420,7 +422,6 @@ test-suite cardano-api-golden
     hedgehog >=1.1,
     hedgehog-extras ^>=0.7,
     microlens,
```
Jimbo4350, line 424: @palas at some point in the future we should do an audit of our dependencies in order to reduce the size of our wasm library.

cardano-api/src/Cardano/Api/Internal/Keys/Shelley.hs
```diff
@@ -2233,3 +2235,7 @@ instance CastVerificationKeyRole DRepExtendedKey DRepKey where
    where
     impossible =
       error "castVerificationKey (DRep): byron and shelley key sizes do not match!"
+
+-- | Parse hex representation of any 'Hash'
+parseHexHash :: P.Parser (Hash PaymentKey)
```
Jimbo4350, line 2240: `parseHexPaymentKeyHash`?  Many different things can be hex encoded.

cardano-api/src/Cardano/Api/Internal/Script.hs
```diff
@@ -959,6 +958,24 @@ instance SerialiseAsRawBytes ScriptHash where
...
+    fail $
+      mconcat
+        [ "Error while parsing ScriptHash: Expected 56-hex-digit hash, but found "
+        , show (length hexStr)
+        , " hex digits. Supplied hash: "
+        , hexStr
+        ]
+  parseRawBytesHex
+ where
+  -- ADDRHASH = Blake2b_224
```
Jimbo4350, line 975: :+1:

cardano-api/src/Cardano/Api/Internal/SerialiseBech32.hs
```diff
@@ -3,6 +3,7 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
```
Jimbo4350, line 6: You can remove this pragma.

Jimbo4350 (approved):

cardano-api/cardano-api.cabal
carbolymer: But on the note of putting cardano-api on a diet - we have both `parsec` and `attoparsec` in dependencies. I'm not sure if both are needed.

cardano-api/test/cardano-api-test/Test/Cardano/Api/Transaction/Autobalance.hs
```diff
@@ -574,7 +576,7 @@ prop_ensure_gov_actions_are_preserved_by_autobalance = H.propertyOnce $ do
           , L.pProcReturnAddr =
               L.RewardAccount
                 { L.raNetwork = L.Testnet
-                , L.raCredential = L.KeyHashObj "0b1b872f7953bccfc4245f3282b3363f3d19e9e001a5c41e307363d7"
+                , L.raCredential = mkCredential "keyHash-0b1b872f7953bccfc4245f3282b3363f3d19e9e001a5c41e307363d7"
```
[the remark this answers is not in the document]
carbolymer: Added.

PR #852, opened 2025-05-28
Add support for `wasm`

palas:
# Changelog

```yaml
- description: |
    Added support for compiling `cardano-api` to `wasm`
  type:
  - feature
```

# Context

See this ADR: https://github.com/input-output-hk/cardano-node-wiki/pull/75

# How to trust this PR

I think the best way is testing it and ensuring the code makes sense. We'll work on adding extensive testing later.

# Checklist

- [x] Self-reviewed the diff


cardano-wasm/app/Bridge.hs
```diff
@@ -0,0 +1,148 @@
...
+
+-- Pure Haskell JS conversion functions
+
+fromJSBigInt :: JSVal -> IO Integer
+fromJSBigInt val = do
+  jsString <- js_toString val
+  let str = fromJSString jsString
+  case reads str of
+    [(n, "")] -> return n
+    _ -> error ("Wrong format for argument when deserialising, expected integer: " ++ show str)
```
carbolymer, line 49: Not so pure if it errors out I guess? 😄 If this throws an error, maybe it should have `HasCallStack` in the function signature?

cardano-wasm/app/Bridge.hs
```diff
@@ -0,0 +1,148 @@
...
+jsValToType :: Api.HasTextEnvelope a => Api.AsType a -> JSVal -> IO a
+jsValToType asType val = do
+  envelope <- jsValToJSON val
+  case Api.deserialiseFromTextEnvelope asType envelope of
+    Left err -> error ("Error deserialising envelope in parameter: " ++ show err)
+    Right type_ -> return type_
+
+-- Conversion functions between Haskell and JS
+
+class ToJSVal haskellType jsType where
```
carbolymer, line 76: Haddock por favor

cardano-wasm/app/Bridge.o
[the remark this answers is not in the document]
palas: Oops

carbolymer (changes requested):

.gitignore
```diff
@@ -45,7 +48,7 @@ tags
 ./*.cert
 
 # For now require that users generate their own hie.yaml
-hie.yaml
+./hie.yaml
```
newhoggy, line 51: Maybe just `/hie.yaml`?  can substitute the `./` with `/` for the keys above too I think.

.github/workflows/wasm.yml
```diff
@@ -0,0 +1,174 @@
...
+      run: |
+        cd cardano-wasm
+        wasm32-wasi-cabal build all --no-semaphore -j1 --ghc-options="-j1"
+
+    # - name: Run tests
+    #   env:
+    #     TMPDIR: ${{ runner.temp }}
+    #     TMP: ${{ runner.temp }}
+    #     KEEP_WORKSPACE: 1
+    #   run: cabal test all --enable-tests --test-show-details=direct
```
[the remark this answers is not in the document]
palas: There aren't any so far, and because the tests would run in `wasm`, I don't know if they would work at all from `cabal`, they would need to be run with a wasm interpreter instead (maybe `cabal` does that automatically, I would have to check). But also, what we should be testing is the resulting JS, I think, and that will probably require setting up Playwright or Node.js.

.gitignore
```diff
@@ -19,6 +19,9 @@ dist/
 result*
 /launch-*
 stack.yaml.lock
+*.hi
+*.o
+Main
```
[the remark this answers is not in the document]
palas: Because of this: https://github.com/IntersectMBO/cardano-api/pull/852#discussion_r2121653689
I included them by mistake when I was doing some manual testing. So this will prevent future mistakes

cardano-wasm/app/Bridge.hs
```diff
@@ -0,0 +1,148 @@
...
+  js_stringify :: JSVal -> IO JSString
+
+foreign import javascript unsafe "($1).toString()"
+  js_toString :: JSVal -> IO JSString
+
+-- Pure Haskell JS conversion functions
+
+fromJSBigInt :: JSVal -> IO Integer
+fromJSBigInt val = do
+  jsString <- js_toString val
```
newhoggy, line 45: Is this assuming that `val` is a string?  Is the caller expected to only supply a `JSVal` that is a string?

.github/workflows/wasm.yml
```diff
@@ -0,0 +1,174 @@
...
+
+    # Now we install the dependencies. If the cache was found and restored in the previous step,
+    # this should be a no-op, but if the cache key was not found we need to build stuff so we can
+    # cache it for the next step.
+    - name: Install dependencies
+      run: |
+        cd cardano-wasm
+        wasm32-wasi-cabal build all --only-dependencies --no-semaphore -j1 --ghc-options="-j1"
+
+    # Always store the cabal cache.
```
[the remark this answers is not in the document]
palas: Ah, maybe I should add `if: always()`, cause it takes quite long to build. I didn't know what that was for, but I just found out

cardano-wasm/cabal.project
```diff
@@ -0,0 +1,93 @@
+active-repositories: hackage.haskell.org, cardano-haskell-packages
+packages: .
```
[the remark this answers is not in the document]
newhoggy: It would be weird if we used a version of `cardano-api` that is different to the one checked out.

cardano-wasm/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+```console
+$(wasm32-wasi-ghc --print-libdir)/post-link.mjs -i "$(env -u CABAL_CONFIG wasm32-wasi-cabal list-bin exe:cardano-wasm | tail -n1)" -o cardano-wasm.js
+```
+
+That will create it with the name `cardano-wasm.js` in the current folder.
+
+You can find more information in [this url](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/wasm.html).
+
+And you can find an example of how to use it in the `example` subfolder. This example assumes that the generated `.wasm` and `.js` files reside in the same folder as the code in `example` subfolder.
```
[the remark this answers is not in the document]
palas: Thanks for trying it, I am glad someone apart from me got it to work. Maybe I should add those gotchas to the explanation

cardano-wasm/app/Bridge.hs
```diff
@@ -0,0 +1,131 @@
...
+    let txBodyContent =
+          Api.defaultTxBodyContent sbe
+            & Api.setTxIns [txIn]
+            & Api.setTxOuts [txOut]
+            & Api.setTxFee txFee
+    let (Right txBody) = Api.createTransactionBody sbe txBodyContent
+    return txBody
+
+foreign export javascript "signTransaction"
+  signTransaction :: JSVal -> JSString -> IO JSVal
```
Jimbo4350, line 140: It is possible to use type synonyms? We should also write haddocks from now.

cardano-wasm/cabal.project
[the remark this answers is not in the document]
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/852/commits/468e1c3a3e45fb1f7a49083aa1895725e4604fec

cardano-wasm/app/Bridge.hs
palas: > Do you have example behaviour of what happens when error is called?

Goode idea, it seems it already has an implicit `HasCallStack`, if I insert an error in Haskell in the line 60 of `Bridge.hs`, I get:
```console
Unhandled Promise Rejection: RuntimeError: Test error!
CallStack (from HasCallStack):
  error, called at app/Bridge.hs:60:3 in cardano-wasm-0.1.0.0-inplace-cardano-wasm:Bridge
```

.github/workflows/wasm.yml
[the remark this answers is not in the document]
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/852/commits/a624e406e0fdf89c56ba9baf150374035d050047

.gitignore
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/852/commits/e851108005ebdeca7387cff6265049bbd13b0f7c

cardano-wasm/app/Bridge.hs
```diff
@@ -0,0 +1,148 @@
...
+
+instance FromJSVal JSVal (Api.TxBody Api.ConwayEra) where
+  fromJSVal = jsValToType (Api.AsTxBody Api.AsConwayEra)
+
+instance FromJSVal JSVal (Api.Tx Api.ConwayEra) where
+  fromJSVal = jsValToType (Api.AsTx Api.AsConwayEra)
+
+instance FromJSVal JSString (Api.SigningKey Api.PaymentKey) where
+  fromJSVal jsString = do
+    let (Right signingKey) = Api.deserialiseFromBech32 (Api.AsSigningKey Api.AsPaymentKey) (Text.pack (fromJSString jsString))
```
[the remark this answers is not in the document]
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/852/commits/8cad356b2468e546bea6f3a981b39b22166f79c4

cardano-wasm/app/WasmApi.hs
```diff
@@ -0,0 +1,41 @@
...
+          Api.TxOutDatumNone
+          Script.ReferenceScriptNone
+      txFee = Api.TxFeeExplicit sbe (fromInteger fees)
+
+      txBodyContent =
+        Api.defaultTxBodyContent sbe
+          & Api.setTxIns [txIn]
+          & Api.setTxOuts [txOut]
+          & Api.setTxFee txFee
+      (Right txBody) = Api.createTransactionBody sbe txBodyContent
```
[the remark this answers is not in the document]
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/852/commits/8cad356b2468e546bea6f3a981b39b22166f79c4

cardano-wasm/app/Bridge.hs
palas: do you mean: `fromJSBigIntTestable :: String -> Either SomeError Integer`?

cardano-wasm/app/Bridge.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/852/commits/92745e83653282786694adfec786751ee9c18a1f

cardano-wasm/README.md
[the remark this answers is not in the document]
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/852/commits/4f7a6326611865473d361465ce8bd6827c7da81d

flake.nix
```diff
@@ -169,12 +173,39 @@
             inherit hydraJobs;
           };
           devShells = let
+            # wasm shell
+            wasmShell = let wasm-pkgs = inputs.wasm-nixpkgs.legacyPackages.${system};
```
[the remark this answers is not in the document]
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/852/commits/d52dc024fffa72a952cfc07d2a6e6aca34fab40d

cardano-wasm/app/Bridge.hs
carbolymer: >Good idea, it seems it already has an implicit HasCallStack, 

No it does not. You only get the location of `error` call, not where the `fromJSBigInt` was called, which is useful as well.

cardano-wasm/cabal.project
```diff
@@ -0,0 +1,97 @@
...
```
carbolymer, line 2: Can we have one main `cabal.project` in the repo? You could have this one [imported](https://cabal.readthedocs.io/en/stable/cabal-project-description-file.html#conditionals-and-imports) in the top level one. This would save us the hassle of keeping track of two index states: here and in the top level `cabal.project`.

cardano-wasm/cardano-wasm.cabal
```diff
@@ -0,0 +1,67 @@
...
+    -Wincomplete-uni-patterns
+    -Wno-unticked-promoted-constructors
+    -Wpartial-fields
+    -Wredundant-constraints
+    -Wunused-packages
+
+executable cardano-wasm
+  main-is: Main.hs
+  hs-source-dirs:
+    app
```
carbolymer, line 30: what's the purpose of `app` directory?

cardano-wasm/src/WasmApi/Tx.hs
```diff
@@ -0,0 +1,177 @@
...
+data SignedTxObject
+  = forall era. SignedTxObject (Exp.Era era) (Ledger.Tx (Exp.LedgerEra era))
+  deriving Typeable
+
+instance ToJSON SignedTxObject where
+  toJSON :: HasCallStack => SignedTxObject -> Aeson.Value
+  toJSON (SignedTxObject era ledgerTx) =
+    obtainCommonConstraints era $
+      let encode :: forall a. EncCBOR a => a -> Text.Text
+          encode = Text.decodeUtf8 . Base16.encode . Ledger.serialize' (eraToVersion era)
```
carbolymer, line 153: Same remark as for `UnsignedTxObject`

cardano-wasm/src/JavaScript/Bridge.hs
```diff
@@ -0,0 +1,360 @@
...
+
+-- * Type Synonyms for JSVal representations
+
+type JSAPIInfo = JSVal
+
+type JSUnsignedTx = JSVal
+
+type JSSignedTx = JSVal
+
+type JSTxId = JSSigningKey
```
Jimbo4350, line 161: I don't understand this type synonym.

cardano-wasm/src/WasmApi/Tx.hs
```diff
@@ -0,0 +1,177 @@
...
+data UnsignedTxObject
+  = forall era. UnsignedTxObject (Exp.Era era) [Ledger.WitVKey Ledger.Witness] (Exp.UnsignedTx era)
+  deriving Typeable
+
+instance ToJSON UnsignedTxObject where
+  toJSON :: HasCallStack => UnsignedTxObject -> Aeson.Value
+  toJSON (UnsignedTxObject era keyWitnesess (Exp.UnsignedTx tx)) =
+    obtainCommonConstraints era $
+      let encode :: forall a. EncCBOR a => a -> Text.Text
+          encode = Text.decodeUtf8 . Base16.encode . Ledger.serialize' (eraToVersion era)
```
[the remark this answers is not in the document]
Jimbo4350: We would need to confirm what the web wallet api expects and use that encoding.

Jimbo4350 (changes requested):
I would break this up into at least 3 PRs: 
- A proof of concept wasm module. 
- Expansion of the api.
- The js/html example


 I also agree with @carbolymer 's comments.

cardano-wasm/cardano-wasm.cabal
```diff
@@ -0,0 +1,67 @@
...
+  if arch(wasm32)
+    ghc-options:
+      -no-hs-main
+      -optl-mexec-model=reactor
+      "-optl-Wl,--strip-all,--export=getAPIInfo,--export=newConwayTx,--export=addTxInput,--export=addSimpleTxOut,--export=setFee,--export=addSigningKey,--export=signTx,--export=txToCbor"
+  other-modules:
+    General.ExceptionHandling
+    JavaScript.Bridge
+    WasmApi.Tx
+    WasmApi.WasmApi
```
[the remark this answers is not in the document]
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/852/commits/552c133c57d4c8a4bf3a7bd3886b6819a70a2dcf

cardano-wasm/src/WasmApi/Tx.hs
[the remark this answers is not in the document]
palas: The wallet does use `base16`. And in this particular bit, it is not important, because the `UnsignedTx` object is opaque. So we could use `base64` just for the opaque things, but then we would be increasing the size of the bundle, so I am not sure it is worth it, and either way the overhead is probably very little in the scale of things

cardano-wasm/src/JavaScript/Bridge.hs
palas: It is a mistake, nice catch. Fixed here: https://github.com/IntersectMBO/cardano-api/commit/c763cf6c045459211a62b787fbcec412c27a5607

cardano-wasm/src/WasmApi/Tx.hs
palas: See: https://github.com/IntersectMBO/cardano-api/pull/852#discussion_r2162071201

cardano-wasm/src/JavaScript/Bridge.hs
```diff
@@ -0,0 +1,360 @@
...
+  fromJSVal :: JSString -> IO String
+  fromJSVal = return . fromJSString
+
+instance (Api.FromJSON a, Typeable a) => FromJSVal JSVal a where
+  fromJSVal = jsValToJSON (show . typeRep $ (Api.Proxy :: Api.Proxy a))
+
+instance FromJSVal JSSigningKey (Api.SigningKey Api.PaymentKey) where
+  fromJSVal :: HasCallStack => JSSigningKey -> IO (Api.SigningKey Api.PaymentKey)
+  fromJSVal jsString = do
+    return $ rightOrError $ Api.deserialiseFromBech32 (Text.pack (fromJSString jsString))
```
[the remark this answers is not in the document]
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/commit/950b265ea6557c7906bf432726533056237a013f

cardano-wasm/README.md
```diff
@@ -0,0 +1,236 @@
+# cardano-wasm
+
+Part of an effort at IOG (@Jimbo4350, @palas, and @amesgen) to build Cardano Haskell libraries to Wasm.
```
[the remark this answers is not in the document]
palas: Haha, I was just making sure I gave credit, I will remove the parenthesis 😅

.github/workflows/haskell-wasm.yml
```diff
@@ -0,0 +1,176 @@
...
+        cat <<EOF >> $GITHUB_ENV
+        MONTHNUM=$(date -u '+%m')
+        GHC=$(ghc --numeric-version)
+        STORE=$(wasm32-wasi-cabal path --store | tail -n 1)
+        EOF
+
+    # From the dependency list we restore the cached dependencies.
+    # We use the hash of `dependencies.txt` as part of the cache key because that will be stable
+    # until the `index-state` values in the `cabal.project` file changes.
+#    - name: Restore cached dependencies
```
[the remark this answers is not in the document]
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/commit/359d683ba3c288bb5c7380679c1705cddb1fcffa

.github/workflows/haskell-wasm.yml
```diff
@@ -0,0 +1,183 @@
...
+        GHC=$(ghc --numeric-version)
+        STORE=$(wasm32-wasi-cabal path --store | tail -n 1)
+        EOF
+
+# Cache is disabled because GHA default builders are not able to build all dependencies
+# because they lack RAM, so having the cache expire would break the CI check.
+# For this reason, we are providing a build of the dependencies instead in
+# the "Restore cached deps" step, and we make the check not required.
+# When we are able to make this CI check self-sufficient, we should reenable the
+# caching and remove the manual restoring of cached deps.
```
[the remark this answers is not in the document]
carbolymer: I think this is missing a lot https://ci.iog.io/build/8189984#tabs-buildsteps

I think we should try bugging devx for help. Maybe building wasm with hydra could populate IOG cache.

cardano-wasm/README.md
```diff
@@ -0,0 +1,236 @@
...
+```
+
+Then we need to compile three libraries to WASM: `libblst`, `libsodium`, and `libsecp256k1`.
+
+In order to not interfere with the system library installation, we will create a folder to serve as our prefix:
+
+```console
+mkdir ~/prefix
+mkdir ~/prefix/{lib,include}
+mkdir ~/prefix/lib/pkgconfig
```
carbolymer, line 50:
```suggestion
mkdir -p ~/prefix/{lib/pkgconfig,include}
```

cardano-wasm/example/index.html
carbolymer, line 1: Definitely out of scope of this PR: Maybe we should publish a live demo on gitub pages for this?

cardano-wasm/src/Cardano/Wasm/Internal/Api/Tx.hs
```diff
@@ -0,0 +1,172 @@
...
+-- It is meant to be an opaque object in JavaScript API.
+data UnsignedTxObject
+  = forall era. UnsignedTxObject (Exp.Era era) [Ledger.WitVKey Ledger.Witness] (Exp.UnsignedTx era)
+
+instance ToJSON UnsignedTxObject where
+  toJSON :: UnsignedTxObject -> Aeson.Value
+  toJSON (UnsignedTxObject era keyWitnesess (Exp.UnsignedTx tx)) =
+    obtainCommonConstraints era $
+      let encode :: forall a. EncCBOR a => a -> Text.Text
+          encode = Text.decodeUtf8 . Base16.encode . Ledger.serialize' (eraToVersion era)
```
carbolymer, line 56: https://github.com/IntersectMBO/cardano-api/pull/852#discussion_r2152436608

Sorry I wasn't specific enough. My point was not about particular encoding but to use already existing instances instead of encoding the value by hand.

I mean:
1. Create `instance SerialiseAsRawBytes (Ledger.Tx (LedgerEra era)) where ...`
2. Write `deriving newtype instance SerialiseAsRawBytes (Exp.UnsignedTx era)`
3. Write
  ```haskell
  deriving (ToJSON, FromJSON) via (UsingRawBytesHex (Exp.UnsignedTx era))
  ```
4. use `toJSON tx` instead of `encode tx` below.

From what I understand we're sticking to base16 representation 👍🏻

cardano-wasm/src/Cardano/Wasm/Internal/Api/Tx.hs
```diff
@@ -0,0 +1,172 @@
...
+        tx' = tx & Ledger.bodyTxL . Ledger.outputsTxBodyL %~ (<> StrictSeq.fromList [shelleyTxOut])
+     in UnsignedTxObject era keyWitnesess $ Exp.UnsignedTx tx'
+ where
+  deserialiseAddress
+    :: HasCallStack
+    => Exp.EraCommonConstraints era
+    => Exp.Era era -> String -> Api.AddressInEra era
+  deserialiseAddress _eon destAddrStr =
+    justOrError
+      "Couldn't deserialise destination address"
```
carbolymer, line 112: Why `error` instead of `MonadThrow` or `Either e`?

cardano-wasm/src/Cardano/Wasm/Internal/Api/Tx.hs
```diff
@@ -0,0 +1,172 @@
...
+-- | An object representing a signed transaction.
+data SignedTxObject
+  = forall era. SignedTxObject (Exp.Era era) (Ledger.Tx (Exp.LedgerEra era))
+
+instance ToJSON SignedTxObject where
+  toJSON :: SignedTxObject -> Aeson.Value
+  toJSON (SignedTxObject era ledgerTx) =
+    obtainCommonConstraints era $
+      let encode :: forall a. EncCBOR a => a -> Text.Text
+          encode = Text.decodeUtf8 . Base16.encode . Ledger.serialize' (eraToVersion era)
```
carbolymer, line 148: Same remark as in `UnsignedTxObject`

cardano-wasm/src/Cardano/Wasm/Internal/Api/Tx.hs
```diff
@@ -0,0 +1,172 @@
...
+            , "tx" .= encode tx
+            ]
+
+instance FromJSON UnsignedTxObject where
+  parseJSON :: HasCallStack => Aeson.Value -> Aeson.Parser UnsignedTxObject
+  parseJSON = Aeson.withObject "UnsignedTxObject" $ \o -> do
+    Exp.Some era <- o Aeson..: "era"
+    let decode :: forall a. DecCBOR (Annotator a) => Text.Text -> Text.Text -> a
+        decode desc cbor = do
+          let cddlBS = rightOrError $ Base16.decode $ Text.encodeUtf8 cbor
```
carbolymer, line 69: That should be a parse failure and not an exception.

carbolymer (changes requested):
My main gripe is throwing exceptions from pure code.

.github/workflows/haskell-wasm.yml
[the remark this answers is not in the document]
palas: Yes, the nix shell is only caching the build tools. I don't know how to build it for wasm using `haskell.nix`, but that would probably solve the issue. There is an effort towards achieving that, but I don't think they got there yet. I can ask devx more directly, and see what they say

cardano-wasm/README.md
palas: Applied here: https://github.com/IntersectMBO/cardano-api/pull/852/commits/19e3351e309a7fc35b6570dcab0074f0f9f9d365

cardano-wasm/src/Cardano/Wasm/Internal/Api/Info.hs
```diff
@@ -0,0 +1,134 @@
...
+instance Aeson.ToJSON VirtualObjectInfo where
+  toJSON :: VirtualObjectInfo -> Aeson.Value
+  toJSON (VirtualObjectInfo name methods) =
+    Aeson.object
+      [ "objectName" Aeson..= name
+      , "methods" Aeson..= methods
+      ]
+
+-- | Aggregate type for all API information.
+data APIInfo = APIInfo
```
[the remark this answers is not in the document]
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/852/commits/07b7146b6f76e611a080b73f746a950aa9bf89a9

cardano-wasm/src/Cardano/Wasm/Internal/Api/Tx.hs
palas: I managed to make a `SerialiseAsRawBytes (UnsignedTx ConwayEra)` instance, but I can't use it for automatic `deriving` of `UnsignedTxObject` because `UnsignedTxObject` has existentials, and deriving doesn't seem to like that.

I could still use `SerialiseAsRawBytes` in the manual instantiation, but I also need to serialise `Ledger.WitVKet Ledger.Witness`. And I could also do an orphan `SerialiseAsRawBytes` instance for `Ledger.WitVKet Ledger.Witness` but:
 - The whole getting that to work seems pretty trick to begin with and may complicate `UnsignedTxObject` by requiring extra constraints.
 - I am already using a standard instance for both (`EncCBOR`), which is a bit ugly to use but it wouldn't be that much difference fundamentaly.
 - That wouldn't really be ergonomic, because I would have to also define `AsType` for that type and it is a ledger type, so it is awkward.

So, I will give up for now on deriving unless you have any ideas to get around the issue.

I added the `SerialiseAsRawBytes` instance to `UnsignedTx ConwayEra` in this commit: https://github.com/IntersectMBO/cardano-api/commit/cdc8498e1a94d16f9de5da3662973d2a794b4712

cardano-wasm/src/Cardano/Wasm/Internal/Api/Tx.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/852/commits/39631d06c48e7e61b9106882321da9702ce3131a

cardano-wasm/src/Cardano/Wasm/Internal/Api/Tx.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/852/commits/39631d06c48e7e61b9106882321da9702ce3131a

cardano-wasm/src/Cardano/Wasm/Internal/Api/Tx.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/852/commits/39631d06c48e7e61b9106882321da9702ce3131a

palas:
Back-up before rebasing/squashing

cardano-wasm/src/Cardano/Wasm/Internal/Api/Tx.hs
```diff
@@ -0,0 +1,182 @@
...
+  case era of
+    Exp.ConwayEra -> Ledger.eraProtVerHigh @(Exp.LedgerEra Exp.ConwayEra)
+
+-- * @UnsignedTx@ object
+
+-- | An object representing a transaction that is being built and hasn't
+-- been signed yet. It abstracts over the era of the transaction.
+-- It is meant to be an opaque object in JavaScript API.
+data UnsignedTxObject
+  = forall era. UnsignedTxObject (Exp.Era era) [Ledger.WitVKey Ledger.Witness] (Exp.UnsignedTx era)
```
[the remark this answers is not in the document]
palas: The advantage of having the `[Ledger.WitVKey Ledger.Witness]` accumulated instead of looking them up in the transaction is that the transaction doesn't tell you which kind of witness they are once you add them (maybe there is a way to find out and I don't know it). But if we add them to the wrapper then we know they must be shelley witnesses because that is the only way to add them there. So that was my thinking

cardano-wasm/src/Cardano/Wasm/Internal/Api/Tx.hs
```diff
@@ -0,0 +1,182 @@
...
+-- | Add a payment key to the list of signatories for an unsigned transaction.
+addSigningKeyImpl :: UnsignedTxObject -> Api.SigningKey Api.PaymentKey -> UnsignedTxObject
+addSigningKeyImpl (UnsignedTxObject era keyWitnesess unsignedTx) signingKey =
+  obtainCommonConstraints era $
+    let witness = Api.WitnessPaymentKey signingKey
+        keyWitness = Exp.makeKeyWitness era unsignedTx witness
+     in UnsignedTxObject era (keyWitnesess ++ [keyWitness]) unsignedTx
+
+-- | Sign an unsigned transaction using the list of signatories.
+signTxImpl :: UnsignedTxObject -> SignedTxObject
```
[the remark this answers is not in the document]
palas: The idea is that you can add the witnesses one at a time, so the API is simpler, and we can add functionalities in the future (like `addByronKeyWitness`, `addScriptWitness`, ...), without breaking the API.

The alternative is to have a function that takes a changing type (like you suggest), made of a list of a complex structured types (the witness object), but I don't think that is a good idea for JavaScript, because types are not checked, so the simpler the parameters the better. And why construct an array of witnesses if you can just add them directly?

On the other hand, it is necessary to apply all witnesses at the same time, because otherwise they get invalidated, so that is why I suggest we just accumulate them and apply them at once when you call `signTx`. 

This also allows the `calculateMinFee` function to count the witnesses for you, instead of you having to provide a number (in most cases).

cardano-wasm/src/Cardano/Wasm/Internal/Api/Tx.hs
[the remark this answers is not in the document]
palas: I see my mistake now, it is not the `WitVKey Witness` that I have to accumulate, it is the `SigningKey PaymentKey`. Addressed here: https://github.com/IntersectMBO/cardano-api/pull/852/commits/c68df85b3227e3a2d3eb4ae544fb22cfaaa58f8d

cardano-wasm/cardano-wasm.cabal
```diff
@@ -0,0 +1,43 @@
+cabal-version: 3.4
+name: cardano-wasm
+version: 0.1.0.0
```
Jimbo4350, line 3: This should be `10.0.0.0` so it is in lockstep with the other package versions.

Jimbo4350 (approved):

cardano-wasm/cardano-wasm.cabal
palas: Addressed by amending the commit

flake.nix
```diff
@@ -149,15 +153,42 @@
             });
           }
         );
+        # wasm shell
+        wasmShell = let wasm-pkgs = inputs.wasm-nixpkgs.legacyPackages.${system};
```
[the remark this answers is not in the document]
palas: Addressed in amended commit

PR #851, opened 2025-05-27
Update TxInsReference haddock to mention public key inputs

carbolymer:
# Changelog

```yaml
- description: |
    Update TxInsReference haddock to mention public key inputs
# uncomment types applicable to the change:
  type:
  # - feature        # introduces a new feature
  # - breaking       # the API has changed in a breaking way
  # - compatible     # the API has changed but is non-breaking
  # - optimisation   # measurable performance improvements
  # - refactoring    # QoL changes
  # - bugfix         # fixes a defect
  # - test           # fixes/modifies tests
  # - maintenance    # not directly related to the code
  # - release        # related to a new release preparation
   - documentation  # change in code docs, haddocks...
```

# Context

https://github.com/IntersectMBO/cardano-node/pull/6174#discussion_r2109107088

Adds clarification public key resolved datum-related types.

# Checklist

- [ ] Commit sequence broadly makes sense and commits have useful messages
- [ ] New tests are added if needed and existing tests are updated. See [Running tests](https://github.com/input-output-hk/cardano-node-wiki/wiki/Running-tests) for more details
- [ ] Self-reviewed the diff

PR #855, opened 2025-06-05
More detailed explanation of public key inputs resolved datums for TxInsReference

carbolymer:
# Changelog

```yaml
- description: |
    More detailed explanation of public key inputs resolved datums for TxInsReference
  type:
   - documentation  # change in code docs, haddocks...
```

# Context

https://github.com/IntersectMBO/cardano-api/pull/851/files#r2127561291

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/src/Cardano/Api/Tx/Internal/Body.hs
+++ b/cardano-api/src/Cardano/Api/Tx/Internal/Body.hs
@@ -397,6 +397,16 @@ deriving instance Eq (TxTotalCollateral era)
 
 deriving instance Show (TxTotalCollateral era)
 
+-- | Transaction reference inputs. Those are not spent and do not require any witnessing to be included in a valid
+-- transaction. Any output can be a reference input. Reference inputs only affect the information that is passed to
+-- scripts by being included in @TxInfo@.
+--
+-- The third parameter of the 'TxInsReference' constructor stores actual resolved datums. When any outputs of referenced
+-- 'TxIns' store only datum hash, the only way to use them in the plutus script is to provide a resolved datum here.
+-- Note that inserting a datum with hash not present in the reference inputs will result in an error on transaction
+-- submission.
+--
+-- See also: https://github.com/intersectmbo/cardano-ledger/releases/latest/download/babbage-ledger.pdf, chapter 3.
 data TxInsReference build era where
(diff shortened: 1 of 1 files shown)


cardano-api/src/Cardano/Api/Internal/Tx/Body.hs
```diff
@@ -572,16 +572,24 @@ deriving instance Eq (TxTotalCollateral era)
 
 deriving instance Show (TxTotalCollateral era)
 
+-- | Transaction reference inputs. Those are not spent and do not require any witnessing to be included in a valid
+-- transaction. Any output can be a reference input. Reference inputs only affect the information that is passed to
+-- scripts by being included in @TxInfo@.
+--
+-- The second 'TxInsReference' constructor parameter stores actual resolved datums. When any outputs of refrenced
```
palas, line 579:
```suggestion
-- The third parameter of the 'TxInsReference' constructor stores actual resolved datums. When any outputs of referenced
```
`refrenced` is a typo, the other changes are a suggestion for clarity.

My reasoning is: if you are speaking about the constructor parameter, I would specify which constructor parameter (the third one, not the second one), and constructor can be specified by name anyway. If you don't want to specify which parameter, then you could you just say the "constructor", instead of the "constructor parameter"

cardano-api/src/Cardano/Api/Internal/Tx/Body.hs
```diff
@@ -572,16 +572,24 @@ deriving instance Eq (TxTotalCollateral era)
 
 deriving instance Show (TxTotalCollateral era)
 
+-- | Transaction reference inputs. Those are not spent and do not require any witnessing to be included in a valid
+-- transaction. Any output can be a reference input. Reference inputs only affect the information that is passed to
+-- scripts by being included in @TxInfo@.
```
palas, line 577:
```suggestion
-- | Transaction reference inputs. Reference inputs are not spent and do not require any witnessing for the
-- transaction that includes them to be valid. Any output can be a reference input. Reference inputs only
-- affect the information that is passed to scripts by being included in @TxInfo@.
```
Just a suggestion for clarity, it is a bit more redundant, so your choice

palas (approved):
Nice 👍

Jimbo4350 (approved):

PR #885, opened 2025-07-02
Add cardano-rpc package

carbolymer:
# Changelog

```yaml
- description: |
    cardano-rpc: Add cardano-rpc package with a dummy gRPC service
  type:
```

# Context

This PR adds new `cardano-rpc` cabal module.

It is not buildable with GHC 9.12 yet, so in GHA GHC 9.10 is used instead.

Tested in: https://github.com/IntersectMBO/cardano-node/pull/6205

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/.github/workflows/hls.yml
+++ b/.github/workflows/hls.yml
@@ -113,4 +113,10 @@ jobs:
 
       - name: Test HLS works
         if: steps.check_trivial_changes.outputs.CHECK_HLS_WORKS > 0
-        run: haskell-language-server
+        run: |
+          # workaround for https://github.com/haskell/haskell-language-server/issues/3735
+          cat <<EOF > hie.yaml
+          cradle:
+            cabal:
+          EOF
+          haskell-language-server
--- a/cardano-rpc/Setup.hs
+++ b/cardano-rpc/Setup.hs
@@ -1,3 +1,10 @@
+#!/usr/bin/env cabal
+{- cabal:
+build-depends: base, Cabal, proto-lens-setup
+-}
+
+-- this is a cabal script because of https://github.com/haskell/haskell-language-server/issues/3735
+
 import Data.ProtoLens.Setup
 
 main = defaultMainGeneratingProtos "proto"


flake.nix
```diff
@@ -104,9 +104,9 @@
               ghcid = "0.8.9";
               cabal-gild = "1.3.1.2";
               fourmolu = "0.18.0.0";
-              haskell-language-server.src = nixpkgs.haskell-nix.sources."hls-2.9";
+              haskell-language-server = "latest";
               # This index-state makes it work for GHC 9.8.2 (it will need to tbe removed for 9.8.4)
-              hlint = { version = "3.8"; index-state = "2024-12-01T00:00:00Z"; };
+              hlint = "3.10"; # { version = "3.10"; index-state = "2024-12-01T00:00:00Z"; };
```
carbolymer, line 109:
```suggestion
              hlint = "3.10";
```

.github/workflows/haskell.yml
```diff
@@ -158,11 +174,11 @@ jobs:
...
-    # - name: Setup tmate session
-    #  if: ${{ failure() }}
-    #  uses: mxschmitt/action-tmate@v3
-    #  with:
-    #     limit-access-to-actor: true
+    - name: Setup tmate session
+      if: ${{ failure() }}
+      uses: mxschmitt/action-tmate@v3
+      with:
+         limit-access-to-actor: true
```
carbolymer, line 181:
```suggestion
    # - name: Setup tmate session
    #   if: ${{ failure() }}
    #   uses: mxschmitt/action-tmate@v3
    #   with:
    #      limit-access-to-actor: true
```

Jimbo4350 (changes requested):
This PR needs to be broken up into smaller pieces. I suggest:
- Open a PR with the dummy example.
- Open a PR per mini-protocol along with their protobuf files.

.github/workflows/haskell.yml
carbolymer: You uncomment this, and then the CI action waits infinitely printing an a ssh command, which allows you to interactively log in and troubleshoot on the GHA runner.

cardano-rpc/CHANGELOG.md
```diff
@@ -0,0 +1,6 @@
+# Changelog for cardano-rpc
+
+## 10.0.0.0
```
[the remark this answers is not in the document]
carbolymer: It will be a node dependency, so I figured it should follow the same suit as cardano-api and cardano-cli, which all of them are basically clients connecting to the node. I guess in case of gRPC will be less problems with compatibility, because it'll be dictated by proto files, and not by N2C implementation. So I guess we can start with numbering from 0.

cardano-rpc/proto/utxorpc/v1alpha/bitcoin/bitcoin.proto
```diff
@@ -0,0 +1,48 @@
+syntax = "proto3";
+
+package utxorpc.v1alpha.bitcoin;
+
+// Represents an input in a Bitcoin transaction.
```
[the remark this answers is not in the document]
carbolymer: I guess. I have taken it from UTxO RPC as is. We don't have to implement it, but I'd keep it for keeping track of the changes w/ UTxO RPC.

.github/workflows/haskell.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc: ["9.6", "9.12"]
```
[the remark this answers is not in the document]
Jimbo4350: I don't think we need 9.12 at the moment. It will have a knock on effect for cardano-node's CI so I would give devx a heads up.

cardano-rpc/CHANGELOG.md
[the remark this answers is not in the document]
Jimbo4350: Ah yes, makes sense.

fourmolu.yaml
```diff
@@ -54,6 +54,8 @@ sort-deriving-clauses: false
 trailing-section-operators: true
 unicode: never
 respectful: false
-fixities: []
+fixities:
```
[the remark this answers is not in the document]
palas: Ah, nice!

cardano-rpc/src/Cardano/Rpc/Server.hs
```diff
@@ -0,0 +1,95 @@
...
+          , tracer = natTracer liftIO tracer
+          , rpcLocalNodeConnectInfo = mkLocalNodeConnectInfo nodeSocketPath networkMagic
+          }
+
+  -- TODO this is logged by node configuration already, so it would make sense to log it again when
+  -- configuration gets reloaded
+  -- putTrace $ "RPC configuration: " <> show rpcConfig
+
+  when isEnabled $
+    runRIO rpcEnv $
```
Jimbo4350, line 84: The environment appears to not be used anywhere. Why not just have:
```diff
   when isEnabled $
-    runRIO rpcEnv $
-      withRunInIO $ \runInIO ->
-        runServerWithHandlers def config . fmap (hoistSomeRpcHandler runInIO) $
-          mconcat
-            [ fromMethods methodsNodeRpc
-            ]
+    runServerWithHandlers def config $
+      fromMethods methodsNodeRpc
```

cardano-rpc/src/Cardano/Rpc/Server/Internal/Monad.hs
```diff
@@ -0,0 +1,36 @@
...
+  )
+where
+
+import Cardano.Api
+
+import Cardano.Rpc.Server.Internal.Env
+
+import RIO
+
+class Has field env where
```
Jimbo4350, line 26: `Has` class seems unnecessary.

Jimbo4350 (changes requested):

cardano-rpc/src/Cardano/Rpc/Server.hs
carbolymer: Removed

cardano-rpc/src/Cardano/Rpc/Server/Internal/Monad.hs
carbolymer: Removed

cardano-rpc/cardano-rpc.cabal
```diff
@@ -0,0 +1,84 @@
+cabal-version: 3.8
+name: cardano-rpc
+version: 0.0.0.0
```
Jimbo4350, line 3: `10.0.0.0`?

cardano-rpc/src/Cardano/Rpc/Server/Config.hs
```diff
@@ -0,0 +1,91 @@
...
+  -> m RpcConfig
+makeRpcConfig
+  RpcConfig
+    { isEnabled = Last mIsEnabled
+    , rpcSocketPath = Last mRpcSocketPath
+    , nodeSocketPath = Last mNodeSocketPath
+    } = do
+    let isEnabled = fromMaybe False mIsEnabled
+        -- default to a some non-existing path. Does not matter if the gRPC endpoint is disabled
+        nodeSocketPath = fromMaybe "./node.socket" mNodeSocketPath
```
Jimbo4350, line 75: Abstracting this as a separate value (as we do for the node configuration) would be clearer IMO.

Jimbo4350 (approved):
LGTM :+1:

cabal.project
```diff
@@ -50,12 +51,11 @@ write-ghc-environment-files: always
 jobs: $ncpus
 semaphore: True
 
--- IMPORTANT
--- Do NOT add more source-repository-package stanzas here unless they are strictly
--- temporary! Please read the section in CONTRIBUTING about updating dependencies.
-
-if impl (ghc >= 9.12)
+if impl (ghc >= 9.10)
```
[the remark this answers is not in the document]
carbolymer: I'm not sure we actually need to change GHC bound. I'll double check.

.github/workflows/haskell.yml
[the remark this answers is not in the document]
carbolymer: FYI @erikd

PR #901, opened 2025-07-14
Add typescript declarations for `cardano-wasm` API and generate it automatically from `ApiInfo`

palas:
# Changelog

```yaml
- description: |
    Added typescript declaration for `cardano-wasm` API
  type:
  - feature
  - documentation
```

# Context

Because JavaScript is untyped, this PR adds TypeScript declarations that both help users of `cardano-wasm` API to get types right, and also they will get detailed contextual documentation when using it from IDEs.

# How to trust this PR

What I checked is that the generated `d.ts` file is the same as the manually written one (pretty much). I also checked that you can use it from VSCode and get autocompletion and contextual documentation.

I would check the style and that there are no bugs. You can also try running the `.d.ts` generation (which is the main executable of `cardano-wasm`)

# Checklist

- [x] Self-reviewed the diff


cardano-wasm/example/cardano-api.d.ts
```diff
@@ -0,0 +1,99 @@
+// cardano-api.d.ts
+
+export default initialize;
+
+/**
+ * Initializes the Cardano API.
+ * @returns A promise that resolves to the main `CardanoAPI` object.
+ */
+declare function initialize(): Promise<CardanoAPI>;
```
carbolymer, line 9: suggestion: use british english, instead of american english?
```suggestion
declare function initialise(): Promise<CardanoAPI>;
```

cardano-wasm/src/Cardano/Wasm/Internal/Api/TypeScriptDefs.hs
```diff
@@ -0,0 +1,134 @@
+module Cardano.Wasm.Internal.Api.TypeScriptDefs where
+
+-- import qualified Data.Text.Lazy.IO as TIO
```
carbolymer, line 3: leftover comment?

cardano-wasm/src/Cardano/Wasm/Internal/Api/TypeScriptDefs.hs
```diff
@@ -0,0 +1,134 @@
+module Cardano.Wasm.Internal.Api.TypeScriptDefs where
```
carbolymer, line 1: Top level exports (which means everything) should have haddocks

cardano-wasm/src/Cardano/Wasm/Internal/Api/TypeScriptDefs.hs
carbolymer, line 1: What's the reason behind generating typescript by hand instead of using existing libraries? https://hackage.haskell.org/packages/search?terms=typescript

Could you add the justification to the PR description?

palas: I couldn't find one that did what we needed (but maybe I didn't look well enough). I understand that the `aeson-typescript` and  the `servant-typescript` ones are just for generating typescript for their own definitions of JSON and HTTP APIs respectively. So I think they don't really fit.

Maybe [language-typescript](https://hackage.haskell.org/package/language-typescript) would fit, but it has 4 downloads, so it is not really mature anyway, not sure it gives us a lot.

I will research a bit more and justify it in the PR like you say.

cardano-wasm/src/Cardano/Wasm/Internal/Api/TypeScriptDefs.hs
palas: Addressed in amended commit

cardano-wasm/src/Cardano/Wasm/Internal/Api/TypeScriptDefs.hs
palas: I haven't found anything better, so I added the explanation to the top level comment of the module in the amended comment.

cardano-wasm/example/cardano-api.d.ts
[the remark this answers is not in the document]
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/901/commits/69d5af63acd9af965c9a90e9a8969450c847b151

I tried adding it as a test suite but I run into several issues:
- First that the libs from `exe` are not accessible (so I would have to split the lib into lib and exe, we probably will have to do that at some point, but well)
- I tried adding it as `build-depend-tool`, and that worked locally, but it broke nix completely (because it tries to pre-fetch dependencies and it couldn't find it, since it was in the same folder, and not in CHaP)

And I foresee that even if I split the lib in two, I will still have problem with nix trying to access a folder that is outside of the project folder (`examples`). So adding it to the workflow seemed like the simplest solution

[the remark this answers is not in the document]
carbolymer: tried that here, there are seemingly unrelated constraints resultion errors: https://github.com/IntersectMBO/cardano-api/pull/907/files

cabal.project
```diff
@@ -29,6 +29,10 @@ if impl(ghc < 9.8)
 
 constraints: process >= 1.6.26.1
 
+-- this slows down the build plan preparation, but it seems that without it
+-- cabal has problems with resolving dependencies
```
carbolymer, line 33:
```suggestion
-- It may slow down build plan preparation, but without it cabal has problems
-- with solving constraints. Remove this when not needed anymore.
```

cardano-wasm/src/Cardano/Wasm/Internal/Api/TypeScriptDefs.hs
```diff
@@ -0,0 +1,188 @@
...
+-- with the specified indentation level (in spaces) and each line
+-- in the list of strings as a separate line in the comment.
+-- The first line starts with `/**`, subsequent lines (corresponding
+-- to each line in the list) start with ` * `,
+-- and the last line ends with ` */`.
+-- The indentation level is used to indent the entire comment block.
+buildMultilineComment :: Int -> [String] -> TLB.Builder
+buildMultilineComment indentLevel commentLines =
+  let indentation = TLB.fromLazyText $ TL.replicate (fromIntegral indentLevel) " "
+      bodyIndentation = indentation <> TLB.fromString " * "
```
carbolymer, line 34: You can avoid this with `OverloadedStrings`
```suggestion
      bodyIndentation = indentation <> " * "
```

carbolymer (approved):

Jimbo4350 (approved):

cardano-wasm/example/cardano-api.d.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * Sets the fee for the transaction.
+     * @param lovelaceAmount The fee amount in lovelace.
+     * @returns The `UnsignedTx` object with the set fee.
+     */
+    setFee(lovelaceAmount: bigint): UnsignedTx;
+
+    /**
+     * Estimates the minimum fee for the transaction.
```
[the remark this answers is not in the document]
palas: I would say it is already an actual library (combined with the generated files). The reason it is in the `example` folder is only for convenience, but I can see how that may send the wrong message. That is also the case for [cardano-api.js](https://github.com/IntersectMBO/cardano-api/blob/d50219349d23ac3857c324f577da9a7986d183d7/cardano-wasm/example/cardano-api.js). So we can restructure things

cabal.project
palas: Applied here: https://github.com/IntersectMBO/cardano-api/pull/901/commits/88979f1d77b050850e4bb9cd2a6ac803b8a18f3a

cardano-wasm/src/Cardano/Wasm/Internal/Api/TypeScriptDefs.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/901/commits/8dc0d12a3f92174a4c829d9115726e5743464159

cardano-wasm/example/cardano-api.d.ts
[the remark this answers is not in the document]
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/901/commits/50928d413d38d019b1f2ab70fb4623c3e3639d58

PR #908, opened 2025-07-17
Add playwright test to `cardano-wasm` to ensure it works in browsers

palas:
# Changelog

```yaml
- description: |
    Added playwright test to `cardano-wasm` and `.#playwright` development shell
  type:
  - test
```

# Context

One of the objectives of https://github.com/input-output-hk/cardano-node-wiki/pull/75 is to ensure that the `JavaScript` library can be used easily in vanilla JS (from the browser). This PR adds a `playwright` test that verifies the example in the `example` folder outputs what is expected when run exclusively on the browser.

As part of this work, this PR also adds a development shell `nix develop .#playwright` that includes the python server and the playwright setup for facilitating testing.

# How to trust this PR

It is a test, so it passing it already gives a lot of assurance. You could modify the test to verify it indeed fails. And I would check that the code structure is appropriate and that there is no dead code. You can also check that the shell works for you, but hydra should do that automatically.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/.github/workflows/haskell-wasm.yml
+++ b/.github/workflows/haskell-wasm.yml
@@ -142,6 +142,21 @@ jobs:
       run: |
         wasm32-wasi-cabal build cardano-wasm --no-semaphore -j1 --ghc-options="-j1"
 
+    - name: Prepare example
+      run: |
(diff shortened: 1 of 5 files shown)


Jimbo4350 (approved):
LGTM however I would have a look at [Quickstrom](https://quickstrom.io/) as well.

PR #917, opened 2025-07-25
Create `Wallet` virtual object and add support for handling payment addresses

palas:
# Changelog

```yaml
- description: |
    Added support for handling payment addresses and keys to `cardano-wasm` API
  type:
```

# Context

Before this PR `cardano-wasm` API already supports constructing transactions by providing a signing key in bech32, a target address in bech32, and a utxo, but it didn't have a way of converting between keys and addresses, or between signing key and verification key, or for generating new keys. This PR addresses that by adding basic handling of payment addresses (without stake address).

# How to trust this PR

I would check I am using the right functions the right way, and that the nomenclature is precise. Also would be good to make sure I am not making design mistakes that will force a redesign as we add new features later.

# Checklist

- [x] Self-reviewed the diff


cardano-wasm/cardano-wasm.cabal
```diff
@@ -37,7 +37,7 @@ executable cardano-wasm
     ghc-options:
       -no-hs-main
       -optl-mexec-model=reactor
-      "-optl-Wl,--strip-all,--export=getApiInfo,--export=newConwayTx,--export=addTxInput,--export=addSimpleTxOut,--export=setFee,--export=estimateMinFee,--export=signWithPaymentKey,--export=alsoSignWithPaymentKey,--export=txToCbor,--export=newGrpcConnection,--export=getEra"
+      "-optl-Wl,--strip-all"
```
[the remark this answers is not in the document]
palas: It doesn't seem to have an effect. I thought it was necessary, but empirically, it doesn't seem to be

cardano-wasm/src/Cardano/Wasm/Internal/Api/Address.hs
```diff
@@ -0,0 +1,98 @@
...
+import Cardano.Wasm.Internal.ExceptionHandling (rightOrError, toMonadFail)
+
+import Data.Aeson ((.=))
+import Data.Aeson qualified as Aeson
+import Data.Aeson.Types qualified as Aeson
+import Data.Text (Text)
+import Data.Text.Encoding qualified as Text
+import GHC.Generics (Generic)
+
+data AddressObject = PaymentAddress
```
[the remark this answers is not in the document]
palas: I will use `Wallet` instead of `Address` then

cardano-wasm/src/Cardano/Wasm/Internal/Api/Info.hs
```diff
@@ -242,9 +281,40 @@ apiInfo =
...
+                      , methodReturnType = NewObject walletObjectName
+                      , methodReturnDoc = "A promise that resolves to a new `Wallet` object."
+                      }
+                  , MethodInfo
+                      { methodName = "restoreMainnetPaymentWalletFromSigningKeyBech32"
+                      , methodDoc = "Restore a mainnet payment wallet from a Bech32 encoded signing key."
+                      , methodParams = [ParamInfo "signingKeyBech32" "string" "The Bech32 encoded signing key."]
+                      , methodReturnType = NewObject walletObjectName
+                      , methodReturnDoc = "A promise that resolves to a new `Wallet` object."
+                      }
```
carbolymer, line 297: I think we should skip `Mainnet` from function names, since it should be a default mode.

cardano-wasm/src/Cardano/Wasm/Internal/Api/Wallet.hs
```diff
@@ -0,0 +1,137 @@
...
+  PaymentWallet Mainnet <$> toMonadFail key
+
+-- | Generate a simple payment wallet for testnet, given the testnet's network magic.
+generateTestnetPaymentWalletImpl :: Int -> IO WalletObject
+generateTestnetPaymentWalletImpl networkMagic = do
+  let seedSize = deterministicSigningKeySeedSize AsPaymentKey
+  randomBytes <- getRandomBytes seedSize
+  let seed = mkSeedFromBytes randomBytes
+      key = deterministicSigningKey AsPaymentKey seed
+  return (PaymentWallet (Testnet (NetworkMagic (fromIntegral networkMagic))) key)
```
carbolymer, line 105: Suggestion: use `$`

cardano-wasm/src/Cardano/Wasm/Internal/JavaScript/Random.hs
```diff
@@ -0,0 +1,33 @@
...
+
+import Data.ByteString.Base16 qualified as Base16
+import Data.ByteString.Char8 qualified as BSC
+import GHC.Wasm.Prim
+
+-- | Create a GRPC-web client for the Cardano API.
+foreign import javascript safe "{ let numBytes = $1; \
+                                  let randomBytes = new Uint8ClampedArray(numBytes); \
+                                  crypto.getRandomValues(randomBytes); \
+                                  return (randomBytes.reduce((acc, next) => acc.concat(next.toString(16).padStart(2, '0')), '')); \
```
carbolymer, line 24: I'm wondering if this code could be redefined and reused. Seems that to hex conversion appears quite often.

palas: I can think of two potential ways:
- One is passing it as a parameter (which is not very neat and complicates things)
- The other one is is to define it globally somewhere, probably in `cardano-api.js` (so it becomes accessible by everybody, but that is not too bad, at least for this function)
I am inclined towards the second one.

carbolymer: Yeah I agree with the second one. I think we could even add `toHex` to `Uint8ClampedArray.prototype` there.

Jimbo4350 (approved):
LGTM

cardano-wasm/src/Cardano/Wasm/Internal/Api/Info.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/917/commits/328637ed53638b0d88e215b8e7628d90f1c57c61

cardano-wasm/src/Cardano/Wasm/Internal/Api/Wallet.hs
```diff
@@ -0,0 +1,137 @@
...
+  randomBytes <- getRandomBytes seedSize
+  let seed = mkSeedFromBytes randomBytes
+      key = deterministicSigningKey AsPaymentKey seed
+  return (PaymentWallet Mainnet key)
+
+-- | Restore a mainnet payment wallet from a Bech32 encoded signing key.
+restoreMainnetPaymentWalletFromSigningKeyBech32Impl :: String -> IO WalletObject
+restoreMainnetPaymentWalletFromSigningKeyBech32Impl signingKeyBech32 = do
+  let key = rightOrError $ deserialiseFromBech32 (Text.pack signingKeyBech32)
+  PaymentWallet Mainnet <$> toMonadFail key
```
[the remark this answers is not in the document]
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/917/commits/a281b698c780bb204564c1c9044dcef84c1bc5f6

cardano-wasm/src/Cardano/Wasm/Internal/Api/Wallet.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/917/commits/a281b698c780bb204564c1c9044dcef84c1bc5f6

PR #920, opened 2025-07-28
Add support for submitting transactions through gRPC-web to the `cardano-wasm` API

palas:
# Changelog

```yaml
- description: |
    Added support for submitting transactions through gRPC-web to the `cardano-wasm` API
  type:
  projects:
  - cardano-wasm
```

# Context

This PR expands the gRPC-web support functionality to include submission of transactions that can already be built by using existing functionalities. It was already possible to do this by using a web browser wallet, but this makes it easier to do it without one.

# How to trust this PR

It is a small PR. It is possible to test it by using the code in the [submit-tx-example](https://github.com/IntersectMBO/cardano-api/tree/submit-tx-example) branch, but it is tricky to set up. I would review the code and see if it makes sense, if you can spot any issues, and if the code style is nicely organised. We'll be testing this functionality many times in the near future as other functionalities are added anyway.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-wasm/lib-wrapper/cardano-api.d.ts
+++ b/cardano-wasm/lib-wrapper/cardano-api.d.ts
@@ -95,6 +95,13 @@ declare interface GrpcConnection {
      * @returns A promise that resolves to the current era number.
      */
     getEra(): Promise<number>;
+
+    /**
+     * Submit a signed and CBOR-encoded transaction to the Cardano node.
+     * @param txCbor The CBOR-encoded transaction as a hex string.
+     * @returns A promise that resolves to the transaction ID.
+     */
+    submitTx(txCbor: string): Promise<string>;
 }
 
 /**
--- a/cardano-wasm/src/Cardano/Wasm/Internal/JavaScript/GRPC.hs
+++ b/cardano-wasm/src/Cardano/Wasm/Internal/JavaScript/GRPC.hs
@@ -4,13 +4,14 @@
 module Cardano.Wasm.Internal.JavaScript.GRPC where
 #else
 
-module Cardano.Wasm.Internal.JavaScript.GRPC (js_newWebGrpcClient, js_getEra) where
+module Cardano.Wasm.Internal.JavaScript.GRPC (js_newWebGrpcClient, js_getEra, js_submitTx) where
 
 import GHC.Wasm.Prim
 
 -- | Create a GRPC-web client for the Cardano API.
 foreign import javascript safe "{ node: new cardano_node.node.NodePromiseClient($1, null, null), \
-                                  query: new cardano_node.query.QueryServicePromiseClient($1, null, null) \
+                                  query: new cardano_node.query.QueryServicePromiseClient($1, null, null), \
+                                  submit: new cardano_node.submit.SubmitServicePromiseClient($1, null, null) \
                                 }"
   js_newWebGrpcClientImpl :: JSString -> IO JSVal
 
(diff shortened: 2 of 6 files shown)


cardano-wasm/src/Cardano/Wasm/Internal/JavaScript/GRPC.hs
```diff
@@ -21,4 +22,21 @@ js_newWebGrpcClient = js_newWebGrpcClientImpl . toJSString
 foreign import javascript safe "($1).node.getEra(new proto.Empty(), {})"
   js_getEra :: JSVal -> IO Int
 
+-- | Submit a transaction to the Cardano API using a GRPC-web client.
+foreign import javascript safe "{ let tx = new cardano_node.submit.AnyChainTx(); \
+                                  tx.setRaw(new Uint8Array($2.match(/[0-9a-fA-F]{2}/g).map((byte) => parseInt(byte, 16)))); \
+                                  let txList = new cardano_node.submit.SubmitTxRequest(); \
+                                  txList.addTx(tx); \
+                                  return (($1).submit.submitTx(txList, {}).then((val) => (val.toObject().resultsList[0].ref == '')?('error: ' + val.toObject().resultsList[0].errorMessage):('ok: ' +val.toObject().resultsList[0].ref))) }"
```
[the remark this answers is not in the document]
palas: `String` is one of the few primitives supported, the way I would bass an object is by serialising it to JSON, so it would be the same. You can see the supported types here: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/wasm.html#marshalable-types-and-jsval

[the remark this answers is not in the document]
palas: It would be more structured, but then we have to do a new datatype with its `FromJSON` instance in Haskell, so I think it is overkill, this seems simpler and easier to debug while it is a simple string. Also the format is not good, because it should be a sum type, and it has a field for the error no matter whether there is an error or not, so I think it is best to get rid of that in the JS glue, rather than transferring it as is to the Haskell, and we can decode this easily as an `Either` which is already much better

cardano-wasm/lib-wrapper/cardano-api.d.ts
```diff
@@ -95,6 +95,13 @@ declare interface GrpcConnection {
      * @returns A promise that resolves to the current era number.
      */
     getEra(): Promise<number>;
+
+    /**
+     * Submit a signed and CBOR-encoded transaction to the Cardano node.
+     * @param txCbor The CBOR-encoded transaction as a hex string.
+     * @returns A promise that resolves to the transaction ID.
+     */
+    submitTx(txCbor: string): Promise<string>;
```
carbolymer, line 104: Shouldn't the argument be `Uint8Array` instead of `string` if it represents binary CBOR representation?

cardano-wasm/src/Cardano/Wasm/Internal/JavaScript/GRPC.hs
```diff
@@ -21,4 +22,21 @@ js_newWebGrpcClient = js_newWebGrpcClientImpl . toJSString
...
+  js_submitTxImpl :: JSVal -> JSString -> IO JSString
+
+js_submitTx :: JSVal -> String -> IO (Either String String)
+js_submitTx client txBytes = do
+  result <- js_submitTxImpl client (toJSString txBytes)
+  let resultStr = fromJSString result
+  case resultStr of
+    ('o':'k':':':' ':res) -> return $ Right res
+    ('e':'r':'r':'o':'r':':':' ':err) -> return $ Left err
+    _ -> return $ Left "Transaction submission failed"
```
carbolymer, line 40: Suggestion: use `MultiWayIf` and `isPrefixOf` for better readability

cardano-wasm/lib-wrapper/cardano-api.d.ts
palas: No, cause it is the hex (base16) encoding of it

carbolymer: Is there benefit of using hex strings instead of `Uint8Array`?

cardano-wasm/src/Cardano/Wasm/Internal/JavaScript/GRPC.hs
palas: I am a bit torn. You've got a point, I think the readable version could be matching on the result of `splitOn ": "`. But I think the current version gets compiled to something much simpler than that, probably it only checks the first character and it uses the checking for also splitting the string in two (instead of doing two passes), and I agree the character chain is not super readable, and also the difference in efficiency is probably completely negligible. But it is still kind of idiomatic, not that unreadable either, and already a bit of a hack all in all.

So I think I am going to leave it there for now, and maybe we can come back to it later

PR #940, opened 2025-08-27
Example wallet using `cardano-wasm`

palas:
# Changelog

```yaml
- description: |
    Added simple wallet example using `cardano-wasm`
  type:
  - feature
```

# Context

This PR adds an example to `cardano-wasm` that showcases most of the functionalities exposed at the moment.

# How to trust this PR

Really the best way is to test it. But I have found that actually this example worked as a test and exposed a few existing issues, which this PR also addresses.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-wasm/wallet-example/index.html
+++ b/cardano-wasm/wallet-example/index.html
@@ -1,0 +1,123 @@
+<html>
+
+<head>
+  <title>cardano-wasm test</title>
+  <link rel="stylesheet" type="text/css" href="./stylesheet.css">
+</head>
+
+<body>
+  <!--<script type="module" src="./cardano_node_grpc_web_pb.js"></script>
+    <script type="module" src="./example.js"></script>-->
+  <div id="header">
+    <h1>Wallet example</h1>
+  </div>
+  <div id="app">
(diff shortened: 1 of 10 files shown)


cardano-wasm/wallet-example/stylesheet.css
carbolymer, line 1: We have multiple examples in the `cardano-wasm` directory: https://github.com/IntersectMBO/cardano-api/tree/master/cardano-wasm. I think we should group them into one `examples` directory.

cardano-wasm/wallet-example/index.html
carbolymer, line 1: I think we should add a README how to run each example. Or a single README in the `examples` folder.

cardano-wasm/wallet-example/stylesheet.css
palas: I was thinking the same, let me do it in another PR so that the diffs remain sane

cardano-wasm/wallet-example/index.html
palas: 100%. I will do this in the next PR too so that it gets reviewed independently

PR #941, opened 2025-08-28
Organise examples

palas:
# Changelog

```yaml
- description: |
    Organised examples into a single folder and added README.md to `simple-wallet` example
  type:
  - documentation  # change in code docs, haddocks...
```

# Context

Because we already have three examples, it is starting to clutter the top-level folder (`cardano-wasm`), so this PR moves the example folders into the new general `example` folder, renames them to not include the "example" word, and updates the appropriate references.

This PR also updates `README.md` files giving them hierarchical structure and in particular adding a new `README.md` file for `simple-wallet` example (closely based in the one from `grpc` example).

# How to trust this PR

The only good way would be to try the examples following the instructions. But they remain as they were just updated the paths to match the new structure. The important thing is ensuring that nothing in the project assumes that the examples were located somewhere else and that the splitting kept everything consistent.

# Checklist

- [x] Self-reviewed the diff


cardano-wasm/examples/grpc/README.md
```diff
@@ -0,0 +1,51 @@
...
+Before running the example make sure you have built the `cardano-wasm` library by following the process described in the readme file of the [`cardano-wasm` folder](../..).
+
+## Running the GRPC example
+
+To run the example in this folder:
+
+1.  Run an instance of the `cardano-node` with the GRPC server enabled and put the socket file for the GRPC server in the root folder of this repo with the name `rpc.socket`. (You can put it somewhere else, but you will have to update the `envoy-conf.yaml` function later.)
+
+>[!NOTE]
+>You can run `cardano-testnet cardano --testnet-magic 42 --conway-era --enable-grpc` and then export `CARDANO_NODE_SOCKET_PATH` variable as shown on the screen.
```
carbolymer, line 13: it's simpler now
```suggestion
>You can run `cardano-testnet cardano --enable-grpc` and then export `CARDANO_NODE_SOCKET_PATH` variable as shown on the screen.
```

carbolymer (approved):

PR #942, opened 2025-08-29
Remove test library dependencies

Jimbo4350:
# Changelog

```yaml
- description: |
    Remove test library dependencies as it is bad practice to depend on them. This is a breaking change because there may be users that depend on both `cardano-api`'s test/gen libraries as well as ledger's test libraries. We think this is highly unlikely however.
  type:
  - breaking    
```

# Context

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff


carbolymer:
Why are we vendoring-in instances from ledger to cardano-api?

carbolymer:
@Jimbo4350 I think you need to bump HLS cache version in .github/workflows/hls.yml to get rid of HLS failure.

cardano-api/gen/Test/Gen/Cardano/Api/Era.hs
```diff
@@ -1,33 +1,31 @@
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE UndecidableInstances #-}
```
carbolymer, line 6: Why is this needed?

cardano-api/test/cardano-api-test/Test/Cardano/Api/Ledger.hs
```diff
@@ -9,31 +8,20 @@ where
...
 import Hedgehog qualified as H
-import Hedgehog.Gen.QuickCheck (arbitrary)
 import Hedgehog.Internal.Property
 import Test.Tasty (TestTree, testGroup)
 import Test.Tasty.Hedgehog (testProperty)
 
--- Keep this here to make sure serialiseAddr/deserialiseAddr are working.
--- They are defined in the Shelley executable spec and have been wrong at
--- least once.
-prop_roundtrip_Address_CBOR :: Property
```
carbolymer, line 31: We don't need to test address cbor serialisation because...?

Jimbo4350: Because ledger already has this test: https://github.com/IntersectMBO/cardano-ledger/blob/938c10ef801328fc2fcf3a152da3c990b1c6fae8/libs/cardano-ledger-core/test/Test/Cardano/Ledger/AddressSpec.hs#L93

carbolymer (approved):
LGTM!

Three things:
1. I think this should be a breaking change, because there's an inference between plutus and ledger testlibs and cardano-api now.
2. Could you try build cardano-node repo with this version of cardano-api, to check that we're not running into duplicated instances errors there?
3. Can you add an explanation to the PR description why this change is being made? Could be useful in the future.

Jimbo4350:
> LGTM!
> 
> Two things:
> 
>     1. I think this should be a breaking change, because there's an interference between plutus and ledger testlibs and cardano-api now.
> 
>     2. Could you try build cardano-node repo with this version of cardano-api, to check that we're not running into duplicated instances errors there?

Updated the description and it builds with `cardano-node` master :+1:.

Issue #922, opened 2025-07-29
Split interface into multiple files


palas:
We may want to start considering multiple interface files (if possible) vs a single monolithic interface file.

_Originally posted by @Jimbo4350 in https://github.com/IntersectMBO/cardano-api/pull/917#discussion_r2236904897_

PR #948, opened 2025-09-05
Output typescript declaration to a file instead of standard output

palas:
# Changelog

```yaml
- description: |
    Modified `cardano-wasm` executable to output to a file instead of to standard output
  type:
  - refactoring
```

# Context

In order to address this issue: https://github.com/IntersectMBO/cardano-api/issues/922. We need to be able to output files, because outputting several files to standard output is not practical, since they would get mixed. So this PR modifies the `cardano-wasm` executable so that it takes a folder and outputs the typescript files to it. This way we can, in the future, output several files instead of just one.

# How to trust this PR

The existing test had to be updated, but the fact that this functionality is tested gives me great assurance. I would look at the non-functional aspects of code, like clarity and structure, and whether the code is future proof.

# Checklist

- [x] Self-reviewed the diff


cardano-wasm/test/cardano-wasm-golden/Test/Golden/Cardano/Wasm/TypeScript.hs
```diff
@@ -1,10 +1,29 @@
...
+
 import Hedgehog as H
 import Hedgehog.Extras qualified as H
 
 hprop_cardano_wasm_typescript_declarations_match_generated :: Property
 hprop_cardano_wasm_typescript_declarations_match_generated =
   H.propertyOnce $ do
-    result <- H.execFlex "cardano-wasm" "CARDANO_WASM" []
-    H.diffVsGoldenFile result "lib-wrapper/cardano-api.d.ts"
+    control $ \runInHedgehog ->
```
carbolymer, line 15: Why not use hedgehog-extras [`workspace`](https://hackage-content.haskell.org/package/hedgehog-extras-0.9.0.0/docs/Hedgehog-Extras-Test-Base.html#v:workspace)?

Issue #955, opened 2025-09-12
make a deserialiseFromCbor instance for TxOut


CarlosLopezDeLara:
Via Phil DiSarro https://input-output-rnd.slack.com/archives/CMT12F94M/p1756434229504609

PR #960, opened 2025-09-18
Add `SerialiseAsCBOR` instance for `TxOut`

palas:
# Changelog

```yaml
- description: |
    Added `HasTypeProxy` `FromCBOR`, `ToCBOR`, and `SerialiseAsCBOR ` instances to `TxOut`.
  type:
  projects:
  - cardano-api
```

# Context

This PR addresses https://github.com/IntersectMBO/cardano-api/issues/955.

# How to trust this PR

The test should help with trust. But there is a bit of an issue with `ToCBOR` instance losing information, so there is a design decision to be make about this. Please, let me know what you think about this.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-api/src/Cardano/Api/Tx/Internal/Output.hs
+++ b/cardano-api/src/Cardano/Api/Tx/Internal/Output.hs
@@ -1,10 +1,13 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveAnyClass #-}
 {-# LANGUAGE DerivingVia #-}
 {-# LANGUAGE DisambiguateRecordFields #-}
 {-# LANGUAGE EmptyCase #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GADTs #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE InstanceSigs #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
@@ -12,6 +15,7 @@
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
 
 module Cardano.Api.Tx.Internal.Output
   ( -- * Transaction outputs
@@ -62,7 +66,7 @@ import Cardano.Api.Era.Internal.Eon.BabbageEraOnwards
 import Cardano.Api.Era.Internal.Eon.Convert
 import Cardano.Api.Era.Internal.Eon.ShelleyBasedEra
 import Cardano.Api.Error (Error (..), displayError)
-import Cardano.Api.Hash
+import Cardano.Api.HasTypeProxy qualified as HTP
 import Cardano.Api.Ledger.Internal.Reexport qualified as Ledger
 import Cardano.Api.Monad.Error
 import Cardano.Api.Parser.Text qualified as P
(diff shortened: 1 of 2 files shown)


cardano-api/test/cardano-api-test/Test/Cardano/Api/CBOR.hs
```diff
@@ -110,6 +110,21 @@ prop_roundtrip_tx_CBOR = H.property $ do
...
+  x <- H.forAll $ genTx era
+  txOut <- H.forAll $ Gen.element $ txOuts $ getTxBodyContent $ getTxBody x
+  let fixedTxOut = hashDatum txOut
+  shelleyBasedEraConstraints era $ H.trippingCbor (proxyToAsType Proxy) fixedTxOut
+ where
+  hashDatum :: TxOut CtxTx era -> TxOut CtxTx era
+  hashDatum txOut@(TxOut aie val datum rs) =
+    case datum of
+      (TxOutSupplementalDatum aeo d) ->
+        TxOut aie val (TxOutDatumHash aeo (hashScriptDataBytes d)) rs
```
carbolymer, line 125: I am not sure how usable will be such deserialised txout. I think after you convert datum to hash, it's no longer included in the transaction output: https://github.com/IntersectMBO/cardano-api/pull/814/files?diff=split&w=1#diff-dbfef7b82f6c05596cde61360e58ef674ad0bea7ac8c3e416178ccbde84b98b9R3107

Tactical merge block, since we're still discussing design.

carbolymer (changes requested):

cardano-api/test/cardano-api-test/Test/Cardano/Api/CBOR.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/960/commits/bdb57611d08a149af5b0c75e94c308d9addfa6d9

cardano-api/src/Cardano/Api/Tx/Internal/Output.hs
carbolymer: We definitely need comments explaining the assumptions here, and why there can't be `ToCBOR TxOut`.

cardano-api/src/Cardano/Api/Tx/Internal/Output.hs
```diff
@@ -121,6 +123,17 @@ data TxOut ctx era
...
+  data AsType (TxOut ctx era) = AsTxOut (AsType era)
+  proxyToAsType :: HTP.Proxy (TxOut ctx era) -> AsType (TxOut ctx era)
+  proxyToAsType _ = AsTxOut (HTP.asType @era)
+
+instance (Typeable ctx, IsShelleyBasedEra era) => FromCBOR (TxOut ctx era) where
+  fromCBOR :: Ledger.Decoder s (TxOut ctx era)
+  fromCBOR =
+    shelleyBasedEraConstraints (shelleyBasedEra @era) $
+      pure (fromShelleyTxOut shelleyBasedEra) <*> L.fromEraCBOR @(ShelleyLedgerEra era)
+
```
carbolymer: I think we could try adding something like:
```haskell
instance TypeError (Text "Cannot serialise 'TxOut' to CBOR." :$$:
                    Text "some justification")
         => ToCBOR (TxOut ctx era) where
   showsPrec = error "unreachable"
```
https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/type_errors.html

cardano-api/test/cardano-api-test/Test/Cardano/Api/CBOR.hs
```diff
@@ -108,6 +110,30 @@ prop_roundtrip_tx_CBOR = H.property $ do
   x <- H.forAll $ genTx era
   shelleyBasedEraConstraints era $ H.trippingCbor (proxyToAsType Proxy) x
 
+prop_roundtrip_tx_out_CBOR :: Property
```
carbolymer: Comment explaining how and why are we testing this roundtrip this way would be nice.

carbolymer (approved):

PR #963, opened 2025-09-24
gRPC: Use buf for code generation instead of cabal/protoc integration

carbolymer:
# Changelog

```yaml
- description: |
    Use buf for code generation instead of cabal/protoc integration.
    Reenable cardano-rpc.
  type:
# uncomment at least one main project this PR is associated with
  projects:
  # - cardano-api
  # - cardano-api-gen
   - cardano-rpc
  # - cardano-wasm
```

# Context

## Current state
In the previous version of `cardano-rpc` package, the code was generated during cabal-build which was prolonging build process and was pulling `proto-lens-protoc` into the dependency tree. This was prone to breakage and was causing complicated issues with cabal's dependency resolution.

## The proposed solution
This PR changes that by using an external tool [buf](https://buf.build/docs/cli/) which allows to generate code from protobuf definitions for various languages. This should be extended for typescript/javascript code generation later.

The caveat here is that it includes generated files in the repository which have following effects:
- They increase diff sizes and thus repository size - this will be more significant once we decide to check-in code from more languages
- Generated files have to be kept in sync - this PR adds a GHA build step for that

The end goal is to not use generated code in the production code (or minimal amount of it), and rely on the packages provided by utxorpc:
- https://hackage.haskell.org/package/utxorpc
- https://www.npmjs.com/package/@utxorpc/spec

The ultimate purpose of keeping the generated code in the repository is the ease of work on proto files when one would like to introduce changes to proto definitions without having to go through [utxorpc](https://github.com/utxorpc/spec) upstream, which is a time consuming process.

## Alternatives

An alternative is to move the generated code using `buf` to a separate repository. The separate repository would be integrated by SRP only when it's needed.

### Pros
More separation, smaller diffs. We would only have to rely on SRP for the period of working on new changes to proto files.

### Cons
More detached work, changes to proto files less visible. Process of updating a separate repo is more time consuming.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-rpc/README.md
+++ b/cardano-rpc/README.md
@@ -2,7 +2,7 @@
 
 ## What's this
 
-The `cardano-rpc` package provides client and server haskell modules for gRPC interface of `cardano-node`. 
+The `cardano-rpc` package provides client and server haskell modules for gRPC interface of `cardano-node`.
 It implements [UTxO RPC](https://utxorpc.org/introduction) protobuf communication protocol specification.
 
 ## Building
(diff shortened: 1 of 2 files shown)


Jimbo4350 (commented):
LGTM. You need to fix CI and can you include some brief documentation as to how the files are generated using `buf`?

PR #966, opened 2025-09-30
Typedoc for cardano wasm

palas:
# Changelog

```yaml
- description: |
    Added nix output to generate `typedoc` documentation for `cardano-wasm`.
  type:
  - feature
  - documentation
  projects:
  - cardano-wasm
```

# Context

As part of the effort to making `cardano-wasm` accessible, this PR adds an output to the nix flake so that the documentation for the `cardano-wasm` library is easy to generate.

I also had to move `nix/npm-deps/*` → `nix/proto-to-js-npm-deps`, because there are two versions now. We could combine them in one, but this way we keep the nix outputs independent which can be handy.

# How to trust this PR

You can check the output works. Other than that, I think it is a matter of ensuring the code is well designed and structured.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/nix/proto-to-js-npm-deps/package-lock.json
+++ b/nix/proto-to-js-npm-deps/package-lock.json
@@ -5,10 +5,8 @@
   "packages": {
     "": {
       "dependencies": {
-        "google-protobuf": "^3.21.4"
-      },
-      "devDependencies": {
-        "grpc-web": "^2.0.2"
+        "google-protobuf": "^3.21.4",
+        "grpc-web": "^1.5.0"
       }
     },
     "node_modules/google-protobuf": {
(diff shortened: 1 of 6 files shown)


flake.nix
```diff
@@ -249,11 +250,12 @@
...
                   proto-js-bundle = proto-js-bundle-drv;
+                  wasm-typedoc = wasm-typedoc-drv;
                 };
             }
-            // lib.optionalAttrs (system != "aarch64-darwin")
-            {
-              packages = {
+            // {
+              packages = { wasm-typedoc = wasm-typedoc-drv; } //
+              lib.optionalAttrs (system != "aarch64-darwin") {
```
carbolymer, line 258: can you `nix fmt` this file?

palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/966/commits/d3cd04da6d7a796e4495cccfa84d53e29f564d9c

PR #995, opened 2025-10-30
Create JS independent WASM library `cardano-wasi`

palas:
# Changelog

```yaml
- description: |
    Create JS independent WASM library (`cardano-wasi`)
  type:
  - feature
  projects:
  - cardano-wasm
```

# Context

This PR exports the same functions as `cardano-wasm` (with the exception of GRPC ones, and with the addition of some memory management primitives) as ccalls so that they can be used from other languages (not just JS).

# How to trust this PR

I've tested it using `wasmer` in Rust, but please review the code to ensure I didn't make any mistakes. Ping me if you would like the Rust code to test it too, I'll add it as a test to the CI in a future PR.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-wasm/app-wasi/Main.hs
+++ b/cardano-wasm/app-wasi/Main.hs
@@ -1,0 +1,4 @@
+module Main where
+
+main :: IO ()
+main = return ()
--- a/cardano-wasm/cardano-wasm.cabal
+++ b/cardano-wasm/cardano-wasm.cabal
@@ -69,7 +69,7 @@ executable cardano-wasi
   import: project-config
   main-is: Main.hs
   hs-source-dirs:
-    app
+    app-wasi
     src-wasi
 
   default-language: Haskell2010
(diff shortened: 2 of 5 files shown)


Jimbo4350 (changes requested):
Preemptively blocking this until I've reviewed it.

Jimbo4350 (approved):
This LGTM. I dug up some old js code I wrote ages ago:

```
// Helper: Write a JS string to WASM memory (returns pointer)
function writeCString(str, wInstance) {
    const lengthNeeded = str.length + 1;
    const ptr = wInstance.exports.malloc_bytes_hs(str.length + 1);  // +1 for null terminator
    const mem = new Uint8Array(wInstance.exports.memory.buffer);


    // Check if pointer is valid and space is sufficient
    if (ptr + lengthNeeded >= mem.length) {
        wInstance.exports.free_hs(ptr);  // Avoid leaking memory
        throw new Error(`writeCString: String '${str}' exceeds allocated WASM memory.`);
    }


    for (let i = 0; i < str.length; i++) {
        mem[ptr + i] = str.charCodeAt(i);
    }
    mem[ptr + str.length] = 0;  // Null-terminate. ptr is the first byte so we add str.length to get the byte after the last string byte
    return ptr;
}
```

Ideally we should expose functions like this from haskell to avoid the use having to manually manage memory.

PR #1010, opened 2025-11-13
Support for renaming JS methods when re-exporting

palas:
# Changelog

```yaml
- description: |
    Add support for renaming JS methods re-exporting and use the functionality to simplify wallet testnet methods
  type:
  - breaking
  projects:
  - cardano-wasm
```

# Context

See https://github.com/IntersectMBO/cardano-api/pull/1003#discussion_r2505636792. Names of functions exported by the API need to be unique, but we can rename them as we re-export them for the JS API function.

# How to trust this PR

The tests passing gives strong guarantee the renaming is done correctly. Other than that, it could be that there is some file that wasn't updated that we missed. And also it would be a matter of checking whether the comments and code design is sensible.

I would look at the commit separately too, because the first one is a refactoring, and the second one is a simple change.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-wasm/src-lib/Cardano/Wasm/Api/Info.hs
+++ b/cardano-wasm/src-lib/Cardano/Wasm/Api/Info.hs
@@ -129,7 +129,9 @@ data MethodInfo = MethodInfo
   { methodName :: String
   -- ^ Name of the global method name in the API (which should match the exported function and it must be unique globally).
   , methodSimpleName :: Maybe String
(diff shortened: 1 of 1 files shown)


palas:
> In the methodParams field is it possible to encode the idea of sum type? I.e require a network magic number or default to mainnet?

@Jimbo4350, the easiest we could do is have a nullable `magic number` parameter (if it is the last parameter I think it could also be optional), and if you don't provide it we can default to mainnet. Still we can want the renaming functionality. Because that way we can have methods with the same name in different objects, like `toCbor`, otherwise we'll need `txToCbor` and `certToCbor`, because they are in the same namespace. This will still be the case at the WASM level, but we can structure them better for JS

Jimbo4350 (approved):

PR #1048, opened 2025-12-05
Revert "refactor(api): migrate ToJSON instances to use experimental Era type"

Jimbo4350:
This reverts commit 21408c8fc897d13e64aedda4a622bd5962d4fe15.

# Changelog

```yaml
- description: |
    The reverted PR breaks the query utxo command in cardano-cli which is supposed to be compatible for all shelley based eras.
  type:
  - bugfix         
  projects:
  - cardano-api
```

# Context

# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/test/cardano-api-test/Test/Cardano/Api/Json.hs
+++ b/cardano-api/test/cardano-api-test/Test/Cardano/Api/Json.hs
@@ -21,9 +21,6 @@ import Hedgehog.Gen qualified as Gen
 import Test.Tasty (TestTree, testGroup)
 import Test.Tasty.Hedgehog (testProperty)
 
-testEra :: ShelleyBasedEra ConwayEra
-testEra = ShelleyBasedEraConway
-
 prop_json_roundtrip_alonzo_genesis :: Property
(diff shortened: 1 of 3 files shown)


cardano-api/test/cardano-api-test/Test/Cardano/Api/Json.hs
```diff
@@ -21,37 +21,34 @@ import Hedgehog.Gen qualified as Gen
...
-
 prop_json_roundtrip_alonzo_genesis :: Property
 prop_json_roundtrip_alonzo_genesis = H.property $ do
   genesis <- forAll genAlonzoGenesis
   tripping genesis encode eitherDecode
 
 prop_json_roundtrip_utxo :: Property
 prop_json_roundtrip_utxo = H.property $ do
-  utxo <- forAll $ genUTxO testEra
+  utxo <- forAll $ genUTxO ShelleyBasedEraBabbage
```
carbolymer: I'm aware that this was just a quick revert, but the previous version was better. Actually, we should make sure those instances rountrip for all eras.

PR #1050, opened 2025-12-09
Node 10.7 integration

geo2a:
# Changelog

```yaml
- description: |
    Integrate new Ledger and Consensus packages for Node 10.7.
  type:
  - breaking
  projects:
  - cardano-api
```

# Context

This PR brings the new Ledger and Consensus packages and attempts to integrate them into cardano-api.

# How to trust this PR

Review the dependency bumps and any type-level changes required to accommodate the new Ledger/Consensus APIs.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/.github/workflows/haskell-wasm.yml
+++ b/.github/workflows/haskell-wasm.yml
@@ -114,9 +114,9 @@ jobs:
 
     - name: Restore cached deps
       run: |
-        wget "https://agrius.feralhosting.com/palas/wasm-cache/3cb3dd8c0e361bcde820ee8c1eb1270346b54c06024aaa7500c4da2a74e2be79.tar.xz"
-        tar -xf 3cb3dd8c0e361bcde820ee8c1eb1270346b54c06024aaa7500c4da2a74e2be79.tar.xz
-        rm 3cb3dd8c0e361bcde820ee8c1eb1270346b54c06024aaa7500c4da2a74e2be79.tar.xz
+        wget "https://agrius.feralhosting.com/palas/wasm-cache/d1a007cb07f43d2418e4d177c7e7dffc25b30cda49c50e669a1672ffadf42fe1.tar.xz"
+        tar -xf d1a007cb07f43d2418e4d177c7e7dffc25b30cda49c50e669a1672ffadf42fe1.tar.xz
+        rm d1a007cb07f43d2418e4d177c7e7dffc25b30cda49c50e669a1672ffadf42fe1.tar.xz
         rm -fr ~/.ghc-wasm/.cabal/store/
         mv store ~/.ghc-wasm/.cabal/
 


cardano-api/src/Cardano/Api/Era/Internal/Eon/AlonzoEraOnwards.hs
```diff
@@ -105,13 +104,11 @@ type AlonzoEraOnwardsConstraints era =
...
   , L.EraTxOut (ShelleyLedgerEra era)
   , L.EraUTxO (ShelleyLedgerEra era)
-  , L.HashAnnotated (L.TxBody (ShelleyLedgerEra era)) L.EraIndependentTxBody
+  , L.HashAnnotated (L.TxBody L.TopTx (ShelleyLedgerEra era)) L.EraIndependentTxBody
   , L.MaryEraTxBody (ShelleyLedgerEra era)
-  , L.NativeScript (ShelleyLedgerEra era) ~ L.Timelock (ShelleyLedgerEra era)
   , Plutus.EraPlutusContext (ShelleyLedgerEra era)
   , L.Script (ShelleyLedgerEra era) ~ L.AlonzoScript (ShelleyLedgerEra era)
   , L.ScriptsNeeded (ShelleyLedgerEra era) ~ L.AlonzoScriptsNeeded (ShelleyLedgerEra era)
-  , L.ShelleyEraTxCert (ShelleyLedgerEra era)
```
carbolymer: I'm not 100% convinced about this change. Depending on the legacy certificates support, we might need to revert this and split `AlonzoEraOnwardsConstraints` into two:
1. from Alonzo to Conway including `ShelleyEraTxCert`
2. from Dijkstra onwards.

This needs to be double checked and if anything pops out during `cardano-cli` integration.

cardano-api/src/Cardano/Api/Query/Internal/Type/QueryInMode.hs
```diff
@@ -613,11 +619,17 @@ toConsensusQueryShelleyBased sbe = \case
           (Consensus.GetCBOR (Consensus.GetPoolState (Set.map unStakePoolKeyHash <$> poolIds)))
       )
   QueryStakeSnapshot mPoolIds ->
-    Some
-      ( consensusQueryInEraInMode
-          era
-          (Consensus.GetCBOR (Consensus.GetStakeSnapshots (fmap (Set.map unStakePoolKeyHash) mPoolIds)))
+    caseShelleyToBabbageOrConwayEraOnwards
```
geo2a: @lehins I'm integrating the move of the `GetStakeSnapshot` query implementation from Consensus to Ledger, and it seems a little more involved that you anticipated.

Specifically, I've had to introduce this change, because I've started seeing type errors about era bound. I think this is because you've added the `SL.EraGov era` in Consensus.

If I understand correctly, it means that this query would not be available in any era anymore, but rather only from Conway. This does not seem like a problem to me, as Mainnet is already in Conway. However, does it mean that we're loosing the capability to issue this query to a syncing node until it has reached Conway?

cardano-api/src/Cardano/Api/Tx/Internal/Body.hs
```diff
@@ -462,7 +463,7 @@ defaultTxFee w = TxFeeExplicit w mempty
 data TxValidityUpperBound era where
   TxValidityUpperBound
     :: ShelleyBasedEra era
-    -> Maybe SlotNo
+    -> StrictMaybe SlotNo
```
Jimbo4350: Same here why break the api?

cardano-api/src/Cardano/Api/Query/Internal/Type/QueryInMode.hs
Jimbo4350: I resolved this

cabal.project
```diff
@@ -134,23 +125,15 @@ if arch(wasm32)
   source-repository-package
     type: git
     location: https://github.com/palas/mempack.git
-    tag: 0211addbbbf51011e5348d3696566eb12ccbef07
-    --sha256: sha256-iLc+foF2AM3vG6deuZ51+faI6buMkubMP75md51hMe8=
+    tag: 3514811b33741c4ac37d6f78b01f5d33088f6303
```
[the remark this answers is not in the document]
palas: You are right, I've removed it in: https://github.com/IntersectMBO/cardano-api/pull/1050/changes/7f9300e57448c56fde6d572e39cc9da0ed1ba3de

PR #1051, opened 2025-12-09
Remove duplicated orphans and re-import testlibs

palas:
# Changelog

```yaml
- description: |
    Remove duplicated `Arbitrary` orphans and re-import `testlib`s
  type:
    - compatible
  projects:
    - cardano-api
```

# Context

There is an issue with overlapping `Arbitrary` instances between the `cardano-api` and `cardano-ledger-core` packages. This is derived from an effort to prevent licensing incompatibility issues in PR: https://github.com/IntersectMBO/cardano-api/pull/942. This effort was deemed unnecessary for testing and `cardano-api:gen` packages, so this PR effectively reverts that effort.

# How to trust this PR

It should be a refactoring, and it should only have effect on the instances exported. It would be good to test whether the issue of overlapping `Arbitrary` instances disappeared.

# Checklist

- [x] Self-reviewed the diff


cardano-api/test/cardano-api-test/Test/Cardano/Api/Orphans.hs
```diff
@@ -1,29 +1,20 @@
...
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
-{-# LANGUAGE TypeApplications #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
 
 module Test.Cardano.Api.Orphans () where
 
+-- Signing Key instances
```
jasagredo: I think this line is duplicated?

PR #1101, opened 2026-02-06
Bump plutus to 1.58

zliu41:
# Changelog

```yaml
- description: |
    Bumped `plutus` to `1.58`
  type:
    - maintenance
  projects:
    - cardano-api
```

### DIFF
--- a/.github/workflows/haskell-wasm.yml
+++ b/.github/workflows/haskell-wasm.yml
@@ -114,9 +114,9 @@ jobs:
 
     - name: Restore cached deps
       run: |
-        wget "https://agrius.feralhosting.com/palas/wasm-cache/96feb8056220c74aa6b5af41ca64ad982f207fc4f99a22e0393a4de12169bf04.tar.xz"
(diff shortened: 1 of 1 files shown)


palas:
Oh, in fact I am getting this error int the wasm test. I must have done something wrong in the patch:
```
RuntimeError: Expected Right, got Left: "Error in $.costModels: CMInternalWriteError \"parsing Int failed, value is either floating or will cause over or underflow 9.223372036854775807e18\""
CallStack (from HasCallStack):
  rightOrError, called at src-lib/Cardano/Wasm/Api/Tx.hs:194:25 in cardano-wasm-10.0.0.0-inplace-cardano-wasi-lib:Cardano.Wasm.Api.Tx
  estimateMinFeeImpl, called at src-wasm
```

palas:
I think I found the culprit: https://github.com/IntersectMBO/plutus/blob/0ddb89848cfd7f23199e23b7d3cb5a090054bea4/plutus-core/plutus-core/src/PlutusCore/Evaluation/Machine/ExMemory.hs#L67-L68

palas (approved):
It works now 👍

PR #1104, opened 2026-02-09
Bump network and consensus for node-10.6.2

zliu41:
# Changelog

```yaml
- description: |
    Bumped `network` and `consensus`
  type:
    - maintenance
  projects:
    - cardano-api
```

### DIFF
--- a/.github/workflows/haskell-wasm.yml
+++ b/.github/workflows/haskell-wasm.yml
@@ -114,9 +114,9 @@ jobs:
 
     - name: Restore cached deps
       run: |
-        wget "https://agrius.feralhosting.com/palas/wasm-cache/14e3f2c1a5d629de5bade4329dc419fd2cfcc7914447f07e02126b28173a522c.tar.xz"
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -46,7 +46,7 @@
     defaultCompiler = "ghc9102";
     # Used for cross compilation, and so referenced in .github/workflows/release-upload.yml. Adapt the
     # latter if you change this value.
-    crossCompilerVersion = "ghc967";
+    crossCompilerVersion = "ghc9122";
```
carbolymer: Why are we switching to GH912 for cross-compilation? We are still building `cardano-node` using 9.6.7 so it is beneficial to have this in CI. Otherwise you will only catch windows cross-compilation issues at the integration stage in cardano-node.

carbolymer (changes requested):
>@carbolymer, in order for the latest ouroboros libraries versions to compile to mingw32, I had to change the mingw32 compiler version to ghc9122. But that breaks cardano-rpc, as you know.

But o-c and o-n are both using GHC 9.6 on windows in CI without issues. What error does this solve?

PR #1179, opened 2026-04-09
Add FromJSON instance for new experimental TxOut

### LINKED ISSUE #926: Refactor `FromJSON (TxOut CtxTx era)`
The `FromJSON (TxOut CtxTx era)` instance is a mess and needs to be refactored.

Jimbo4350:
# Context

Implements `FromJSON` instances for the new `TxOut era` type (which wraps `L.TxOut era`) in the experimental API, matching the JSON format produced by the `ToJSON` instances added in #1176.

Closes #926

# Changelog

```yaml
- description: |
    Add FromJSON instance for new experimental TxOut type with per-era
    concrete instances, inline datum parsing from inlineDatumRaw with
    hash validation, and reference script support via scriptInAnyLangToLedgerScript
  type:
   - feature
  projects:
   - cardano-api
```

# How to trust this PR

- Concrete `FromJSON` instances for each era (Shelley through Conway), mirroring the `ToJSON` structure from #1176
- Pre-Alonzo eras parse address and value only via shared `txOutBaseParseJson` helper
- Alonzo adds datum hash parsing via `dataHashTxOutL`
- Babbage+ adds inline datum parsing from `inlineDatumRaw` (hex-encoded original CBOR bytes) with hash validation against `inlineDatumhash`, and reference script parsing via `scriptInAnyLangToLedgerScript`
- Supplemental datums deliberately unsupported — the ledger TxOut does not carry them
- Helper functions: `addrFromJson` (reverse of `addrToJson`), `scriptInAnyLangToLedgerScript` (reverse of `ledgerScriptToScriptInAnyLang`)
- Property test verifies JSON round-trip (`fromJSON . toJSON = id`) across all Shelley-based eras

# Checklist

- [x] Commit sequence is logical and each commit has a good commit message
- [x] New tests are added if needed and old tests still pass
- [x] Code is formatted with `fourmolu`
- [x] Self-reviewed the diff

### DIFF
--- a/.changes/20260420_cardano_api_fromjson_experimental_txout.yml
+++ b/.changes/20260420_cardano_api_fromjson_experimental_txout.yml
@@ -3,4 +3,4 @@ pr: 1179
 kind:
   - feature
 description: |
-  Add FromJSON instances for the new experimental TxOut type, matching the JSON format produced by the corresponding ToJSON instances. Inline datums are parsed from inlineDatumRaw with hash validation against inlineDatumhash. Reference scripts are parsed via the existing FromJSON ScriptInAnyLang instance and converted to ledger scripts. Supplemental datums are deliberately unsupported since the ledger TxOut does not carry them.
+  Add `FromJSON` instances for the new experimental `TxOut` type.
--- a/.changes/20260420_cardano_api_fromjson_experimental_txout.yml
+++ b/.changes/20260420_cardano_api_fromjson_experimental_txout.yml
@@ -1,0 +1,6 @@
+project: cardano-api
+pr: 1179
+kind:
+  - feature
+description: |
+  Add FromJSON instances for the new experimental TxOut type, matching the JSON format produced by the corresponding ToJSON instances. Inline datums are parsed from inlineDatumRaw with hash validation against inlineDatumhash. Reference scripts are parsed via the existing FromJSON ScriptInAnyLang instance and converted to ledger scripts. Supplemental datums are deliberately unsupported since the ledger TxOut does not carry them.
--- a/cardano-api/src/Cardano/Api/Experimental/Tx/Internal/BodyContent/New.hs
+++ b/cardano-api/src/Cardano/Api/Experimental/Tx/Internal/BodyContent/New.hs
@@ -74,6 +74,7 @@ module Cardano.Api.Experimental.Tx.Internal.BodyContent.New
 where
 
 import Cardano.Api.Address
+import Cardano.Api.Era.Internal.Eon.ShelleyBasedEra (ShelleyBasedEra (..), ShelleyLedgerEra)
 import Cardano.Api.Error
 import Cardano.Api.Experimental.AnyScriptWitness
 import Cardano.Api.Experimental.Certificate qualified as Exp
@@ -102,7 +103,7 @@ import Cardano.Api.Governance.Internal.Action.VotingProcedure
 import Cardano.Api.Key.Internal
 import Cardano.Api.Ledger.Internal.Reexport (StrictMaybe (..))
 import Cardano.Api.Ledger.Internal.Reexport qualified as L
-import Cardano.Api.Monad.Error (liftMaybe)
+import Cardano.Api.Monad.Error (failEitherWith, liftMaybe)
 import Cardano.Api.Plutus.Internal.Script
   ( PlutusScript (..)
   , PlutusScriptVersion (..)
@@ -127,8 +128,7 @@ import Cardano.Api.Value.Internal
   ( PolicyAssets
   , PolicyId
   , Value
-  , fromMaryValue
-  , lovelaceToValue
+  , fromLedgerValue
   , policyAssetsToValue
   , toMaryValue
   )
@@ -163,7 +163,6 @@ import Data.Sequence.Strict qualified as Seq
 import Data.Set (Set)
 import Data.Set qualified as Set
 import Data.Text.Encoding qualified as Text
-import Data.Typeable (cast)
 import GHC.Exts (IsList (..))
 import Lens.Micro
 
@@ -374,16 +373,11 @@ eraSpecificLedgerTxBody era ledgerbody bc =
 data TxOut era where
   TxOut :: L.EraTxOut era => L.TxOut era -> TxOut era
 
--- | Pre-Alonzo eras have no datums or reference scripts, so the JSON
--- output is just address and value.
-instance ToJSON (TxOut L.ShelleyEra) where
-  toJSON (TxOut o) = txOutBaseJson o
+instance ToJSON (TxOut L.ShelleyEra) where toJSON = txOutToJson ShelleyBasedEraShelley
 
-instance ToJSON (TxOut L.AllegraEra) where
-  toJSON (TxOut o) = txOutBaseJson o
+instance ToJSON (TxOut L.AllegraEra) where toJSON = txOutToJson ShelleyBasedEraAllegra
 
-instance ToJSON (TxOut L.MaryEra) where
-  toJSON (TxOut o) = txOutBaseJson o
+instance ToJSON (TxOut L.MaryEra) where toJSON = txOutToJson ShelleyBasedEraMary
 
 -- | Note: Unlike the legacy API's @TxOut@, this instance does not render
 -- supplemental datums. At the ledger level, a supplemental datum is not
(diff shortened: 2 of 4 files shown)


cardano-api/src/Cardano/Api/Experimental/Tx/Internal/BodyContent/New.hs
```diff
@@ -489,6 +496,130 @@ deriving instance (Show (TxOut era))
...
+instance FromJSON (TxOut L.ConwayEra) where
+  parseJSON = Aeson.withObject "TxOut" babbageOnwardsTxOutParseJson
+
+-- | Parse the base fields (address and value) shared by all eras.
+txOutBaseParseJson :: L.EraTxOut era => Aeson.Object -> Parser (L.TxOut era)
+txOutBaseParseJson o = do
+  addr <- addrFromJson =<< o .: "address"
+  apiVal <- parseJSON =<< o .: "value"
+  let mv = toMaryValue apiVal
+  val <- case cast mv of
```
palas: I think this usage of `cast` could use a comment explaining how it works

cardano-api/src/Cardano/Api/Experimental/Tx/Internal/BodyContent/New.hs
```diff
@@ -489,6 +496,130 @@ deriving instance (Show (TxOut era))
...
+    Just v -> pure v
+    Nothing -> case cast (L.coin mv) of
+      Just v -> pure v
+      Nothing -> fail "txOutBaseParseJson: value is unsupported for this era"
```
palas:
```suggestion
    Nothing -> do
      let L.MaryValue _ ma = mv
      unless (ma == mempty) $
        fail "txOutBaseParseJson: ada-only era output cannot carry a multi-asset value"
      case cast (L.coin mv) of
        Just v -> pure v
        Nothing -> fail "txOutBaseParseJson: value is unsupported for this era"
```
I think it would be for the best to check there are no non-ada assets in the case of pre-mary eras

cardano-api/src/Cardano/Api/Experimental/Tx/Internal/BodyContent/New.hs
```diff
@@ -489,6 +496,130 @@ deriving instance (Show (TxOut era))
...
+            <> show expectedHash
+            <> ", got "
+            <> show (L.hashBinaryData binaryData)
+      pure $ L.Datum binaryData
+    Nothing -> do
+      when (isJust mInlineDatumHash) $
+        fail "babbageOnwardsTxOutParseJson: inlineDatumhash present without inlineDatumRaw"
+      case mDatumHash of
+        Just dh -> pure $ L.DatumHash dh
+        Nothing -> pure L.NoDatum
```
palas:
```suggestion
      maybe L.NoDatum L.DatumHash mDatumHash
```
Maybe a good place to use the [maybe](https://hackage-content.haskell.org/package/base-4.22.0.0/docs/Prelude.html#v:maybe) function.

palas (approved):
Looks good. I just made a couple of small suggestions.

`inlineDatumhash` is really weird, both the capitalisation and it being omitted on Babbage when it inline datum is not present (instad of just being `null`). But it is outside of the scope of the PR

.changes/20260420_cardano_api_fromjson_experimental_txout.yml
```diff
@@ -0,0 +1,12 @@
...
+kind:
+  - feature
+description: |
+  Add FromJSON instances for the new experimental TxOut type, matching
+  the JSON format produced by the corresponding ToJSON instances. Inline
+  datums are parsed from inlineDatumRaw with hash validation against
+  inlineDatumhash. Reference scripts are parsed via the existing
+  FromJSON ScriptInAnyLang instance and converted to ledger scripts.
+  Supplemental datums are deliberately unsupported since the ledger
+  TxOut does not carry them.
```
carbolymer: redundant line wraps - they will leak into the final changelog

cardano-api/src/Cardano/Api/Experimental/Tx/Internal/BodyContent/New.hs
carbolymer: I think we don't need cast. This is why we have eons. `cast` is a bit abused here, because instead of type equality proof, you are trying and failing, we trying again and failing.... With eons, you only have to deal with already known cases. If new era arrives you will get compile time error.

The problem with cast is the final "else" case: when new era gives us new type equality (not supported here yet) you will get runtime error instead of compile-time error.

cardano-api/src/Cardano/Api/Experimental/Tx/Internal/BodyContent/New.hs
```diff
@@ -489,6 +496,130 @@ deriving instance (Show (TxOut era))
...
+  mRefScript <- o .:? "referenceScript"
+  -- Determine datum
+  datum <- case mInlineDatumRaw of
+    Just rawHex -> do
+      expectedHash <- case mInlineDatumHash of
+        Nothing -> fail "babbageOnwardsTxOutParseJson: inlineDatumRaw present without inlineDatumhash"
+        Just h -> pure h
+      rawBytes <- case Base16.decode (Text.encodeUtf8 rawHex) of
+        Left err -> fail $ "babbageOnwardsTxOutParseJson: failed to hex-decode inlineDatumRaw: " <> show err
+        Right bs -> pure bs
```
carbolymer: The explicit case matching on Maybe and Either repeats itself too many times in this PR. I think this warrants creating helper functions to make the code more readable.

We have `failEitherWith` for example:
```suggestion
      rawBytes <- failEitherWith (("babbageOnwardsTxOutParseJson: failed to hex-decode inlineDatumRaw: " <>) . show) $ 
                    Base16.decode (Text.encodeUtf8 rawHex)
```

cardano-api/src/Cardano/Api/Experimental/Tx/Internal/BodyContent/New.hs
```diff
@@ -489,6 +496,130 @@ deriving instance (Show (TxOut era))
...
+    Nothing -> do
+      when (isJust mInlineDatumHash) $
+        fail "babbageOnwardsTxOutParseJson: inlineDatumhash present without inlineDatumRaw"
+      case mDatumHash of
+        Just dh -> pure $ L.DatumHash dh
+        Nothing -> pure L.NoDatum
+  -- Determine reference script
+  refScript <- case mRefScript of
+    Nothing -> pure SNothing
+    Just script -> SJust <$> scriptInAnyLangToLedgerScript script
```
carbolymer:
```suggestion
  refScript <- fmap maybeToStrictMaybe $ forM mRefScript scriptInAnyLangToLedgerScript 
```

cardano-api/src/Cardano/Api/Experimental/Tx/Internal/BodyContent/New.hs
```diff
@@ -489,6 +496,130 @@ deriving instance (Show (TxOut era))
...
+  -- Construct TxOut
+  pure . TxOut $
+    baseTxOut
+      & L.datumTxOutL .~ datum
+      & L.referenceScriptTxOutL .~ refScript
+
+-- | Convert a 'ScriptInAnyLang' to a ledger 'L.Script'. Reverse of 'ledgerScriptToScriptInAnyLang'.
```
carbolymer: This function should be placed near its reverse.

carbolymer (changes requested):
I think usage of cast is imprecise here and will be an annoyance down the road. I think we should stop for a moment and decide how we would like to support legacy eras using experimental API.

Moreover, PR has a lot of duplication from pattern matching on Maybe and Either.

carbolymer (approved):

.changes/20260420_cardano_api_fromjson_experimental_txout.yml
```diff
@@ -0,0 +1,6 @@
+project: cardano-api
+pr: 1179
+kind:
+  - feature
+description: |
+  Add FromJSON instances for the new experimental TxOut type, matching the JSON format produced by the corresponding ToJSON instances. Inline datums are parsed from inlineDatumRaw with hash validation against inlineDatumhash. Reference scripts are parsed via the existing FromJSON ScriptInAnyLang instance and converted to ledger scripts. Supplemental datums are deliberately unsupported since the ledger TxOut does not carry them.
```
carbolymer: Tbh this is too detailed for user-facing changelog.

cardano-api/src/Cardano/Api/Experimental/Tx/Internal/BodyContent/New.hs
```diff
@@ -485,10 +484,144 @@ ledgerScriptToScriptInAnyLang script =
...
+    ShelleyBasedEraAlonzo -> do
+      let base = L.mkBasicTxOut addr mv
+      mDatumHash <- o .:? "datumhash"
+      pure . TxOut $ case mDatumHash of
+        Nothing -> base
+        Just dh -> base & L.dataHashTxOutL .~ SJust dh
+    ShelleyBasedEraBabbage ->
+      babbageOnwardsTxOutParseJson (L.mkBasicTxOut addr mv :: L.TxOut L.BabbageEra) o
+    ShelleyBasedEraConway ->
+      babbageOnwardsTxOutParseJson (L.mkBasicTxOut addr mv :: L.TxOut L.ConwayEra) o
```
carbolymer:
```suggestion
      babbageOnwardsTxOutParseJson (L.mkBasicTxOut addr mv) o
    ShelleyBasedEraConway ->
      babbageOnwardsTxOutParseJson (L.mkBasicTxOut addr mv) o
```

PR #1180, opened 2026-04-09
Fix broken cross-package Haddock links on docs site

### LINKED ISSUE #601: [BUG] - CHaP packages links in Haddocks are broken
**Internal/External**

**Area**
*Other* 

**Summary**
This is related to: https://github.com/IntersectMBO/cardano-api/issues/437

There are still invalid links in the Haddock pages
* On this page https://cardano-api.cardano.intersectmbo.org/cardano-api/internal/Cardano-Api-Address.html#t:Address the link to the ledger type `Address` https://cardano-api.cardano.intersectmbo.org/cardano-ledger-byron/Cardano-Chain-Common.html#t:Address results in HTTP 404
* On this page https://cardano-api.cardano.intersectmbo.org/cardano-api/Cardano-Api-Shelley.html#v:ShelleyAddress in this code part:
   ```
  ShelleyAddress
    :: Shelley.Network
    -> Shelley.PaymentCredential StandardCrypto
    -> Shelley.StakeReference StandardCrypto
    -> Address ShelleyAddr
   ```
   * `Shelley.PaymentCredential` points to `https://cardano-api.cardano.intersectmbo.org/cardano-ledger-core/Cardano-Ledger-Credential.html#t:PaymentCredential` resulting in 404 error

Jimbo4350:
# Changelog

```yaml
- description: |
    Fix broken cross-package Haddock links on the hosted documentation site.
    Links to dependency packages (cardano-ledger-*, plutus-*, cardano-base,
    etc.) were relative paths pointing to directories that don't exist on the
    site, resulting in 404s. A post-processing script now rewrites these to
    absolute URLs pointing at the correct external documentation hosts, or to
    annotated unclickable spans where no upstream doc site exists.
  type:
   - bugfix
   - documentation
  projects:
   - cardano-api
```

# Context

Fixes #601. The hosted Haddock site at cardano-api.cardano.intersectmbo.org only contains docs for packages in this repo. When Haddock generates cross-references to dependency types, it produces relative links like `../cardano-ledger-core/Cardano-Ledger-Credential.html` — but those directories don't exist on the site.

The `cabal haddock-project --html-location` flag only accepts a single URL template, but our dependencies are spread across multiple hosts. The post-processing script `scripts/fix-haddock-links.sh` instead resolves each cross-package href and rewrites it after haddock generation.

Resolution per CHaP package, first hit wins:

1. **Name-suffix heuristic** under `*.cardano.intersectmbo.org` — strip trailing `-token` segments and HEAD-probe each candidate's `doc-index.html`. Catches `cardano-ledger-*`, `plutus-*`, `ouroboros-*`, etc.

   *Worked example* — `cardano-ledger-api`:
   - candidate 1: `https://cardano-ledger-api.cardano.intersectmbo.org/cardano-ledger-api/doc-index.html` → 404 (no such subdomain)
   - candidate 2: `https://cardano-ledger.cardano.intersectmbo.org/cardano-ledger-api/doc-index.html` → **200** ✓

   Package resolves to base `https://cardano-ledger.cardano.intersectmbo.org`. Probing stops at the first 200; we never need candidate 3 (`cardano`).

2. **`IOG_DOC_BASES` fallback** — known doc-site roots for irregular subdomains (e.g. `cardano-base` lives at `base.cardano.intersectmbo.org`).

Hrefs that don't resolve become annotated unclickable `<span>`s with tooltips. Non-CHaP packages (base, bytestring, etc.) are deliberately not linked — Haddock's URL shapes don't line up cleanly with Hackage. Module-level 404s at otherwise-valid sites are rescued via parent-module fallback where possible (`Foo-Bar.html` → `Foo.html#t:Bar`); otherwise also become spans. Full design and CI policy (actionable vs unfixable buckets, `KNOWN_UNDOCUMENTED` allowlist) documented in the script preamble.

# Tracking-issue automation

When the workflow's `Fix cross-package Haddock links` step fails on master, a follow-up GH Actions step opens (or comments on) a single rolling tracking issue labelled `haddock-ci-failure` with the PR opener as assignee. One issue per outage period — subsequent failures append a comment instead of duplicating; closing the issue resets the cycle.

**Validation** on a separate test branch (`issue-601-test-tracking`) with a synthetic break (`typed-protocols` removed from `IOG_DOC_BASES`, master-ref gate temporarily relaxed):

- **Run 1** — no open tracking issue → step opened a new one:
  [run](https://github.com/IntersectMBO/cardano-api/actions/runs/25013918872) · [issue #1191](https://github.com/IntersectMBO/cardano-api/issues/1191)
- **Run 2** — issue still open → step commented on it, no duplicate:
  [run](https://github.com/IntersectMBO/cardano-api/actions/runs/25016887665) · [comment](https://github.com/IntersectMBO/cardano-api/issues/1191#issuecomment-4330476583)
  (`gh issue list --label haddock-ci-failure --state open` still shows only #1191.)

Test branch and test issue will be deleted/closed after merge.

# How to trust this PR

- The script only does `sed`-style rewrites on the post-haddock HTML output — it cannot affect source code or build artefacts.
- All rewritten URLs are HEAD-validated; dead ones become unclickable spans, so the published site has zero clickable 404s.
- Tracking-issue automation validated end-to-end via the runs linked above.

# Checklist

- [x] Self-reviewed the diff


scripts/fix-haddock-links.sh
```diff
@@ -0,0 +1,98 @@
...
+#   ../base/Data-Bool.html -> https://hackage.haskell.org/package/base/docs/Data-Bool.html
+# The package name directory is replaced by /package/<pkg>/docs/
+HACKAGE_PACKAGES=(
+  base
+  bytestring
+  containers
+  deepseq
+  mtl
+  ghc-internal
+)
```
carbolymer, line 160: This fix is unclear to me. Why only those packages?

Additionally, I don't see broken links to those packages. But for example all links to `proto-lens` are broken. This seems rather ad-hoc rather than a proper fix/workaround.

scripts/fix-haddock-links.sh
```diff
@@ -0,0 +1,98 @@
...
+
+  # ouroboros-consensus repo packages
+  "ouroboros-consensus|https://ouroboros-consensus.cardano.intersectmbo.org/haddocks"
+
+  # ouroboros-network repo packages
+  "cardano-diffusion|https://ouroboros-network.cardano.intersectmbo.org"
+
+  # io-sim repo packages
+  "io-classes|https://input-output-hk.github.io/io-sim"
+
```
carbolymer, line 140:
```suggestion
```

carbolymer (changes requested):

.github/workflows/github-page.yml
```diff
@@ -39,6 +39,10 @@ jobs:
           mkdir website
           cabal haddock-project --output=./website --internal --foreign-libraries
 
+      - name: Fix cross-package Haddock links
+        run: |
+          ./scripts/fix-haddock-links.sh ./website
```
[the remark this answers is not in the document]
Jimbo4350: Fixed in 0396d5f3e — added `.changes/20260428_cardano_api_fix_haddock_links.yml`.

scripts/fix-haddock-links.sh
```diff
@@ -0,0 +1,748 @@
...
+    [[ "$name" == *-* ]] || break
+    name="${name%-*}"
+  done
+}
+
+probe_site() {
+  local code
+  code=$(curl -sI -o /dev/null -w "%{http_code}" --connect-timeout 5 --max-time 10 \
+    "${1}/${2}/doc-index.html" 2>/dev/null || echo "000")
+  [[ "$code" == "200" || "$code" == "301" || "$code" == "302" ]]
```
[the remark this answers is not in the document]
Jimbo4350: Fixed in 0396d5f3e — `probe_site` plus the Phase 2b probe, Phase 3 main validator, and Phase 3a rescue probe now all accept 200/301/302/307/308.

scripts/fix-haddock-links.sh
```diff
@@ -0,0 +1,748 @@
...
+# GitHub Actions, actionable entries emit ::warning:: annotations so
+# they surface in the job UI; unfixable ones are plain log lines.
+
+set -euo pipefail
+
+WEBSITE_DIR="${1:?Usage: $0 <website-directory>}"
+[ -d "$WEBSITE_DIR" ] || { echo "Error: $WEBSITE_DIR is not a directory" >&2; exit 1; }
+
+TMPFILES=()
+trap 'rm -f "${TMPFILES[@]}"' EXIT
```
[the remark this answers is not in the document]
Jimbo4350: Skipping — non-issue on this CI: GNU `rm -f` with no operands silently exits 0, and bash 4.4+ doesn't trip `set -u` on empty-array expansion. The trap is safe on ubuntu-latest, which is where this workflow runs.

scripts/fix-haddock-links.sh
```diff
@@ -0,0 +1,748 @@
...
+  local code
+  code=$(curl -sI -o /dev/null -w "%{http_code}" --connect-timeout 5 --max-time 10 \
+    "${1}/${2}/doc-index.html" 2>/dev/null || echo "000")
+  [[ "$code" == "200" || "$code" == "301" || "$code" == "302" ]]
+}
+
+resolve_url() {
+  local pkg="$1" is_chap="$2" base
+  if [[ "$is_chap" != "yes" ]]; then echo "HACKAGE"; return; fi
+  while IFS= read -r base; do
```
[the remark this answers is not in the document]
Jimbo4350: Already addressed — PR description was updated yesterday to say non-CHaP packages are deliberately not linked, matching the implementation.

scripts/fix-haddock-links.sh
```diff
@@ -0,0 +1,98 @@
...
+
+WEBSITE_DIR="${1:?Usage: $0 <website-directory>}"
+
+if [ ! -d "$WEBSITE_DIR" ]; then
+  echo "Error: $WEBSITE_DIR is not a directory" >&2
+  exit 1
+fi
+
+# Mapping of package-name-prefix -> base URL
+# Each entry is: "prefix|base_url"
```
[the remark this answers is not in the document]
Jimbo4350: Obsolete — this comment was on an earlier version of the script that had a hardcoded mapping table. Resolution is now via the name-suffix heuristic + `IOG_DOC_BASES` fallback (see commit 0cc7cf13d and the script preamble).

scripts/fix-haddock-links.sh
```diff
@@ -0,0 +1,98 @@
...
+done
+
+for pkg in "${HACKAGE_PACKAGES[@]}"; do
+  SED_ARGS+=(-e "s|href=\"\\.\\./${pkg}/|href=\"https://hackage.haskell.org/package/${pkg}/docs/|g")
+  SED_ARGS+=(-e "s|href=\"\\.\\./\\.\\./${pkg}/|href=\"https://hackage.haskell.org/package/${pkg}/docs/|g")
+done
+
+echo "Fixing cross-package Haddock links in $WEBSITE_DIR..."
+
+find "$WEBSITE_DIR" -name '*.html' -print0 | xargs -0 -P "$(nproc)" sed -i "${SED_ARGS[@]}"
```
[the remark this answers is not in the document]
Jimbo4350: Skipping — this script runs only on `ubuntu-latest` from `.github/workflows/github-page.yml`; cardano-api's CI is Linux-only and the nix devShell is Linux too. macOS portability would be defensive churn for a use case we don't have.

.changes/20260428_cardano_api_fix_haddock_links.yml
```diff
@@ -0,0 +1,17 @@
+project: cardano-api
+pr: 1180
+kind:
+  - bugfix
+  - documentation
+description: |
+  Fix broken cross-package Haddock links on the hosted documentation site.
+  Links to dependency packages (cardano-ledger-*, plutus-*, cardano-base, etc.)
```
carbolymer: this line wrap needs to be removed

scripts/fix-haddock-links.sh
```diff
@@ -0,0 +1,777 @@
...
+derive_name_candidates() {
+  local name="$1"
+  while true; do
+    printf '%s\n' "https://${name}.cardano.intersectmbo.org"
+    [[ "$name" == *-* ]] || break
+    name="${name%-*}"
+  done
+}
+
+probe_site() {
```
carbolymer, line 179: Every `curl` in the script is a single-shot attempt (5s connect, 10s max). No retries anywhere.

Running the script's output through a full URL probe turned up two transient failures: `cardano-ledger-dijkstra` returned 503, `plutus-core/ExBudget` timed out. Both returned 200 on immediate retry. Phase 3 had already converted them to unclickable `<span>`s, so the published site would have lost two working links for no reason.

The blast radius depends on where the transient hits:
- Phase 2 (`probe_site`): the heuristic skips the correct candidate and tries the next suffix. If all candidates fail, the package lands in `UNMAPPED_CHAP` and fails CI.
- Phase 3 (validation): a rewritten valid URL gets permanently downgraded to a `<span>`. This is the worst case since it silently degrades the output.
- CHaP index fetch: `curl --fail` with no retry, so a CDN blip kills the whole build.

`--retry 3 --retry-delay 2 --retry-all-errors` on the curl calls would fix this. Phase 3 is where it matters most.

scripts/fix-haddock-links.sh
```diff
@@ -0,0 +1,777 @@
...
+)
+
+# Some doc sites organise module pages into subdirectories beneath the
+# package root (e.g. ouroboros-network publishes Ouroboros-Network-Magic.html
+# under ouroboros-network/api/, not ouroboros-network/ directly). When a
+# module page URL 404s at the flat path, the rescue logic retries under
+# each of these subdirectories.
+DOC_SUBDIRS=(api protocols framework)
+
+derive_name_candidates() {
```
carbolymer, line 169: This strips all the way down to single-token names. `derive_name_candidates "cardano-ledger-api"` produces `cardano.cardano.intersectmbo.org` as its last candidate. If that subdomain ever resolves (wildcard DNS, some future deployment), the heuristic wins with the wrong site silently.

Adding a second break would cap it at two-token names:

```bash
derive_name_candidates() {
  local name="$1"
  while true; do
    printf '%s\n' "https://${name}.cardano.intersectmbo.org"
    [[ "$name" == *-* ]] || break
    name="${name%-*}"
    [[ "$name" == *-* ]] || break
  done
}
```

.github/workflows/github-page.yml
```diff
@@ -65,3 +70,158 @@ jobs:
...
+              // Also add the new breaker as an assignee (idempotent — if
+              // they're already assigned the API silently no-ops).
+              const existing = issues[0];
+              await github.rest.issues.createComment({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: existing.number,
+                body: commentBody,
+              });
+              await github.rest.issues.addAssignees({
```
carbolymer: GitHub's API requires assignees to be repo collaborators. If an external contributor's PR breaks haddock-links on master, `addAssignees` returns 422 and crashes the step. Same for `assignees: [author]` in `issues.create` at line 224.

Depending on timing, the issue or comment may already have been created before the crash, leaving things half-done.

```javascript
try {
  await github.rest.issues.addAssignees({
    owner: context.repo.owner,
    repo: context.repo.repo,
    issue_number: existing.number,
    assignees: [author],
  });
} catch (e) {
  console.log(`Could not assign ${author} (may not be a collaborator): ${e.message}`);
}
```

Same treatment for the `issues.create` path: create the issue without `assignees`, then try `addAssignees` separately with the catch.

carbolymer (changes requested):
Nice work, haddocks look good now! But we need to add some improvements to the workflow and the script. Those need to be addressed:
- https://github.com/IntersectMBO/cardano-api/pull/1180/changes#r3159565460
- https://github.com/IntersectMBO/cardano-api/pull/1180/changes#r3159784547
- https://github.com/IntersectMBO/cardano-api/pull/1180/changes#r3159784102

scripts/fix-haddock-links.sh
```diff
@@ -0,0 +1,781 @@
...
+#          publish their internal modules. (It's plausible that our
+#          --internal flag in cabal haddock-project contributes; we
+#          have not verified. Dropping --internal would cost us our
+#          own internal-module pages on the published site, so we
+#          leave it on and accept the span noise.)
+#       b. Packages in KNOWN_UNDOCUMENTED (no published Haddocks
+#          anywhere, only source on GitHub — e.g. kes-agent).
+#       c. Haddock-emitted absolute Hackage URLs that lack a package
+#          version (Hackage's routing requires one). A handful, out
+#          of our control, treated as noise.
```
palas, line 93: There is another potential reason. Haddocks are for `master` branch, and `master` branch is linked to releases of other packages, so references may just not match because the symbol referenced may have disappeared, or moved since last release. We are still linking to it, but they won't be in the generated haddock in the repo of the dependency just because the haddock of the repo of the dependency is also built from `master` (not from the release we are linking, which may not even be the last one)

scripts/fix-haddock-links.sh
```diff
@@ -0,0 +1,781 @@
...
+# local re-export hrefs to point at the upstream doc site.
+
+echo "Phase 2b: Rewriting re-exported type links..."
+
+REEXPORTS_FILE=$(mktemp_tracked)
+python3 - "$WEBSITE_DIR" <<'PYEOF' > "$REEXPORTS_FILE"
+import os, re, sys
+website = sys.argv[1]
+pattern = re.compile(
+    r'id="(t|v):([^"]+)" class="def">[^<]*</a>\s*<a href="file:///[^"]*?/([a-zA-Z][\w-]*)-\d+\.[^/]*/share/doc/html/src/([^"#]+)\.html'
```
palas, line 380: This doesn't seem to be working on my machine, maybe because I get links like this:
```
file:///Users/palas/.local/state/cabal/store/ghc-9.14.1-inplace/crdn-ldgr-p-1.13.0.0-9cac6f16/share/doc/html/src/Cardano.Ledger.Api.html
```
But maybe it works on the GHA runner, which is what matters

scripts/fix-haddock-links.sh
```diff
@@ -0,0 +1,781 @@
...
+fi
+if [[ ! -s "$CHAP_PKGS_FILE" ]]; then
+  echo "Error: CHaP package index is empty (fetch succeeded but produced no entries)" >&2
+  exit 1
+fi
+declare -A CHAP_SET
+while IFS= read -r pkg; do CHAP_SET["$pkg"]=1; done < "$CHAP_PKGS_FILE"
+
+# Single HTML scan for all cross-package link targets
+DISCOVERED_PKGS=$(grep -rohP 'href="\.\./(\.\./)?\K[a-zA-Z][a-zA-Z0-9_.-]*(?=/)' \
```
palas:
```suggestion
DISCOVERED_PKGS=$(grep -rohP --include='*.html' 'href="\.\./(\.\./)?\K[a-zA-Z][a-zA-Z0-9_.-]*(?=/)' \
```

Otherwise it tries to match things that are not HTML

scripts/fix-haddock-links.sh
```diff
@@ -0,0 +1,781 @@
...
+  done < "$REEXPORTS_FILE"
+
+  # Probe all unique candidate URLs in parallel
+  REEXPORT_VALID_FILE=$(mktemp_tracked)
+  if [[ -s "$REEXPORT_CANDIDATES" ]]; then
+    # shellcheck disable=SC2016
+    awk -F'\t' '{print $4}' "$REEXPORT_CANDIDATES" | sort -u | \
+      xargs -P 16 -I{} sh -c \
+        'code=$(curl -sI -o /dev/null -w "%{http_code}" --connect-timeout 5 --max-time 10 --retry 3 --retry-delay 2 --retry-all-errors "{}"); if [ "$code" = "200" ] || [ "$code" = "301" ] || [ "$code" = "302" ] || [ "$code" = "307" ] || [ "$code" = "308" ]; then echo "{}"; fi' \
+        > "$REEXPORT_VALID_FILE" 2>/dev/null
```
palas, line 448:
```suggestion
    awk -F'\t' '{print $4}' "$REEXPORT_CANDIDATES" | sort -u | \
      xargs -P 16 -I{} sh -c \
        'code=$(curl -sI -o /dev/null -w "%{http_code}" --connect-timeout 5 --max-time 10 --retry 3 --retry-delay 2 --retry-all-errors "$1"); if [ "$code" = "200" ] || [ "$code" = "301" ] || [ "$code" = "302" ] || [ "$code" = "307" ] || [ "$code" = "308" ]; then echo "$1"; fi' _ {} \
        > "$REEXPORT_VALID_FILE" 2>/dev/null
```
This avoids injection, which is not really a worry, but I figure it is probably going to be more reliable too, if we happen to get any weird character in URLs. This issue is in more places

palas (commented):
If this works, I think it is very positive. I would merge it and give it a go. Maybe iterate if we see issues in the future. It is not critical code, so I don't think it is worth spending a huge amount of time ensuring it is bullet-proof. And `bash` is super finicky, so I think properly reviewing it would require a PhD in `bash` and a few days. So here are a couple of potential issues I found

scripts/fix-haddock-links.sh
carbolymer: I suppose the script should warn against executing on darwin instead of failing silently

carbolymer (approved):

scripts/fix-haddock-links.sh
Jimbo4350: Only intended to run in CI on ubuntu-latest. macOS was never a target.

scripts/fix-haddock-links.sh
Jimbo4350: Applied in 7d1d6312f across all three xargs sh -c probe call sites (Phase 2 dead-URL probe, Phase 3 reexport probe, Phase 3a rescue probe). Apostrophes in Haskell module names (e.g. `Foo'`) would have broken the previous `{}` substitution by closing the outer `'...' wrapper, so this is a real fix, not just defensive.

scripts/fix-haddock-links.sh
Jimbo4350: Good point — this is a fourth unfixable sub-cause that was missing from the doc block. Added in 519aadcd5 as sub-cause (d) alongside the existing (a) umbrella-vs-implementation, (b) `KNOWN_UNDOCUMENTED`, and (c) Hackage-URL-without-version cases.

PR #1181, opened 2026-04-09
makeUnsignedTx: error when Plutus scripts present without protocol params

Jimbo4350:
# Changelog

```yaml
- description: |
    makeUnsignedTx now returns Either MakeUnsignedTxError and errors when
    Plutus scripts are present but protocol parameters are missing, instead
    of silently omitting the script integrity hash (script_data_hash).
  type:
# uncomment at least one main project this PR is associated with
  projects:
   - cardano-api
  # - cardano-api-gen
  # - cardano-rpc
  # - cardano-wasm
```

# Context

Addresses [IntersectMBO/cardano-cli#1363](https://github.com/IntersectMBO/cardano-cli/issues/1363).

`convPParamsToScriptIntegrityHash` silently returned `SNothing` when protocol parameters were `Nothing` but Plutus scripts required computing the script integrity hash (`script_data_hash`, field 11 in the Conway CDDL). The `StrictMaybe` monad short-circuited on `maybeToStrictMaybe Nothing`, masking the real error.

This caused `build-raw` in cardano-cli to produce invalid transaction bodies when using reference scripts without `--protocol-params-file`.

## Changes

- `makeUnsignedTx` now returns `Either MakeUnsignedTxError (UnsignedTx (LedgerEra era))` (breaking change)
- `convPParamsToScriptIntegrityHash` checks whether the hash needs to be computed first, then fails explicitly if protocol params are missing
- `TxBodyErrorAutoBalance` and `TxFeeEstimationError` gain constructors to propagate the new error
- Test call sites updated

# How to trust this PR

- The happy path (protocol params always present) is unchanged — `convPParamsToScriptIntegrityHash` returns `Right (SJust hash)` as before
- The `estimateBalancedTxBody` / `makeTransactionBodyAutoBalance` callers always set protocol params on `TxBodyContent`, so the `Left` branch is unreachable there
- The fix targets the silent `SNothing` path that only fires when callers omit protocol params with Plutus scripts present

# Checklist

- [ ] New tests are added if needed and existing tests are updated
- [x] Self-reviewed the diff

### DIFF
--- a/cardano-api/src/Cardano/Api/Experimental.hs
+++ b/cardano-api/src/Cardano/Api/Experimental.hs
@@ -12,6 +12,7 @@ module Cardano.Api.Experimental
     -- ** Transaction-related
     UnsignedTx (..)
   , SignedTx (..)
+  , MakeUnsignedTxError (..)
   , makeUnsignedTx
   , makeKeyWitness
   , signTx
--- a/cardano-api/src/Cardano/Api/Experimental/Tx.hs
+++ b/cardano-api/src/Cardano/Api/Experimental/Tx.hs
@@ -117,6 +117,7 @@ module Cardano.Api.Experimental.Tx
     -- * Contents
     UnsignedTx (..)
   , SignedTx (..)
+  , MakeUnsignedTxError (..)
   , makeUnsignedTx
   , makeKeyWitness
   , signTx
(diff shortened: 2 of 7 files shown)


carbolymer (approved):

palas (approved):
Looks good, and it would be good to add a regression test too

PR #1185, opened 2026-04-17
Remove proto-lens SRP

### LINKED ISSUE #1184: CI: `buf` version unpinned, causing generated proto files to drift
## Problem

Multiple open PRs fail the `[cardano-rpc] Check that generated files from proto definitions are up to date` CI step with the same output:

```
Generated files differ from repository files:
M	cardano-rpc/gen/Proto/Cardano/Rpc/Node.hs
M	cardano-rpc/gen/Proto/Utxorpc/V1beta/Cardano/Cardano.hs
M	cardano-rpc/gen/Proto/Utxorpc/V1beta/Query/Query.hs
M	cardano-rpc/gen/Proto/Utxorpc/V1beta/Submit/Submit.hs
```

Affected PRs observed so far:
- #1123 (2026-04-14)
- #1175 (2026-04-16)

Neither PR touches proto files or proto tooling. Master's last CI run (779f539e4 on 2026-04-14) passed this same check with the same committed gen files.

## Root cause

`.github/workflows/haskell.yml` downloads `buf` from the unpinned `releases/latest` URL:

```yaml
- name: '[Linux] [cardano-rpc] Install buf tool'
  if: runner.os == 'Linux'
  run: |
    curl -sSL "https://github.com/bufbuild/buf/releases/latest/download/buf-Linux-x86_64" -o "/usr/local/bin/buf"
    chmod +x /usr/local/bin/buf
```

When `buf` publishes a new release, its generated output can differ from what was committed when the last regen ran. Every PR opened after that point fails the check until someone regenerates with the new `buf`.

## Fix options

1. Pin `buf` to a specific version in the workflow (matches the proto-lens-protoc install, which respects the cabal.project index-state).
2. Regenerate `cardano-rpc/gen/` with current `buf` and commit to master — only resolves until the next `buf` release.

Option 1 is the durable fix.

carbolymer:
# Context

Remove proto-lens SRP. Fix buf version drift.

Fix: https://github.com/IntersectMBO/cardano-api/issues/1184

# Checklist

- [ ] Self-reviewed the diff


angerman (approved):
Hooray!

flake.nix
```diff
@@ -209,7 +188,10 @@
             shellcheck
             snappy
             protobuf
-            buf
+            # buf version must match `.github/workflows/haskell.yml` (buf is
+            # not backwards-compatible across minor versions for
+            # `buf generate` output).
+            unstable.buf
```
Jimbo4350, line 195: `unstable.buf` doesn't actually pin a version. `inputs.unstable.url = "nixpkgs/nixos-unstable"` is locked by `flake.lock` today, so it's deterministic for this checkout but the next `nix flake update` can silently move `buf` off `v1.66.1` and reintroduce the exact drift this PR is trying to fix. The comment above asserts alignment with CI but nothing enforces it. Could we fetch the pinned binary the same way CI does (or otherwise derive both versions from a single source of truth)?

Jimbo4350 (commented):
Two small things beyond what Copilot flagged.

**`cardano-api/cardano-api.cabal:339`** (not in the diff, flagging here): the main library bound was bumped to `cardano-ledger-core >=1.20`, but the test-suite's `cardano-ledger-core:{cardano-ledger-core, testlib} >=1.17` is now stale. Cabal resolves to `>=1.20` via the library constraint so it builds, but the bound here is misleading. Worth bumping to `>=1.20` for consistency.

.github/workflows/haskell.yml
```diff
@@ -97,8 +97,10 @@ jobs:
 
     - name: '[Linux] [cardano-rpc] Install buf tool'
       if: runner.os == 'Linux'
+      # buf version must match the one in flake.nix dev shell (buf is not
+      # backwards-compatible across minor versions for `buf generate` output).
       run: |
-        curl -sSL "https://github.com/bufbuild/buf/releases/latest/download/buf-Linux-x86_64" -o "/usr/local/bin/buf"
+        curl -sSL "https://github.com/bufbuild/buf/releases/download/v1.66.1/buf-Linux-x86_64" -o "/usr/local/bin/buf"
```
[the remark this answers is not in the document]
carbolymer: will use buf from nixpkgs

flake.nix
carbolymer: > the next nix flake update can silently move buf off v1.66.1 and reintroduce the exact drift this PR is trying to fix.

This is fine in my opinion, as flake update is a thing we can control. I'll do the opposite and implement the check using `buf` from nixpkgs.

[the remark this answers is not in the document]
carbolymer: will use buf from nixpkgs

carbolymer: I've split buf generate into separate job - that should speed up GHA since it'll run in parallel. It'll also use the same buf as nix development shell, so it'll be pinned to a version from nixpkgs.

Jimbo4350 (approved):

PR #1199, opened 2026-05-06
Widen Exp.SignedTx to all Shelley-based eras

Jimbo4350:
# Context

Reparameterise `Cardano.Api.Experimental.SignedTx` and its `SerialiseAsRawBytes` instance on `ShelleyLedgerEra era` instead of `LedgerEra era`, so `SignedTx era` types for every Shelley-based era (Shelley → Dijkstra) — not just Conway+.

The motivation is unblocking the cardano-cli debug transaction view (`runTransactionViewCmd` handles `InAnyShelleyBasedEra`). Migrating it off the legacy `TxBodyContent` decomposition onto `Exp.SignedTx era` + ledger-side lenses regresses on pre-Conway tx files unless `Exp.SignedTx` is reachable for all Shelley-based eras.

# How to trust this PR

The change is deliberately surgical — 12 lines across two files, plus a changelog fragment.

**Why the `LedgerEra` family is left alone:** `EraCommonConstraints` (Era.hs:301) bundles Conway-onwards ledger constraints (`L.ConwayEraGov`, `L.ConwayEraTxBody`, `L.ConwayEraTxCert`, `L.ConwayEraCertState`, `L.GovState ~ L.ConwayGovState`) over `LedgerEra era`. Widening `LedgerEra` would break every site that uses `obtainCommonConstraints` or `IsEra era` — i.e. `Tx.hs` builders, `Certificate.hs`, `BodyContent/New.hs`. Switching only `SignedTx` (which needs only `L.EraTx (ShelleyLedgerEra era)`, available for every Shelley-based era via `ShelleyBasedEraConstraints`) avoids that ripple entirely.

**Build paths still Conway+ gated:** `signTx`, `makeKeyWitness`, and `convertTxBodyToUnsignedTx` take `Era era`, whose constructors are `ConwayEra`/`DijkstraEra` only. For those eras, `LedgerEra era ~ ShelleyLedgerEra era` is supplied by `obtainCommonConstraints`, so `SignedTx signedTx` (Tx.hs:332) elaborates without code change.

**Byron is irrelevant:** `ShelleyLedgerEra` is undefined for Byron, and `runTransactionViewCmd` already excludes Byron via `InAnyShelleyBasedEra`. Byron tx is the separate `ATxAux` type.

**Reachability:** `ShelleyLedgerEra` is already reachable to downstream code via `import Cardano.Api` (transitively re-exported through `Cardano.Api.Era` → `Cardano.Api.Era.Internal.Eon.ShelleyBasedEra`), so no new exports were needed.

To verify:
- `cabal build cardano-api -j4` — clean.
- `cabal test cardano-api:cardano-api-test --test-show-details=direct -j4` — all 220 tests pass locally.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-api/src/Cardano/Api/Experimental/Tx.hs
+++ b/cardano-api/src/Cardano/Api/Experimental/Tx.hs
@@ -281,7 +281,7 @@ makeKeyWitness era (UnsignedTx unsignedTx) wsk =
 
 -- | A transaction that has been witnesssed
 data SignedTx era
-  = L.EraTx (LedgerEra era) => SignedTx (Ledger.Tx Ledger.TopTx (LedgerEra era))
+  = L.EraTx (ShelleyLedgerEra era) => SignedTx (Ledger.Tx Ledger.TopTx (ShelleyLedgerEra era))
 
 deriving instance Eq (SignedTx era)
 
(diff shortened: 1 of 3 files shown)


cardano-api/src/Cardano/Api/Experimental/Tx.hs
```diff
@@ -281,7 +281,7 @@ makeKeyWitness era (UnsignedTx unsignedTx) wsk =
 
 -- | A transaction that has been witnesssed
 data SignedTx era
```
carbolymer, line 283: We should switch to ledger eras here eventually

carbolymer (approved):

cardano-api/src/Cardano/Api/Experimental/Tx.hs
Jimbo4350: 💯

PR #1201, opened 2026-05-07
Add regression test: makeUnsignedTx errors without protocol params for Plutus scripts

Jimbo4350:
# Context

Companion regression test for [#1181](https://github.com/IntersectMBO/cardano-api/pull/1181) ("makeUnsignedTx: error when Plutus scripts present without protocol params"), which merged the fix but landed without a regression test. This PR closes that gap.

# How to trust this PR

- Single test added to `Test.Cardano.Api.Experimental`: `prop_makeUnsignedTx_plutus_without_pparams`.
- Builds a `TxBodyContent` with a Plutus spending witness and **no** protocol parameters, asserts `makeUnsignedTx` returns `Left MakeUnsignedTxMissingProtocolParams`.
- Run:
  ```
  cabal test cardano-api-test -j4 --test-options='-p "/makeUnsignedTx/"'
  ```

# Checklist

- [x] Self-reviewed the diff
- [ ] Changelog fragment added in `.changes/` (will add once PR number is assigned)

### DIFF
--- a/.changes/20260507_cardano_api_regression_test_makeunsignedtx_plutus_pparams.yml
+++ b/.changes/20260507_cardano_api_regression_test_makeunsignedtx_plutus_pparams.yml
@@ -1,0 +1,6 @@
+project: cardano-api
+pr: 1201
+kind:
+  - test
+description: |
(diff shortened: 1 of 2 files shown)


palas (approved):
👍 I tried to test it against before the fix, but it didn't even return an `Either`

carbolymer (approved):

PR #1235, opened 2026-06-15
getLedgerTablesUTxOValues: use getOriginalTxIn instead of coerceMapKeys

Jimbo4350:
# Context

Replace `coerceMapKeys` with `Shelley.getOriginalTxIn` in `getLedgerTablesUTxOValues`. The previous implementation used `coerceMapKeys` to convert `Shelley.TxIn` map keys, which required suppressing the GHC 9.10+ `-Wno-x-ord-preserving-coercions` warning. Using `getOriginalTxIn` makes the conversion explicit and safe, removing the need for the suppression pragma.

# How to trust this PR

The diff is small — one import removed, one pragma removed, one line changed in `ejectTables`. Verify that `fromShelleyTxIn . Shelley.getOriginalTxIn` does the same thing as `fromShelleyTxIn $ coerceMapKeys` previously did.

```bash
cabal build cardano-api -j4
```

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-api/src/Cardano/Api/LedgerState.hs
+++ b/cardano-api/src/Cardano/Api/LedgerState.hs
@@ -1,5 +1,4 @@
 {-# LANGUAGE BangPatterns #-}
-{-# LANGUAGE CPP #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE FlexibleContexts #-}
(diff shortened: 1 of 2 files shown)


cardano-api/src/Cardano/Api/LedgerState.hs
```diff
@@ -12,10 +12,6 @@
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeOperators #-}
 
-#if __GLASGOW_HASKELL__ >= 910
```
carbolymer, line 15: `CPP` extension still remains in the file

carbolymer (changes requested):

carbolymer (approved):

PR #1238, opened 2026-06-25
 Fix non-canonical inline datum JSON round-trip

v0d1ch:
● Fix non-canonical inline datum JSON round-trip

  Problem

  FromJSON (TxOut) crashes when parsing a TxOut whose inline datum was encoded with non-canonical CBOR bytes (e.g. definite-length arrays instead of the indefinite-length form Plutus normally emits).

  The error:
  "Inline datum not equivalent to inline datum hash"

  This happens because ToJSON and FromJSON disagree on how to reconstruct HashableScriptData:

  - ToJSON serialises the inline datum as three fields:
    - inlineDatum — the human-readable JSON representation
    - inlineDatumhash — blake2b_256(originalCborBytes)
    - inlineDatumRaw — the original CBOR bytes as hex
  - FromJSON ignored inlineDatumRaw entirely and reconstructed HashableScriptData via scriptDataFromJson, which always re-serialises to canonical CBOR. For non-canonical original bytes:
    - H(canonical) ≠ H(original) = inlineDatumhash
    - → parse fails with the above error

  Non-canonical Plutus CBOR is valid and accepted by the ledger. It arises when scripts use definite-length array encoding for constructor fields instead of the canonical indefinite-length form. Any node that stores such a UTxO entry and later replays it from a JSON snapshot (e.g. from a SQLite event log) would crash on
  deserialisation.

  Fix

  inlineDatumRaw is always present in JSON produced by ToJSON (as null for non-inline outputs, as hex for inline ones). FromJSON now reads this field first and uses deserialiseFromCBOR AsHashableScriptData to reconstruct HashableScriptData with the original bytes preserved. The scriptDataFromJson path is kept as a fallback for JSON
  produced by external tools or older versions that may not include inlineDatumRaw.

  The fix is applied to all six affected sites: Babbage, Conway, and Dijkstra eras in both FromJSON (TxOut CtxTx era) and FromJSON (TxOut CtxUTxO era). The repeated logic is extracted into a parseInlineDatum helper.

  Testing

  Added prop_json_roundtrip_txout_noncanonical_inline_datum — a property test that generates a TxOut with a non-canonical inline datum (definite-length CBOR constructor) and asserts the JSON round-trip is identity. The test fails without the fix and passes with it.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/.changes/20260625_fix_non_canonical_inline_datum_json_roundtrip.yml
+++ b/.changes/20260625_fix_non_canonical_inline_datum_json_roundtrip.yml
@@ -1,0 +1,7 @@
+project: cardano-api
+pr: 1238
+kind:
+  - bugfix
+description: |
+  FromJSON (TxOut) crashes when parsing a TxOut whose inline datum was encoded with non-canonical CBOR bytes (e.g. definite-length arrays instead of the indefinite-length form Plutus normally emits).
+  
--- a/.changes/20260625_fix_non_canonical_inline_datum_json_roundtrip.yml
+++ b/.changes/20260625_fix_non_canonical_inline_datum_json_roundtrip.yml
@@ -1,0 +1,7 @@
+project: cardano-api
+pr: 1238
+kind:
+  - bugfix
+description: |
+  FromJSON (TxOut) no longer crashes when parsing a TxOut whose inline datum was encoded with non-canonical CBOR bytes (e.g. definite-length arrays instead of the indefinite-length form Plutus normally emits).
+  
--- a/cardano-api/gen/Test/Gen/Cardano/Api/Typed.hs
+++ b/cardano-api/gen/Test/Gen/Cardano/Api/Typed.hs
@@ -44,6 +44,7 @@ module Test.Gen.Cardano.Api.Typed
 
     -- * Scripts
   , genHashableScriptData
+  , genNonCanonicalHashableScriptData
   , genReferenceScript
   , genScript
   , genValidScript
@@ -188,7 +189,7 @@ import Data.Int (Int64)
 import Data.Maybe
 import Data.Ratio (Ratio, (%))
 import Data.String
-import Data.Word (Word32, Word64)
+import Data.Word (Word32, Word64, Word8)
 import GHC.Exts (IsList (..))
 import GHC.Stack
 import Numeric.Natural (Natural)
@@ -390,6 +391,26 @@ genHashableScriptData = do
   case deserialiseFromCBOR AsHashableScriptData $ serialiseToCBOR sd of
     Left e -> error $ "genHashableScriptData: " <> show e
     Right r -> return r
+
+-- | Generate 'HashableScriptData' whose CBOR uses a definite-length array
+-- instead of the canonical indefinite-length form that Plutus normally emits.
+-- This means 'hashScriptDataBytes' of the result differs from
+-- 'hashScriptDataBytes' of its canonical re-encoding, exposing any JSON
+-- round-trip that reconstructs CBOR rather than preserving original bytes.
+genNonCanonicalHashableScriptData :: HasCallStack => Gen HashableScriptData
+genNonCanonicalHashableScriptData = do
+  constrIdx <- Gen.integral (Range.linear 0 6 :: Range.Range Int)
+  args <- Gen.list (Range.linear 1 5) (Gen.integral (Range.linear 0 23 :: Range.Range Int))
+  -- Plutus constructor n uses CBOR tag 121+n.
+  -- Canonical encoding wraps fields in an indefinite-length array (0x9f..0xff).
+  -- We use a definite-length array (0x80+len) to produce non-canonical bytes.
+  let tagBytes = [0xd8, fromIntegral (0x79 + constrIdx)] :: [Word8]
+      arrayHdr = [fromIntegral (0x80 + length args)] :: [Word8]
+      argBytes = map fromIntegral args :: [Word8]
+      bytes = BS.pack (tagBytes <> arrayHdr <> argBytes)
+  case deserialiseFromCBOR AsHashableScriptData bytes of
+    Left e -> error $ "genNonCanonicalHashableScriptData: " <> show e
+    Right r -> pure r
 
 {-# DEPRECATED genScriptData "Use genHashableScriptData" #-}
 genScriptData :: Gen ScriptData
--- a/cardano-api/gen/Test/Gen/Cardano/Api/Typed.hs
+++ b/cardano-api/gen/Test/Gen/Cardano/Api/Typed.hs
@@ -44,6 +44,7 @@ module Test.Gen.Cardano.Api.Typed
 
     -- * Scripts
   , genHashableScriptData
+  , genNonCanonicalHashableScriptData
   , genReferenceScript
   , genScript
   , genValidScript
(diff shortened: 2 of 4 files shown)


Crypto2099:
Thanks for getting on this so quickly @v0d1ch! This is currently causing my hydra heads to fail to restart when a non-canonical datum is present in the ledger and the hydra-node crashes for any reason and attempts to validate its snapshot state from the state database.

noonio:
Any change of getting this merged? @Jimbo4350

cardano-api/src/Cardano/Api/Tx/Internal/Output.hs
```diff
@@ -431,6 +431,26 @@ txOutToJsonValue era (TxOut addr val dat refScript) =
...
+  -> Aeson.Parser HashableScriptData
+parseInlineDatum o dVal h fallback = do
+  mRaw <- o .:? "inlineDatumRaw"
+  hashableData <- case mRaw of
+    Just rawHex -> do
+      rawBytes <- either fail pure $ Base16.decode (Text.encodeUtf8 rawHex)
+      either (fail . show) pure $ deserialiseFromCBOR AsHashableScriptData rawBytes
+    Nothing -> fallback dVal
+  if hashScriptDataBytes hashableData /= h
+    then fail "Inline datum not equivalent to inline datum hash"
```
Jimbo4350: Can you also render the original object in the failure?

cardano-api/src/Cardano/Api/Tx/Internal/Output.hs
```diff
@@ -462,13 +482,12 @@ instance IsShelleyBasedEra era => FromJSON (TxOut CtxTx era) where
...
-              case scriptDataJsonToHashable ScriptDataJsonDetailedSchema dVal of
-                Left err ->
-                  fail $ "Error parsing TxOut JSON: " <> displayError err
-                Right hashableData -> do
-                  if hashScriptDataBytes hashableData /= h
-                    then fail "Inline datum not equivalent to inline datum hash"
-                    else return $ TxOutDatumInline BabbageEraOnwardsBabbage hashableData
+              hashableData <-
+                parseInlineDatum o dVal h $ \v ->
+                  case scriptDataJsonToHashable ScriptDataJsonDetailedSchema v of
```
Jimbo4350: Why not just inline this in `parseInlineDatum`? The parsing function you pass in is the same.

Jimbo4350 (approved):
Nice catch and LGTM. Just two small changes.

v0d1ch:
@Jimbo4350 I think this PR is ready to be merged now.

.changes/20260625_fix_non_canonical_inline_datum_json_roundtrip.yml
```diff
@@ -0,0 +1,7 @@
+project: cardano-api
+pr: 1238
+kind:
+  - bugfix
+description: |
+  FromJSON (TxOut) crashes when parsing a TxOut whose inline datum was encoded with non-canonical CBOR bytes (e.g. definite-length arrays instead of the indefinite-length form Plutus normally emits).
```
carbolymer: It's no longer crashing after this PR I guess? 😉 
A nit, but a different tense should be used then.

cardano-api/gen/Test/Gen/Cardano/Api/Typed.hs
```diff
@@ -391,6 +392,26 @@ genHashableScriptData = do
...
+  args <- Gen.list (Range.linear 1 5) (Gen.integral (Range.linear 0 23 :: Range.Range Int))
+  -- Plutus constructor n uses CBOR tag 121+n.
+  -- Canonical encoding wraps fields in an indefinite-length array (0x9f..0xff).
+  -- We use a definite-length array (0x80+len) to produce non-canonical bytes.
+  let tagBytes = [0xd8, fromIntegral (0x79 + constrIdx)] :: [Word8]
+      arrayHdr = [fromIntegral (0x80 + length args)] :: [Word8]
+      argBytes = map fromIntegral args :: [Word8]
+      bytes = BS.pack (tagBytes <> arrayHdr <> argBytes)
+  case deserialiseFromCBOR AsHashableScriptData bytes of
+    Left e -> error $ "genNonCanonicalHashableScriptData: " <> show e
```
carbolymer: a comment indicating that the left case is impossible would be nice

cardano-api/src/Cardano/Api/Tx/Internal/Output.hs
```diff
@@ -431,6 +433,29 @@ txOutToJsonValue era (TxOut addr val dat refScript) =
...
+-- | Parse 'HashableScriptData' from a JSON object, preferring the raw CBOR
+-- bytes in @inlineDatumRaw@ when present to preserve non-canonical encodings,
+-- falling back to the detailed-schema JSON for objects that lack the field.
+parseInlineDatum
+  :: Aeson.Object
+  -> Aeson.Value
+  -> Hash ScriptData
+  -> Aeson.Parser HashableScriptData
+parseInlineDatum o dVal h = do
+  mRaw <- o .:? "inlineDatumRaw"
```
carbolymer, line 446: This obfuscates a bit which fields are used to obtain inline datum: right now the field names are split into two places: in the helper function and in the call sites. We should keep obtaining of the field values in one place. So I suggest using one of the following:
1. Move obtaining `inlineDatumRaw`, `inlineDatum`, `inlineDatumHash` from the `Value` here
2. Accept the fields' values as function parameters instead of digging out raw datum separately here. This wouldn't be so nicely typed since the function would have to accept three `Value`s.

cardano-api/src/Cardano/Api/Tx/Internal/Output.hs
```diff
@@ -431,6 +433,29 @@ txOutToJsonValue era (TxOut addr val dat refScript) =
...
+  hashableData <- case mRaw of
+    Just rawHex -> do
+      rawBytes <- either fail pure $ Base16.decode (Text.encodeUtf8 rawHex)
```
carbolymer: At this point we ignore `inlineDatum` field. This is fine I think. But I think we should state this more clearly in the comment, that the field is ignored.

cardano-api/src/Cardano/Api/Tx/Internal/Output.hs
```diff
@@ -462,13 +487,8 @@ instance IsShelleyBasedEra era => FromJSON (TxOut CtxTx era) where
...
             (Just dVal, Just h) -> do
-              case scriptDataJsonToHashable ScriptDataJsonDetailedSchema dVal of
-                Left err ->
-                  fail $ "Error parsing TxOut JSON: " <> displayError err
-                Right hashableData -> do
-                  if hashScriptDataBytes hashableData /= h
-                    then fail "Inline datum not equivalent to inline datum hash"
-                    else return $ TxOutDatumInline BabbageEraOnwardsBabbage hashableData
+              hashableData <- parseInlineDatum o dVal h
+              return $ TxOutDatumInline BabbageEraOnwardsBabbage hashableData
```
carbolymer: I'd fold this into
```haskell
    (Just dVal, Just h) -> 
      TxOutDatumInline BabbageEraOnwardsBabbage <$> parseInlineDatum o dVal h
```
but this is fine as it is now. 👍🏻

carbolymer (approved):
Looks correct. I have some code style remarks.

Thanks!

Jimbo4350 (approved):

v0d1ch:
Thanks @carbolymer I think your comments are addressed now.

carbolymer (approved):

carbolymer:
@v0d1ch Thanks, LGTM! Could you squash the commits, please?

PR #1240, opened 2026-07-01
cardano-wasm: add stake-key witnessing (signWithStakeKey / alsoSignWithStakeKey)

palas:
# Context

  After #1239 fixed certificate deserialisation, stake **registration** works end to end, but **delegation** and
  **unregistration** transactions still can't be completed: they need a witness from the stake credential, and the only
  signing methods: `UnsignedTx.signWithPaymentKey` and `SignedTx.alsoSignWithPaymentKey`, reject stake signing keys
  (`stake_sk…`). Those transactions therefore go out missing a required witness and are rejected on submit.

  This PR adds stake-key witnessing to the `cardano-wasm` API:

  - `UnsignedTx.signWithStakeKey(signingKey)`
  - `SignedTx.alsoSignWithStakeKey(signingKey)`

  A vkey witness is key-agnostic (the ledger matches the vkey hash against the required-signer set), so the new methods
  mirror the payment-key ones exactly and only differ in decoding a stake signing key and building a `WitnessStakeKey`.
  Signing is not era-specific, so there are no era variants. Only non-extended `stake_sk` keys are accepted, matching the
  payment path (`addr_sk` only).

  The API method list has a single source (`Cardano.Wasm.Api.Info`) that generates both the `*.d.ts` and the runtime JS API,
  so most of the diff is that one edit plus its regenerated golden `.d.ts`. The JS bridge, the WASI bridge, and the
  `cardano-wasi` wasm export list are updated to match.

# How to trust this PR
  
  - The witness construction lives in `cardano-wasm/src-lib/Cardano/Wasm/Api/Tx.hs` (`signWithStakeKeyImpl` /
  `alsoSignWithStakeKeyImpl`): a line-for-line mirror of the payment versions, with `WitnessStakeKey` instead of
  `WitnessPaymentKey`.
  - New test in `cardano-wasm/npm-wrapper/api.test.js` builds a delegation transaction, signs it with the payment key and
  then `alsoSignWithStakeKey`.
  - To run:
    - `cabal test cardano-wasm:cardano-wasm-golden`: verifies the generated TypeScript declarations.
    - `npm install && npm run build && npm test` in `cardano-wasm/npm-wrapper`: exercises the new methods through the built wasm.

# Checklist

  - [x] New tests are added if needed and existing tests are updated. See [Running 
  tests](https://github.com/input-output-hk/cardano-node-wiki/wiki/Running-tests) for more details
  - [x] Self-reviewed the diff


cardano-wasm/src-lib/Cardano/Wasm/Api/Tx.hs
```diff
@@ -167,6 +169,24 @@ signWithPaymentKeyImpl (UnsignedTxObject era fullUnsignedTx@(Exp.UnsignedTx tx))
...
+    let witness = Exp.makeKeyWitness era fullUnsignedTx . Api.WitnessStakeKey $ signingKey
+        txWits =
+          Ledger.mkBasicTxWits
+            & Ledger.addrTxWitsL
+              .~ Set.fromList [witness]
+        txWithWits =
+          obtainCommonConstraints
+            era
+            (tx & Ledger.witsTxL .~ txWits)
+     in SignedTxObject
```
carbolymer: **Suggestion (maintenance):** `signWithStakeKeyImpl` and `alsoSignWithStakeKeyImpl` are near-verbatim copies of their payment-key counterparts, differing only in the witness constructor (`WitnessStakeKey` vs `WitnessPaymentKey`). Both constructors belong to `ShelleyWitnessSigningKey`, which is what `makeKeyWitness` already consumes.

A shared internal helper would collapse the duplication:

```haskell
signWithKeyImpl :: UnsignedTxObject -> Api.ShelleyWitnessSigningKey -> SignedTxObject
-- ... the current body, taking the already-constructed witness

alsoSignWithKeyImpl :: SignedTxObject -> Api.ShelleyWitnessSigningKey -> SignedTxObject
-- ... ditto, with <>~ instead of .~
```

Then each public function becomes a one-liner:

```haskell
signWithStakeKeyImpl utx = signWithKeyImpl utx . Api.WitnessStakeKey
signWithPaymentKeyImpl utx = signWithKeyImpl utx . Api.WitnessPaymentKey
```

This cuts 4 duplicated ~14-line blocks down to 2 helpers + 4 wrappers. The risk with the current form is that a future change to the witnessing logic (e.g. a new witness field, an era-specific fix) applied to one copy but missed in another would silently produce differently-witnessed transactions.

Not blocking - the duplication pattern was pre-existing. But worth considering before adding the next key type.

carbolymer (commented):
👍🏻

cardano-wasm/src-wasi/Cardano/Wasi/Internal/Conversion.hs
```diff
@@ -59,3 +60,8 @@ stringToSigningKey :: CString -> IO (Api.SigningKey Api.PaymentKey)
 stringToSigningKey signingKeyCString = do
   string <- peekCString signingKeyCString
   rightOrError $ Api.deserialiseFromBech32 (Text.pack string)
+
+stringToStakeSigningKey :: CString -> IO (Api.SigningKey Api.StakeKey)
+stringToStakeSigningKey signingKeyCString = do
+  string <- peekCString signingKeyCString
+  rightOrError $ Api.deserialiseFromBech32 (Text.pack string)
```
carbolymer: **Suggestion (minor):** This is byte-for-byte identical to `stringToSigningKey` above - only the return type differs. Since `deserialiseFromBech32` is already polymorphic (constrained by `SerialiseAsBech32`), you could generalise to a single helper:

```haskell
stringToSigningKey :: SerialiseAsBech32 (Api.SigningKey keyrole) => CString -> IO (Api.SigningKey keyrole)
```

The call sites already pin the type via `<*>`, so inference resolves the role. Each new key type (DRep, committee, etc.) would then work without adding another copy.

cardano-wasm/npm-wrapper/api.test.js
```diff
@@ -46,4 +46,40 @@ describe('Cardano API', () => {
...
+            .setFee(10_000n);
+
+        // Sign with the payment key, then add a witness from the stake key
+        const signedTx = (await tx.signWithPaymentKey(paymentSecretKey))
+            .alsoSignWithStakeKey(stakeSecretKey);
+        expect(signedTx).toBeDefined();
+
+        const txCbor = await signedTx.txToCbor();
+
+        expect(txCbor).toBe("84a400d9010281825820be6efd42a3d7b9a00d09d77a5d41e55ceaf0bd093a8aa8a893ce70d9caafd97800018182581d6082935e44937e8b530f32ce672b5d600d0a286b4e8a52c6555f659b871a004c4b400219271004d901028183028200581ca9461e687627cddc5f54ffc988bc44321189538c601f1ad1b7979d9b581cbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbba100d9010282825820adfc1c30385916da87db1ba3328f0690a57ebb2a6ac9f6f86b2d97f943adae005840b5df44de4b1302d1b031363cfb9636be2f395561dcf083d1192656677506bfa401d988d745f2ee2d11ed9e563f35cc2dfbb605c7cd58a613e42377e5e2a8da01825820ee31f83c88a71219a6fcf9bee0da9bc22620588f5a15a6145553504df9649e5c58402a2a37f2391bdf3a1ff52a7ed2103473f35b43022ed101bddcb18a557f7e9f9eac75c3562b142cc15dc9ebd006da34f4076ff413914ec90e4359d757d73bfc06f5f6");
```
carbolymer, line 82: **Observation (test gap):** This test exercises `signWithPaymentKey` followed by `alsoSignWithStakeKey`, which covers `alsoSignWithStakeKeyImpl`. But the standalone `signWithStakeKey` method (the first-signer path through `signWithStakeKeyImpl`) has no test, despite being exported through all three bridges.

This is a legitimate code path for delegation flows - a user could write:

```js
(await tx.signWithStakeKey(stakeKey)).alsoSignWithPaymentKey(paymentKey)
```

and get a valid two-witness transaction. The 4 symmetric signing methods are all exported, but only 3 are tested. Worth adding a test that uses `signWithStakeKey` as the first signer.

carbolymer (approved):

Jimbo4350 (approved):
LGTM

cardano-wasm/src-lib/Cardano/Wasm/Api/Tx.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-api/commit/db4a6b1745b46302bd4c4751b1f9fe42754edc22

cardano-wasm/src-wasi/Cardano/Wasi/Internal/Conversion.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-api/commit/4971692b750ee2fac91890d9d6549f0279f91036

cardano-wasm/npm-wrapper/api.test.js
palas: Addressed in: https://github.com/IntersectMBO/cardano-api/commit/ed0289f8d349d78d58a5cb47e91949f9efe369a5

PR #1241, opened 2026-07-01
Update dependencies

erikd:
# Context

* Bump aeson lower bound to avoid potential DoS attack
* Update dependencies.
* Update index-states
* 
# How to trust this PR

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/.changes/20260701_updates.yml
+++ b/.changes/20260701_updates.yml
@@ -1,0 +1,31 @@
+# Changelog fragment template - copy this file and fill in the fields.
+# Or use 'herald new' for interactive creation.
+#
+# Files starting with _ are ignored by herald.
+#
+# Available projects and kinds are defined in .herald.yml
+
+# Which project this change belongs to (see 'projects' in .herald.yml)
+# Uncomment exactly one:
+project: cardano-api
+# project: cardano-api-gen
+# project: cardano-rpc
+# project: cardano-wasm
+
+# Pull request number associated with this change
+pr: 1241
+
+# One or more change kinds (see 'kinds' in .herald.yml)
+kind:
+#   - breaking       # the API has changed in a breaking way
+#   - feature        # introduces a new feature
+#   - compatible     # the API has changed but is non-breaking
+#   - bugfix         # fixes a defect
+#   - optimisation   # measurable performance improvements
+#   - documentation  # change in code docs, haddocks
+#   - refactoring    # code quality improvements
+#   - test           # fixes or modifies tests
+   - maintenance    # not directly related to the code
+#   - release        # related to a new release preparation
+description: |
+  Update dependencies, increase lower bound of aeson dependency.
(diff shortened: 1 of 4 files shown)


cabal.project
```diff
@@ -125,6 +125,17 @@ if arch(wasm32)
...
+allow-newer:
+  , cborg-json:aeson
+  , criterion-measurement:aeson
+  , deriving-aeson:aeson
+  , grapesy:aeson
+  , grpc-spec:aeson
+  , hedgehog-extras:aeson
+  , microstache:aeson
+  , monad-control:transformers
+  , http-api-data:text-iso8601
```
carbolymer, line 138: shouldn't that go below `cabal-allow-newer begin`?

carbolymer (approved):
LGTM, just small thing: https://github.com/IntersectMBO/cardano-api/pull/1241/changes#r3506150477

.changes/20260701_updates.yml
```diff
@@ -0,0 +1,31 @@
...
+#   - compatible     # the API has changed but is non-breaking
+#   - bugfix         # fixes a defect
+#   - optimisation   # measurable performance improvements
+#   - documentation  # change in code docs, haddocks
+#   - refactoring    # code quality improvements
+#   - test           # fixes or modifies tests
+   - maintenance    # not directly related to the code
+#   - release        # related to a new release preparation
+description: |
+  Describe your change here.
```
Jimbo4350: Change needs to be described.

cabal.project
```diff
@@ -125,6 +125,17 @@ if arch(wasm32)
 -- Do NOT add more source-repository-package stanzas here unless they are strictly
 -- temporary! Please read the section in CONTRIBUTING about updating dependencies.
 
+allow-newer:
```
Jimbo4350, line 129: Can you gate this behind a version conditional like the ghc >=9.14 block right below it instead of applying it unconditionally?

Jimbo4350 (changes requested):

cabal.project
erikd: No this set (depending on `aeson` and it transitive dependendcies) are for all versions of GHC.

cabal.project
erikd: This MUST be applied unconditionally. It applies to all versions of GHC.

.changes/20260701_updates.yml
erikd: Done.

cabal.project
erikd: I have added a comment on this.

cabal.project
carbolymer: I didn't mean gating it behind GHC version. I just don't understand why did you add `-- cabal-allow-newer begin` block.

Jimbo4350 (approved):

cabal.project
erikd: That is a marker that is added by a tool I got LLMs to write.

PR #1243, opened 2026-07-08
cardano-wasm: expose the transaction id (getTxId)

palas:
# Context

Adds a `getTxId` method to both `UnsignedTx` and `SignedTx` in the `cardano-wasm` API. It returns the transaction id (the blake2b-256 hash of the transaction body) as a hex string.

Since the id only depends on the body, it is available before signing and is not changed by adding witnesses. This is useful for flows that don't submit through the API (e.g. air-gapped signing), where the user may still want to display or look up the id.

# How to trust this PR

- The core logic is just `getUnsignedTxIdImpl` and `getSignedTxIdImpl` in `cardano-wasm/src-lib/Cardano/Wasm/Api/Tx.hs`. Both hash the transaction body with `getTxIdShelley`, so the unsigned and signed ids agree by construction.
- Everything else is plumbing: WASI and JS bridge exports, API metadata in `Info.hs`, TypeScript typings, and the updated golden file.
- The new test in `cardano-wasm/npm-wrapper/api.test.js` checks the id against an independently computed hash of the expected body CBOR, and verifies it is unchanged after signing once and after adding a second witness. It runs in the WASM CI job, or locally with `npm test` in `cardano-wasm/npm-wrapper`.

# Checklist

- [x] Self-reviewed the diff


carbolymer (approved):

.changes/20260708_120000_cardano-wasm_pablo.lamela_expose_tx_id.yml
```diff
@@ -0,0 +1,9 @@
+description: Added `getTxId` to the `cardano-wasm` API (on `UnsignedTx` and `SignedTx`),
+  returning the transaction id (the blake2b-256 hash of the transaction body) as a
+  hex string. The id is available as soon as the transaction body is final and is
+  not affected by signing, so flows that do not submit through the app (e.g.
+  air-gapped signing) can display, copy, or link the id.
```
carbolymer: redundant line breaks

Jimbo4350 (approved):

.changes/20260708_120000_cardano-wasm_pablo.lamela_expose_tx_id.yml
palas: Addressed in amended commit

PR #1245, opened 2026-07-08
Add readAnyScriptBytes and readFileAnyScript to the experimental API

Jimbo4350:
# Context

Adds script reading to the experimental API, so downstream consumers (e.g. `cardano-cli`) can decode a script file without knowing its language or serialisation format up front:

- `readAnyScriptBytes` decodes an `AnyScript (LedgerEra era)` from bytes: it first tries a `TextEnvelope`, using the envelope type to select Plutus (`decodeAnyPlutusScript`) or simple script (`deserialiseSimpleScript`) CBOR decoding, and falls back to the legacy JSON-only simple script encoding for inputs that are not text envelopes.
- `readFileAnyScript` is the file-reading wrapper, returning `FileError AnyScriptDecodeError`.
- `AnyScriptDecodeError` is the new error type covering the JSON and CBOR failure cases, with an `Error` instance.

The legacy JSON fallback uses the established `TODO Dijkstra` placeholder for the Dijkstra era, consistent with the rest of the experimental API.

# How to trust this PR

```bash
cabal build cardano-api -j4
```

The change is additive: no existing functions are modified, only new exports from `Cardano.Api.Experimental`.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-api/src/Cardano/Api/Experimental/AnyScript.hs
+++ b/cardano-api/src/Cardano/Api/Experimental/AnyScript.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GADTs #-}
@@ -10,26 +11,46 @@
 module Cardano.Api.Experimental.AnyScript
   ( AnyScript (..)
   , AsType (..)
+  , AnyScriptDecodeError (..)
   , deserialiseAnyPlutusScriptOfLanguage
   , deserialiseAnySimpleScript
   , hashAnyScript
+  , readAnyScriptBytes
+  , readFileAnyScript
   )
 where
 
+import Cardano.Api.Error (Error (..), FileError (..), fileIOExceptT)
 import Cardano.Api.Experimental.Era
 import Cardano.Api.Experimental.Plutus.Internal.Script hiding (AnyPlutusScript)
+import Cardano.Api.Experimental.Plutus.Internal.Script qualified as PlutusScript
 import Cardano.Api.Experimental.Simple.Script
 import Cardano.Api.HasTypeProxy
+import Cardano.Api.IO (File (..), FileDirection (In), readFileBlocking)
 import Cardano.Api.Ledger.Internal.Reexport qualified as L
+import Cardano.Api.Plutus.Internal.Script (toAllegraTimelock)
+import Cardano.Api.Plutus.Internal.Script qualified as OldScript
+  ( AsType (AsScript, AsSimpleScript)
+  , SimpleScript
+  )
+import Cardano.Api.Pretty (pretty, vsep, (<+>))
 import Cardano.Api.Serialise.Cbor
+import Cardano.Api.Serialise.Json (JsonDecodeError (..), deserialiseFromJSON)
+import Cardano.Api.Serialise.TextEnvelope (TextEnvelope (..), TextEnvelopeType (..))
+import Cardano.Api.Serialise.TextEnvelope.Internal (textEnvelopeType)
 
 import Cardano.Ledger.Binary qualified as CBOR
 import Cardano.Ledger.Core qualified as L
 import Cardano.Ledger.Plutus.Language qualified as Plutus
 
+import Control.Monad.Trans.Except (runExceptT)
+import Control.Monad.Trans.Except.Extra (firstExceptT, hoistEither)
+import Data.Bifunctor (bimap)
+import Data.ByteString (ByteString)
 import Data.ByteString qualified as BS
 import Data.Either.Combinators (maybeToRight, rightToMaybe)
 import Data.Foldable (asum)
+import Data.Text qualified as Text
 import Data.Type.Equality ((:~:) (..))
 import Data.Typeable (Typeable, eqT)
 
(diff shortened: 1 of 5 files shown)


cardano-api/src/Cardano/Api/Experimental/AnyScript.hs
```diff
@@ -111,3 +127,57 @@ deserialiseAnyPlutusScriptOfLanguage bs lang = do
...
+  :: forall era
+   . Era era
+  -> ByteString
+  -> Either AnyScriptDecodeError (AnyScript (LedgerEra era))
+readAnyScriptBytes era bs =
+  case deserialiseFromJSON bs :: Either JsonDecodeError TextEnvelope of
+    Right te ->
+      let scriptBs = teRawCBOR te
+          TextEnvelopeType anyScriptType = teType te
+       in case textToPlutusLanguage (Text.pack anyScriptType) of
```
carbolymer: The `Nothing` branch of `textToPlutusLanguage` is a catch-all: any non-Plutus text envelope type (e.g. `PaymentSigningKeyShelley_ed25519`, `StakePoolVerificationKey_ed25519`) gets fed to `deserialiseSimpleScript`, producing an opaque CBOR decode error instead of a clear "not a script envelope type" rejection.

Consider either matching the known simple script envelope type explicitly before decoding, or rejecting unrecognised types with a dedicated `AnyScriptDecodeError` constructor.

cardano-api/src/Cardano/Api/Experimental/AnyScript.hs
```diff
@@ -111,3 +127,57 @@ deserialiseAnyPlutusScriptOfLanguage bs lang = do
...
+            Just lang ->
+              case obtainCommonConstraints era $
+                     decodeAnyPlutusScript @(LedgerEra era) scriptBs lang
+                     :: Either CBOR.DecoderError (PlutusScript.AnyPlutusScript (LedgerEra era)) of
+                Left e -> Left (AnyScriptCborError e)
+                Right (PlutusScript.AnyPlutusScript ps) -> Right (AnyPlutusScript ps)
+            Nothing ->
+              bimap AnyScriptCborError AnySimpleScript $
+                obtainCommonConstraints era (deserialiseSimpleScript scriptBs)
+    Left _ ->
```
carbolymer: The text envelope JSON parse error is silently discarded here. If the input is a slightly malformed text envelope (e.g. missing `teRawCBOR`), the first `deserialiseFromJSON` fails with a useful error - but it's thrown away, and the user instead sees the irrelevant legacy-SimpleScript JSON error ("expected object with 'type' field" or similar).

Consider preserving the text envelope error when the fallback also fails, or at least surfacing it in preference to the fallback error.

cardano-api/src/Cardano/Api/Experimental/AnyScript.hs
```diff
@@ -111,3 +127,57 @@ deserialiseAnyPlutusScriptOfLanguage bs lang = do
   s :: (PlutusScriptInEra lang (LedgerEra era)) <-
     obtainCommonConstraints (useEra @era) (deserialisePlutusScriptInEra lang bs)
   return $ AnyPlutusScript s
+
+data AnyScriptDecodeError
```
carbolymer, line 136: `AnyScriptDecodeError` collapses four distinct decode paths into two constructors: `AnyScriptCborError` covers both Plutus and simple script CBOR failures, `AnyScriptJsonError` covers both text envelope and legacy simple script JSON failures. A caller (or user reading the error) cannot tell which path was attempted.

Consider adding a decode-path tag or the attempted envelope type to the constructors, e.g. `AnyScriptCborError ScriptKind CBOR.DecoderError`.

carbolymer:
The staircase flattens once the three decode paths are split out and you lean on `Maybe` for commit-or-fallback dispatch, `Either`'s `do` for error threading, and `<$>` to avoid the intermediate bind:

```haskell
readAnyScriptBytes
  :: forall era
   . Era era
  -> ByteString
  -> Either AnyScriptDecodeError (AnyScript (LedgerEra era))
readAnyScriptBytes era bs =
  fromMaybe (readLegacySimpleScript era bs) $
    readTextEnvelopeScript era <$> rightToMaybe (deserialiseFromJSON bs)

readTextEnvelopeScript
  :: forall era
   . Era era
  -> TextEnvelope
  -> Either AnyScriptDecodeError (AnyScript (LedgerEra era))
readTextEnvelopeScript era (TextEnvelope (TextEnvelopeType teType') _ scriptBs) =
  first AnyScriptCborError . obtainCommonConstraints era $
    fromMaybe (AnySimpleScript <$> deserialiseSimpleScript scriptBs) do
      lang <- textToPlutusLanguage $ Text.pack teType'
      pure $ repackage <$> decodeAnyPlutusScript @(LedgerEra era) scriptBs lang
 where
  repackage (PlutusScript.AnyPlutusScript ps) = AnyPlutusScript ps

readLegacySimpleScript
  :: forall era
   . Era era
  -> ByteString
  -> Either AnyScriptDecodeError (AnyScript (LedgerEra era))
readLegacySimpleScript era bs = do
  script <- first AnyScriptJsonError $ deserialiseFromJSON bs
  case era of
    ConwayEra ->
      obtainConwayConstraints era $
        pure . AnySimpleScript . SimpleScript $ toAllegraTimelock script
    DijkstraEra -> error "TODO Dijkstra: Simple script not supported"

readFileAnyScript
  :: Era era
  -> File content In
  -> IO (Either (FileError AnyScriptDecodeError) (AnyScript (LedgerEra era)))
readFileAnyScript era (File fp) = runExceptT do
  content <- fileIOExceptT fp readFileBlocking
  firstExceptT (FileError fp) . hoistEither $ readAnyScriptBytes era content
```

What changed:

- **`readAnyScriptBytes`**: `<$>` maps `readTextEnvelopeScript era` over the `Maybe TextEnvelope` from `rightToMaybe`. `fromMaybe` provides the legacy fallback. A successful text envelope parse commits to that path; CBOR failures inside propagate as `Left`, they don't silently retry the fallback. `rightToMaybe` discards the text-envelope JSON error intentionally - it's not an error, it's "this isn't a text envelope".
- **`readTextEnvelopeScript`**: `TextEnvelope` deconstructed in the argument pattern - no accessor functions. `fromMaybe ... do` on `Maybe` is the Alternative pattern: `textToPlutusLanguage` returns `Nothing` -> fall back to simple script; `Just lang` -> try Plutus. A single `first AnyScriptCborError` over both arms retires the two separate error-mapping expressions. Note: the Plutus repackaging uses `<$>` rather than a `do` bind - a `PlutusScript.AnyPlutusScript ps <- ...` bind would be a refutable match (existential constructor), so GHC desugars it through `fail` which needs `MonadFail (Either e)`.
- **`readLegacySimpleScript`**: `Either`'s `do` flattens the staircase. `first AnyScriptJsonError` replaces the explicit `Left e -> Left (...)` arm.
- **`readFileAnyScript`**: `File fp` deconstructed in the pattern; `runExceptT do` via `BlockArguments`.
- Import delta: `Data.Bifunctor (bimap)` -> `Data.Bifunctor (first)`.

carbolymer (approved):

Issue #1246, opened 2026-07-08
gRPC: `bytes` fields encoded as text or wrong payload (`TxOutput.address`, `Datum.hash`)


carbolymer:
## Summary

An audit of every proto `bytes` field populated by the UTxO RPC server found two fields that carry the wrong encoding.
`TxOutput.address` is populated with the human-readable address text (bech32 for Shelley, base58 for Byron) instead of the raw ledger address bytes.
`Datum.hash` is populated with the datum's CBOR instead of its hash when the datum is inline.
Both diverge from the other UTxO RPC implementations and, for the address, from our own predicate and reward-account encoding.

Evidence from other implementations:

- Dolos uses pallas-utxorpc, whose `map_tx_output` sets `address: x.address().map(|a| a.to_vec())` ([pallas-utxorpc/src/v1beta/mod.rs](https://github.com/txpipe/pallas/blob/main/pallas-utxorpc/src/v1beta/mod.rs)); `Address::to_vec` is documented in pallas-addresses as "Encode the address as its raw byte representation (header + payload)".
- Blink Labs cardano-node-api uses gouroboros, whose `ShelleyTransactionOutput.Utxorpc()` sets the address from `o.OutputAddress.Bytes()` ([ledger/shelley/shelley.go](https://github.com/blinklabs-io/gouroboros/blob/main/ledger/shelley/shelley.go)), i.e. raw header plus payload bytes.
- Both servers match `SearchUtxos` `exact_address` predicates against raw address bytes, so a client that echoes a `TxOutput.address` from one of those servers into a predicate gets consistent results; against our server it gets nothing.
- For `Datum.hash`, pallas-utxorpc `map_tx_datum` sets `hash: x.original_hash()` for inline datums (the blake2b-256 hash) and the bare hash for hash-only datums; the proto comment reads "Hash of this datum as seen on-chain".

The server is new and unreleased, so the window to fix the wire format without breaking clients is now.

## Finding 1: `TxOutput.address` carries bech32/base58 text

`txOutToUtxoRpcTxOutput` in `cardano-rpc/src/Cardano/Rpc/Server/Internal/UtxoRpc/Type.hs` populates the field with `T.encodeUtf8 (serialiseAddress addressInEra)`:

```haskell
& U5c.address .~ T.encodeUtf8 (shelleyBasedEraConstraints sbe $ serialiseAddress addressInEra)
```

[...]

PR #1247, opened 2026-07-08
cardano-rpc: Populate transaction bodies in FetchBlock responses

carbolymer:
# Context

Part of the node kernel access work (ADR-019): `FetchBlock` previously returned raw block bytes, the block header and the timestamp, but left `body.tx` empty.

This PR populates the transaction bodies. All 14 fields of the UTxO RPC `Tx` message are mapped: inputs, outputs, reference inputs, certificates, withdrawals, mint, witnesses, collateral, fee, validity, successful, auxiliary data, hash and governance proposals.

- Transactions are converted at ledger level for every Shelley-based era; era-gated fields (reference inputs, mint, collateral, proposals) stay empty where the era predates them.
- Spending, withdrawal and certificate redeemers are wired to their entries by script witness index, and `successful` reflects the phase-2 `IsValid` flag.
- A new `Certificate` module maps all 19 certificate variants (Dijkstra-safe via concrete-era dispatch); a new `Governance` module maps all 7 governance actions including partial parameter updates.
- Byron transactions are converted directly from the Byron ledger types, since cardano-api's `Tx era` has no Byron constructor. Byron fees are implicit in the UTxO and stay unset; epoch boundary blocks yield no transactions. This makes `FetchBlock` serve transaction data for the full mainnet history.

The first four commits are small cardano-api prerequisites, each with its own changelog fragment: re-export `byronBlockRaw` from `Cardano.Api.Consensus`, add `ConwayEraPParams` to `EraCommonConstraints`, export the per-era `toScriptIndex` variants, and make `genTx` sign the transaction body it returns.

Known limitations, deliberate and documented in code:

- `TxInput.as_output` is unset (resolving it requires UTxO lookups).
- Byron `fee` is unset (implicit in Byron; requires UTxO lookups).
- The Byron positional witness-to-input pairing is not representable in the proto `WitnessSet`; `native_bytes` is the fidelity escape hatch.
- Pre-existing `bytes`-field encoding divergences (`TxOutput.address` as text, inline `Datum.hash` carrying CBOR) are tracked separately in #1246 and not changed here.

# How to trust this PR

This PR is integrated and tested end-to-end in IntersectMBO/cardano-node#6579.

The conversion is covered at three levels:

- Property tests (`cardano-rpc-test`, 77 tests): per-era field projections of generated transactions against the ledger values across all seven Shelley-based eras, certificate arm routing, injected-redeemer wiring (spending/rewarding/certifying at known indices), a wire-level protobuf encode/decode roundtrip forcing total evaluation across all eras, Byron golden tests decoding real block CBOR, and a forged-annotation guard pinning Byron transaction ids to the on-chain bytes rather than a re-serialisation.
- E2E (IntersectMBO/cardano-node#6579): submits a payment transaction and a native-script minting transaction over gRPC, locates their blocks, fetches them via `FetchBlock` and asserts the mapped fields (hash, fee, inputs, outputs, validity, mint, witness set contents) against the submitted values, plus the `NOT_FOUND` and `INVALID_ARGUMENT` error paths.
- The mapping was cross-checked against other UTxO RPC implementations (Dolos/pallas, Blink Labs/gouroboros) for the byte-encoding conventions of addresses, asset names, hashes and witnesses.

Review commit by commit; every commit builds and passes tests on its own. The conversion lives in `Cardano.Rpc.Server.Internal.UtxoRpc.Type.{Tx,Certificate,Governance,Byron}` and the handler wiring in `Cardano.Rpc.Server.Internal.UtxoRpc.Sync`.

Run the tests:

```
cabal test cardano-rpc-test
TASTY_PATTERN='/RPC FetchBlock/' cabal test cardano-testnet-test
```

# Checklist

- [ ] Self-reviewed the diff


cardano-rpc/src/Cardano/Rpc/Server/Internal/UtxoRpc/Type/Certificate.hs
```diff
@@ -0,0 +1,320 @@
...
+
+import Cardano.Api.Address
+import Cardano.Api.Era
+import Cardano.Api.Ledger qualified as L
+import Cardano.Api.Serialise.Raw
+import Cardano.Rpc.Proto.Api.UtxoRpc.Query qualified as U5c
+import Cardano.Rpc.Proto.Api.UtxoRpc.Query qualified as UtxoRpc
+import Cardano.Rpc.Server.Internal.Orphans ()
+
+import Cardano.Ledger.BaseTypes qualified as L
```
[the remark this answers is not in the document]
carbolymer: bollocks

cardano-rpc/src/Cardano/Rpc/Server/Internal/UtxoRpc/Type/Tx.hs
```diff
@@ -0,0 +1,323 @@
...
+import Cardano.Rpc.Server.Internal.UtxoRpc.Type.PlutusData (scriptDataToUtxoRpcPlutusData)
+import Cardano.Rpc.Server.Internal.UtxoRpc.Type.Script (scriptToUtxoRpcScript)
+import Cardano.Rpc.Server.Internal.UtxoRpc.Type.TxEval (mkProtoRedeemer)
+import Cardano.Rpc.Server.Internal.UtxoRpc.Type.TxOutput
+  ( policyAssetsToUtxoRpcMultiassets
+  , txOutToUtxoRpcTxOutput
+  )
+
+import Cardano.Crypto.DSIGN.Class qualified as DSIGN
+import Cardano.Ledger.Api qualified as L
```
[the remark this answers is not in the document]
carbolymer: bollocks

cardano-rpc/test/cardano-rpc-test/Test/Cardano/Rpc/FetchBlockTx.hs
```diff
@@ -0,0 +1,563 @@
...
+import Cardano.Api.Plutus (ScriptData (..), toAlonzoData, unsafeHashableScriptData)
+import Cardano.Api.Serialise.Raw
+import Cardano.Api.Tx
+import Cardano.Api.Value (multiAssetToPolicyAssets)
+import Cardano.Rpc.Proto.Api.UtxoRpc.Query qualified as U5c
+import Cardano.Rpc.Server.Internal.UtxoRpc.Type
+
+import Cardano.Ledger.Api qualified as L
+import Cardano.Ledger.BaseTypes qualified as L
+import Cardano.Ledger.Hashes qualified as L (hashAnnotated)
```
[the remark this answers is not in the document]
carbolymer: bollocks

cardano-rpc/src/Cardano/Rpc/Server/Internal/UtxoRpc/Sync.hs
```diff
@@ -61,6 +66,14 @@ fetchBlockMethod request = do
     slotToUTCTime systemStart eraHistory slot
       & either (const throwPastHorizon) (pure . round . (* 1000) . utcTimeToPOSIXSeconds)
   let BlockHeader _ _ (BlockNo height) = getBlockHeader block
+      -- Byron transactions are not representable as cardano-api's 'Tx era',
+      -- so they are converted straight from the Byron ledger types
+      txs = case block of
+        ByronBlock consensusBlock ->
+          byronBlockTxs (byronBlockRaw consensusBlock)
```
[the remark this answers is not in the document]
carbolymer: Tracking this in: https://github.com/IntersectMBO/cardano-api/issues/1246

cardano-api/src/Cardano/Api/Tx/Internal/Body.hs
```diff
@@ -170,6 +170,9 @@ module Cardano.Api.Tx.Internal.Body
   , collectTxBodyScriptWitnesses
   , collectTxBodyScriptWitnessRequirements
   , toScriptIndex
+  , toScriptIndexAlonzo
+  , toScriptIndexConway
+  , toScriptIndexDijkstra
```
palas: I don't understand why these are needed. Why not use `toScriptIndex`?

cardano-rpc/src/Cardano/Rpc/Server/Internal/UtxoRpc/Type/Tx.hs
```diff
@@ -0,0 +1,322 @@
...
+      -- collateral return and total collateral exist from Babbage onwards;
+      -- the outer 'Maybe' is era support, the inner one field presence
+      collateralReturn :: Maybe (L.TxOut (ShelleyLedgerEra era))
+      collateralReturn = join $ body ^. L.collateralReturnTxBodyG
+      totalCollateral :: Maybe L.Coin
+      totalCollateral = join $ body ^. L.totalCollateralTxBodyG
+      collateral
+        | null collateralInputs
+        , Nothing <- collateralReturn
+        , Nothing <- totalCollateral =
```
palas, line 112: I didn't know about this syntax

cardano-rpc/src/Cardano/Rpc/Server/Internal/UtxoRpc/Type/Tx.hs
```diff
@@ -0,0 +1,322 @@
...
+        ShelleyBasedEraAlonzo -> []
+        ShelleyBasedEraBabbage -> []
+        ShelleyBasedEraConway -> conwayOnwardsProposals
+        ShelleyBasedEraDijkstra -> conwayOnwardsProposals
+      conwayOnwardsProposals
+        :: L.ConwayEraPParams (ShelleyLedgerEra era)
+        => [Proto UtxoRpc.GovernanceActionProposal]
+      conwayOnwardsProposals =
+        maybe [] (map proposalProcedureToUtxoRpcProposal . toList) $
+          body ^. L.proposalProceduresTxBodyG
```
palas: I would put this function as a subfunction of proposals, for scoping and for making it easier to parse visually

palas (commented):
Nice work! It all looks solid. I still want to look in more depth at the tests tomorrow, cause I didn't have time today. And I don't know about the exporting of per-era `toScriptIndex` functions, is it necessary?

cardano-api/src/Cardano/Api/Tx/Internal/Body.hs
carbolymer: To avoid using of eons. I can either:
1. Use `toScriptIndex`
2. Use eons, which are soft-deprecated at this point.

I'll have another look if I can avoid using those.

cardano-rpc/src/Cardano/Rpc/Server/Internal/UtxoRpc/Type/Tx.hs
carbolymer: plain old pattern guards https://wiki.haskell.org/Pattern_guard

cardano-api/src/Cardano/Api/Tx/Internal/Body.hs
carbolymer: I've removed those reexports, added experimental `toScriptPurposeIndex`. But I still have to use eons. 😞

cardano-rpc/test/cardano-rpc-test/Test/Cardano/Rpc/FetchBlockTx.hs
```diff
@@ -0,0 +1,576 @@
...
+    === shelleyBasedEraConstraints sbe (T.encodeUtf8 (serialiseAddress returnAddress))
+  collateralReturn ^. U5c.coin === inject (L.Coin returnCoin)
+
+  H.note_ "The proposal carries the injected deposit, return account, anchor and action"
+  [protoProposal] <- H.noteShow $ protoTx ^. U5c.proposals
+  protoProposal ^. U5c.deposit === inject (L.Coin proposalDeposit)
+  protoProposal ^. U5c.rewardAccount === serialiseToRawBytes stakeAddress
+  protoProposal ^. U5c.anchor . U5c.url === "https://example.com/anchor"
+  protoProposal ^. U5c.anchor . U5c.contentHash
+    === L.hashToBytes (L.extractHash (L.hashAnnotated (L.AnchorData "anchor data")))
```
palas: `"https://example.com/anchor"` and `"anchor data"` are duplicated constants which are meant to be the same (they are also defined in 288 and 289). At least the string literals for them should come from a single source

cardano-rpc/test/cardano-rpc-test/Test/Cardano/Rpc/ByronTx.hs
```diff
@@ -0,0 +1,235 @@
...
+  H.note_ "The first transaction carries inputs, outputs and witnesses, and no fee"
+  firstTx <- H.nothingFail $ listToMaybe mainBlockTxs
+  H.assertWith firstTx $ not . null . (^. U5c.inputs)
+  H.assertWith firstTx $ not . null . (^. U5c.outputs)
+  H.assertWith firstTx $ \protoTx ->
+    not (null (protoTx ^. U5c.witnesses . U5c.bootstrapWitnesses))
+      || not (null (protoTx ^. U5c.witnesses . U5c.vkeywitness))
+  firstTx ^. U5c.maybe'fee === Nothing
+
+  H.note_ "The boundary block fixture decodes and its header hash matches mainnet"
```
palas, line 206: I would argue before and after this, they should be separate tests/functions, or at least sub-functions, cause they are completely independent as far as I can see

cardano-rpc/test/cardano-rpc-test/Test/Cardano/Rpc/FetchBlockTx.hs
```diff
@@ -0,0 +1,576 @@
...
+
+-- | The name and presence check of the proto oneof arm to which
+-- 'txCertToUtxoRpcCertificate' must route a Conway ledger certificate.
+expectedCertificateArm
+  :: L.TxCert (ShelleyLedgerEra ConwayEra)
+  -> (String, Proto U5c.Certificate -> Bool)
+expectedCertificateArm = \case
+  L.ConwayTxCertDeleg delegCert ->
+    case delegCert of
+      L.ConwayRegCert _ L.SNothing ->
```
palas, line 571: It seems the `ConwayRegCert _ L.SNothing` and `ConwayUnRegCert _ L.SNothing` branches are not exercised by `genTx`. Not sure if it is worth tuning `genTx` so that they are

palas (approved):
Great job. It all looks very sound to me. I left a couple of really minor nits

cardano-api/src/Cardano/Api/Experimental/Plutus/Internal/IndexedPlutusScriptWitness.hs
```diff
@@ -143,6 +145,44 @@ toPlutusScriptPurpose index WitTxCert{} = L.mkCertifyingPurpose (L.AsIx index)
...
 
+-- | Classify a ledger redeemer pointer ('L.PlutusPurpose' 'L.AsIx') into the
+-- 'PlutusScriptPurpose' category it belongs to and the index within that
+-- category. This is the read direction, the inverse of 'toPlutusScriptPurpose'.
+--
+-- The 'AlonzoEraOnwards' witness both restricts callers to the eras that have
+-- plutus purposes and pins the era, so no constraints or type applications
+-- are needed at use sites.
+toPlutusScriptPurposeIndex
+  :: AlonzoEraOnwards era
```
Jimbo4350: This can be parameterized on `ShelleyBasedEra era` and return `Maybe (PlutusScriptPurpose, Word32)`. Where it is called has access to a `ShelleyBasedEra era` value.

cardano-rpc/src/Cardano/Rpc/Server/Internal/UtxoRpc/Type/Script.hs
```diff
@@ -76,17 +77,22 @@ utxoRpcNativeScriptToSimpleScript scriptRpc
...
-      defMessage & U5c.plutusV2 .~ serialiseToRawBytes ps
-    PlutusScript PlutusScriptV3 ps ->
-      defMessage & U5c.plutusV3 .~ serialiseToRawBytes ps
-    PlutusScript PlutusScriptV4 ps ->
-      defMessage & U5c.plutusV4 .~ serialiseToRawBytes ps
+  scriptToUtxoRpcScript script
+
+-- | Convert a script to the UTxO RPC 'UtxoRpc.Script' message, dispatching on
+-- the script language to select the corresponding oneof field.
+scriptToUtxoRpcScript :: Script lang -> Proto UtxoRpc.Script
```
Jimbo4350, line 105: It is possible to use `AnyScript era` instead here? `Script lang` is the old api.

cardano-rpc/src/Cardano/Rpc/Server/Internal/UtxoRpc/Type/Tx.hs
```diff
@@ -0,0 +1,327 @@
...
+
+-- | Extract the auxiliary scripts of a transaction, lifted to 'ScriptInEra'.
+-- Shelley auxiliary data carries no scripts, Allegra and Mary carry timelock
+-- scripts, and Alonzo onwards additionally carry Plutus scripts. The dispatch
+-- matches on the concrete eras because the auxiliary data and script types of
+-- each era family only line up at concrete eras.
+txAuxDataScripts
+  :: ShelleyBasedEra era
+  -> L.TxAuxData (ShelleyLedgerEra era)
+  -> [ScriptInEra era]
```
Jimbo4350: Is it possible to use `AnyScript era` here? `ScriptInEra era` is the old api.

Jimbo4350 (changes requested):
There is some use of the legacy api here that should be addressed.

cardano-rpc/test/cardano-rpc-test/Test/Cardano/Rpc/FetchBlockTx.hs
carbolymer: Defined values and reused.

cardano-rpc/test/cardano-rpc-test/Test/Cardano/Rpc/ByronTx.hs
carbolymer: Fair point. Split.

cardano-rpc/test/cardano-rpc-test/Test/Cardano/Rpc/FetchBlockTx.hs
carbolymer: You are right - they cannot be exercised: the Conway branch of the certificate generators goes through `mkRegDepositTxCert`/`mkUnRegDepositTxCert`, which the ledger hard-wires to `SJust`. Since the no-deposit form routes to a different proto arm (`stakeRegistration`/`stakeDeregistration` instead of `regCert`/`unregCert`), I have added a deterministic unit test of `txCertToUtxoRpcCertificate` covering both `SNothing` cases instead of tuning `genTx` - same coverage, without touching a generator shared by the other test suites.

cardano-api/src/Cardano/Api/Experimental/Plutus/Internal/IndexedPlutusScriptWitness.hs
carbolymer: `Maybe` is not necessary here. I think we can use SBE and live with `error`s in the impossible cases.

cardano-rpc/src/Cardano/Rpc/Server/Internal/UtxoRpc/Type/Script.hs
carbolymer: I don't think so, but we can avoid the old api. That would be a bigger change though. I've created a follow-up issue for that: https://github.com/IntersectMBO/cardano-api/issues/1264

cardano-rpc/src/Cardano/Rpc/Server/Internal/UtxoRpc/Type/Tx.hs
carbolymer: `txAuxDataScripts` now returns the ledger `Script`s directly - no `ScriptInEra` - and the caller converts them with the new ledger-direct `ledgerScriptToUtxoRpcScript`, which also handles Dijkstra guard scripts gracefully (#1263).

`AnyScript era` does not fit this direction of conversion: its plutus constructor wraps `PlutusRunnable`, so building one from block data would require `decodePlutusRunnable` - a full script deserialisation with an error path - when the proto only needs the language tag and the raw bytes the ledger already provides. Its native constructor preserves the raw per-era ledger representations (`MultiSig`/`Timelock`/`DijkstraNativeScript`), so the per-era dispatch would not go away either. Converting from the ledger types directly is total and needs no old-API or experimental wrappers. The remaining old-API usage in cardano-rpc is tracked in #1264.

Jimbo4350 (approved):

PR #1248, opened 2026-07-08
Add writeFileTextEnvelopeWithOwnerPermissions

palas:
# Context

`writeFileTextEnvelope` creates files with the default permissions, which is not appropriate for sensitive data such as signing keys.

This PR adds `writeFileTextEnvelopeWithOwnerPermissions`, a variant that creates the file with owner-only permissions. It also adds a haddock note to `writeFileTextEnvelope` pointing users to the new function for sensitive data.

# How to trust this PR

- The new function is a one-line wrapper around the existing `writeLazyByteStringFileWithOwnerPermissions`, which is already used elsewhere for this exact purpose.
- The rest of the diff is re-exports and haddocks.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-api/src/Cardano/Api/Serialise/TextEnvelope/Internal.hs
+++ b/cardano-api/src/Cardano/Api/Serialise/TextEnvelope/Internal.hs
@@ -284,7 +284,7 @@ deserialiseFromTextEnvelopeJSONAnyOf types bs =
 -- | Write a value to a file in the text envelope format.
 --
 -- Note that this does /not/ set conservative file permissions: the file is
--- created with the default permissions. When writing sensitive data such as
+-- created with the default permissions which should be restricted with `umask`. When writing sensitive data such as
 -- signing keys, use 'writeFileTextEnvelopeWithOwnerPermissions' instead,
 -- which tries to restrict access to the file owner where the platform
 -- supports it (see its documentation for the exact guarantees).
--- a/cardano-api/src/Cardano/Api/Serialise/TextEnvelope/Internal.hs
+++ b/cardano-api/src/Cardano/Api/Serialise/TextEnvelope/Internal.hs
@@ -303,7 +303,8 @@ writeFileTextEnvelope outputFile mbDescr a =
 -- * On POSIX systems, the file is created with @0600@ permissions (read
 --   and write for the file owner only, further filtered by the process's
 --   @umask@), and its ownership is set to the current (real) user. If the
---   file already exists, its permission bits are left unchanged.
+--   file already exists, it is truncated, but its permission bits are
+--   left unchanged.
 --
 -- * On Windows, the contents are written to a freshly created temporary
 --   file which is then renamed to the target path. This guarantees the
--- a/cardano-api/src/Cardano/Api/Serialise/TextEnvelope/Internal.hs
+++ b/cardano-api/src/Cardano/Api/Serialise/TextEnvelope/Internal.hs
@@ -286,7 +286,8 @@ deserialiseFromTextEnvelopeJSONAnyOf types bs =
 -- Note that this does /not/ set conservative file permissions: the file is
 -- created with the default permissions. When writing sensitive data such as
 -- signing keys, use 'writeFileTextEnvelopeWithOwnerPermissions' instead,
--- which tries to restricts access to the file owner (depending on the platform).
+-- which tries to restrict access to the file owner where the platform
+-- supports it (see its documentation for the exact guarantees).
 writeFileTextEnvelope
   :: HasTextEnvelope a
   => File content Out
@@ -296,9 +297,21 @@ writeFileTextEnvelope
 writeFileTextEnvelope outputFile mbDescr a =
   writeLazyByteStringFile outputFile (textEnvelopeToJSON mbDescr a)
 
--- | Like 'writeFileTextEnvelope', but the file is created with owner-only
--- permissions, i.e. it is readable and writable by the file owner only
--- (depending on the platform).
+-- | Like 'writeFileTextEnvelope', but the file is created so that only its
+-- owner has access to it, to the extent the platform allows it:
+--
+-- * On POSIX systems, the file is created with @0700@ permissions
+--   (@ownerModes@, further filtered by the process's @umask@), so no group
+--   or other access, and its ownership is set to the current (real) user.
+--   If the file already exists, its permission bits are left unchanged.
+--
+-- * On Windows, the contents are written to a freshly created temporary
+--   file which is then renamed to the target path. This guarantees the
+--   file is owned by the current user, but no explicit ACL is set: the
+--   file inherits the access control list of the target directory.
+--
+-- * On WASM, this is currently a no-op: no file is written at all.
+--
 -- Use this when writing sensitive data such as signing keys.
 writeFileTextEnvelopeWithOwnerPermissions
   :: HasTextEnvelope a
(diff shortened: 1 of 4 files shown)


cardano-api/src/Cardano/Api/Serialise/TextEnvelope/Internal.hs
```diff
@@ -280,6 +281,12 @@ deserialiseFromTextEnvelopeJSONAnyOf
 deserialiseFromTextEnvelopeJSONAnyOf types bs =
   decodeTextEnvelopeJSON bs >>= deserialiseFromTextEnvelopeAnyOf types
 
+-- | Write a value to a file in the text envelope format.
+--
+-- Note that this does /not/ set conservative file permissions: the file is
+-- created with the default permissions. When writing sensitive data such as
+-- signing keys, use 'writeFileTextEnvelopeWithOwnerPermissions' instead,
+-- which restricts access to the file owner.
```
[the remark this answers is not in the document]
palas: Addressed

cardano-api/src/Cardano/Api/Serialise/TextEnvelope/Internal.hs
```diff
@@ -289,6 +296,18 @@ writeFileTextEnvelope
 writeFileTextEnvelope outputFile mbDescr a =
   writeLazyByteStringFile outputFile (textEnvelopeToJSON mbDescr a)
 
+-- | Like 'writeFileTextEnvelope', but the file is created with owner-only
+-- permissions, i.e. it is readable and writable by the file owner only.
+-- Use this when writing sensitive data such as signing keys.
```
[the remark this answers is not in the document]
palas: Addressed

cardano-api/src/Cardano/Api/Serialise/TextEnvelope/Internal.hs
```diff
@@ -289,6 +296,19 @@ writeFileTextEnvelope
 writeFileTextEnvelope outputFile mbDescr a =
   writeLazyByteStringFile outputFile (textEnvelopeToJSON mbDescr a)
 
+-- | Like 'writeFileTextEnvelope', but the file is created with owner-only
+-- permissions, i.e. it is readable and writable by the file owner only
+-- (depending on the platform).
```
carbolymer: So `600`? "depending on the platform" - It would be nice to be more specific here.

carbolymer (approved):
LGTM. I feel that this issue is mostly user error. The users should use umask `077` instead of some OS default one. I'm not sure about windows.

Jimbo4350 (approved):

palas:
> LGTM. I feel that this issue is mostly user error. The users should use umask 077 instead of some OS default one. I'm not sure about windows.

Yes, I agree, this is just making it more obvious for them

cardano-api/src/Cardano/Api/Serialise/TextEnvelope/Internal.hs
palas: Huh, apparently it is actually `700`, that doesn't make a lot of sense, does it?

palas: I've changed that now, and added a comment

cardano-api/src/Cardano/Api/Serialise/TextEnvelope/Internal.hs
```diff
@@ -280,6 +281,13 @@ deserialiseFromTextEnvelopeJSONAnyOf
 deserialiseFromTextEnvelopeJSONAnyOf types bs =
   decodeTextEnvelopeJSON bs >>= deserialiseFromTextEnvelopeAnyOf types
 
+-- | Write a value to a file in the text envelope format.
+--
+-- Note that this does /not/ set conservative file permissions: the file is
+-- created with the default permissions. When writing sensitive data such as
```
carbolymer:
```suggestion
-- created with the default permissions which should be restricted with `umask`. When writing sensitive data such as
```
or something along those lines

cardano-api/src/Cardano/Api/Serialise/TextEnvelope/Internal.hs
```diff
@@ -289,6 +297,31 @@ writeFileTextEnvelope
 writeFileTextEnvelope outputFile mbDescr a =
   writeLazyByteStringFile outputFile (textEnvelopeToJSON mbDescr a)
 
+-- | Like 'writeFileTextEnvelope', but the file is created so that only its
+-- owner has access to it, to the extent the platform allows it:
+--
+-- * On POSIX systems, the file is created with @0600@ permissions (read
+--   and write for the file owner only, further filtered by the process's
+--   @umask@), and its ownership is set to the current (real) user. If the
+--   file already exists, its permission bits are left unchanged.
```
carbolymer: The platform notes here check out against the three implementations (nice to have them written down at last), but there is one gap hiding behind "If the file already exists": on POSIX this path does not truncate.

`handleFileForWritingWithOwnerPermissionImpl` opens the fd with `defaultFileFlags{creat = Just ownerReadWriteMode}`, and `defaultFileFlags` has `trunc = False`, so no `O_TRUNC`.
The callback then `hPut`s from offset 0 and nothing ever shrinks the file.
Overwriting an existing file with shorter content leaves the old tail bytes in place and still returns `Right ()`:

```haskell
ghci> Prelude.writeFile p (replicate 100 'A')
ghci> writeLazyByteStringFileWithOwnerPermissions (File p) (L8.pack "short")
Right ()
ghci> Prelude.readFile p
"shortAAAAAAAAAAAA...A"   -- "short" followed by 95 stale bytes
```

So "Like 'writeFileTextEnvelope'" does not quite hold for pre-existing files: `writeFileTextEnvelope` goes through `LBS.writeFile`, which truncates.
A key file rewritten through this function can end up as corrupt JSON while reporting success.

To be clear, this is pre-existing behaviour of `handleFileForWritingWithOwnerPermissionImpl` (it affects `writeTextFileWithOwnerPermissions` and friends too), not something this PR introduces.
But since this PR writes the contract down, it would be good to either:

- set `trunc = True` in the `WriteOnly` branch of `openFileDescriptor` (both CPP branches), or
- extend this sentence to say that existing files are not truncated.

If you go for truncation, one trade-off to keep in mind: with `O_TRUNC` a write that fails midway destroys the previous file contents, whereas the Windows implementation's temp-file-plus-rename keeps the old file intact until the new one is complete.
Doing the same temp+rename on POSIX would fix the truncation problem and the partial-write problem at once, at the cost of the "permission bits of an existing file are left unchanged" behaviour (the replacement would always end up 0600, which is arguably what you want for keys anyway).

carbolymer (approved):
There is a bug with overwriting files - we're not truncating them. Since it already existed before this change, I'm not blocking this PR.

cardano-api/src/Cardano/Api/Serialise/TextEnvelope/Internal.hs
palas: Applied here: https://github.com/IntersectMBO/cardano-api/pull/1248/commits/bd819d231721e6e293b77161aea2a5705324e1f3

cardano-api/src/Cardano/Api/Serialise/TextEnvelope/Internal.hs
palas: Ok, I've applied the truncation in: https://github.com/IntersectMBO/cardano-api/pull/1248/changes/a8fd42c17997865b2a943dfba271a31779907f47

And I've created an issue for the `temp+rename` here: https://github.com/IntersectMBO/cardano-api/issues/1254

PR #1249, opened 2026-07-08
Return an error instead of throwing on invalid UTF-8 in deserialisation helpers

palas:
# Context

`deserialiseInput`, `deserialiseInputAnyOf` and `deserialiseAnyVerificationKeyBech32` (and thereby `deserialiseAnyVerificationKey`) used `Text.decodeUtf8`, which throws an impure exception when the input is not valid UTF-8. For example, `deserialiseInput` on the bytes `"\xc3\x28"` threw instead of returning `Left InputInvalidError`.

This PR switches them to `Text.decodeUtf8'` and treats invalid UTF-8 as a normal decoding failure.

# How to trust this PR

- Invalid UTF-8 can never be valid Bech32, so mapping the decode failure to the existing "not valid Bech32" path (`DeserialiseInputErrorFormatMismatch` / `Bech32DecodingError`) does not change behaviour for any valid input.
- The three changed sites are mechanical: wrap the old code in the `Right` case of `decodeUtf8'`.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-api/src/Cardano/Api/Key/Internal/SomeAddressVerificationKey.hs
+++ b/cardano-api/src/Cardano/Api/Key/Internal/SomeAddressVerificationKey.hs
@@ -28,11 +28,12 @@ import Cardano.Api.Serialise.TextEnvelope.Internal
 import Cardano.Chain.Common qualified as Common
 import Cardano.Crypto.Signing qualified as Crypto
 
-import Codec.Binary.Bech32 qualified as Bech32
+import Control.Exception (displayException)
 import Data.Aeson qualified as Aeson
 import Data.Bifunctor (first)
 import Data.ByteString (ByteString)
 import Data.Text (Text)
+import Data.Text qualified as Text
 import Data.Text.Encoding qualified as Text
 import Formatting (build, sformat, (%))
 
@@ -146,7 +147,7 @@ deserialiseAnyVerificationKeyBech32
 deserialiseAnyVerificationKeyBech32 bs =
   case Text.decodeUtf8' bs of
     -- The input is not valid UTF-8, so it cannot be valid Bech32.
-    Left _ -> Left $ Bech32DecodingError $ Bech32.StringToDecodeContainsInvalidChars []
+    Left err -> Left $ Bech32InvalidUtf8 $ Text.pack $ displayException err
     Right text -> deserialiseAnyOfFromBech32 allBech32VerKey text
  where
   allBech32VerKey
--- a/cardano-api/src/Cardano/Api/Key/Internal/SomeAddressVerificationKey.hs
+++ b/cardano-api/src/Cardano/Api/Key/Internal/SomeAddressVerificationKey.hs
@@ -28,12 +28,10 @@ import Cardano.Api.Serialise.TextEnvelope.Internal
 import Cardano.Chain.Common qualified as Common
 import Cardano.Crypto.Signing qualified as Crypto
 
-import Control.Exception (displayException)
 import Data.Aeson qualified as Aeson
 import Data.Bifunctor (first)
 import Data.ByteString (ByteString)
 import Data.Text (Text)
-import Data.Text qualified as Text
 import Data.Text.Encoding qualified as Text
 import Formatting (build, sformat, (%))
 
@@ -147,7 +145,7 @@ deserialiseAnyVerificationKeyBech32
 deserialiseAnyVerificationKeyBech32 bs =
   case Text.decodeUtf8' bs of
     -- The input is not valid UTF-8, so it cannot be valid Bech32.
-    Left err -> Left $ Bech32InvalidUtf8 $ Text.pack $ displayException err
+    Left err -> Left $ Bech32InvalidUtf8 err
     Right text -> deserialiseAnyOfFromBech32 allBech32VerKey text
  where
   allBech32VerKey
--- a/cardano-api/src/Cardano/Api/Key/Internal/SomeAddressVerificationKey.hs
+++ b/cardano-api/src/Cardano/Api/Key/Internal/SomeAddressVerificationKey.hs
@@ -28,6 +28,7 @@ import Cardano.Api.Serialise.TextEnvelope.Internal
 import Cardano.Chain.Common qualified as Common
 import Cardano.Crypto.Signing qualified as Crypto
 
+import Codec.Binary.Bech32 qualified as Bech32
 import Data.Aeson qualified as Aeson
 import Data.Bifunctor (first)
 import Data.ByteString (ByteString)
@@ -142,8 +143,11 @@ deserialiseAnyVerificationKey bs =
 
 deserialiseAnyVerificationKeyBech32
   :: ByteString -> Either Bech32DecodeError SomeAddressVerificationKey
-deserialiseAnyVerificationKeyBech32 =
-  deserialiseAnyOfFromBech32 allBech32VerKey . Text.decodeUtf8
+deserialiseAnyVerificationKeyBech32 bs =
+  case Text.decodeUtf8' bs of
+    -- The input is not valid UTF-8, so it cannot be valid Bech32.
+    Left _ -> Left $ Bech32DecodingError $ Bech32.StringToDecodeContainsInvalidChars []
+    Right text -> deserialiseAnyOfFromBech32 allBech32VerKey text
  where
   allBech32VerKey
     :: [FromSomeType SerialiseAsBech32 SomeAddressVerificationKey]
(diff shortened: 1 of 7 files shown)


cardano-api/src/Cardano/Api/Key/Internal/SomeAddressVerificationKey.hs
```diff
@@ -28,6 +28,7 @@ import Cardano.Api.Serialise.TextEnvelope.Internal
 import Cardano.Chain.Common qualified as Common
 import Cardano.Crypto.Signing qualified as Crypto
 
+import Codec.Binary.Bech32 qualified as Bech32
```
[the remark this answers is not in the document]
palas: It is a possibility, but lenient decoding may decode valid data from an invalid bytestring, so it could be misleading

cardano-api/src/Cardano/Api/Key/Internal/SomeAddressVerificationKey.hs
```diff
@@ -142,8 +143,11 @@ deserialiseAnyVerificationKey bs =
 
 deserialiseAnyVerificationKeyBech32
   :: ByteString -> Either Bech32DecodeError SomeAddressVerificationKey
-deserialiseAnyVerificationKeyBech32 =
-  deserialiseAnyOfFromBech32 allBech32VerKey . Text.decodeUtf8
+deserialiseAnyVerificationKeyBech32 bs =
+  case Text.decodeUtf8' bs of
+    -- The input is not valid UTF-8, so it cannot be valid Bech32.
+    Left _ -> Left $ Bech32DecodingError $ Bech32.StringToDecodeContainsInvalidChars []
+    Right text -> deserialiseAnyOfFromBech32 allBech32VerKey text
```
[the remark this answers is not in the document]
palas: It is a possibility, but lenient decoding may decode valid data from an invalid bytestring, so it could be misleading

cardano-api/src/Cardano/Api/Serialise/DeserialiseAnyOf.hs
```diff
@@ -140,12 +136,16 @@ deserialiseInput acceptedFormats inputBs =
...
+    case Text.decodeUtf8' inputBs of
+      -- The input was not valid UTF-8, so it cannot be valid Bech32.
+      Left _ -> DeserialiseInputErrorFormatMismatch
+      Right inputText ->
+        case deserialiseFromBech32 inputText of
+          Right res -> DeserialiseInputSuccess res
+          -- The input was not valid Bech32.
+          Left (Bech32DecodingError _) -> DeserialiseInputErrorFormatMismatch
+          -- The input was valid Bech32, but some other error occurred.
+          Left err -> DeserialiseInputError $ InputBech32DecodeError err
```
carbolymer: what do you think about:
```haskell
    deserialiseBech32 = decodeText inputBs >>= fromBech32
     where
      decodeText =
        first (const DeserialiseInputErrorFormatMismatch)
          . Text.decodeUtf8'
      fromBech32 text =
        case deserialiseFromBech32 text of
          Right res -> Right $ DeserialiseInputSuccess res
          Left (Bech32DecodingError _) -> Right DeserialiseInputErrorFormatMismatch
          Left err -> Right $ DeserialiseInputError $ InputBech32DecodeError err
```
separates the flow from the error handling a bit

(this also applies to the other change in this PR)

carbolymer (approved):

cardano-api/src/Cardano/Api/Key/Internal/SomeAddressVerificationKey.hs
```diff
@@ -142,8 +143,11 @@ deserialiseAnyVerificationKey bs =
 
 deserialiseAnyVerificationKeyBech32
   :: ByteString -> Either Bech32DecodeError SomeAddressVerificationKey
-deserialiseAnyVerificationKeyBech32 =
-  deserialiseAnyOfFromBech32 allBech32VerKey . Text.decodeUtf8
+deserialiseAnyVerificationKeyBech32 bs =
+  case Text.decodeUtf8' bs of
+    -- The input is not valid UTF-8, so it cannot be valid Bech32.
+    Left _ -> Left $ Bech32DecodingError $ Bech32.StringToDecodeContainsInvalidChars []
```
Jimbo4350: We should render the `UnicodeException` error.

Jimbo4350 (changes requested):
We shouldn't throw away the `UnicodeException` error. It needs to be rendered.

cardano-api/src/Cardano/Api/Key/Internal/SomeAddressVerificationKey.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-api/pull/1249/changes/3fec64a55e6d37244928d4de112c380a8f087567

cardano-api/src/Cardano/Api/Serialise/DeserialiseAnyOf.hs
palas: Sounds good, less nesting. Applied it here: https://github.com/IntersectMBO/cardano-api/pull/1249/changes/bc818047a0e79948dc7e7fbbef56419cf48c5379

cardano-api/src/Cardano/Api/Serialise/DeserialiseAnyOf.hs
```diff
@@ -139,13 +135,19 @@ deserialiseInput acceptedFormats inputBs =
...
-      -- The input was not valid Bech32.
-      Left (Bech32DecodingError _) -> DeserialiseInputErrorFormatMismatch
-      -- The input was valid Bech32, but some other error occurred.
-      Left err -> DeserialiseInputError $ InputBech32DecodeError err
+  deserialiseBech32 = either id fromBech32 $ decodeText inputBs
+   where
+    -- The input was not valid UTF-8, so it cannot be valid Bech32.
+    decodeText =
+      first (const DeserialiseInputErrorFormatMismatch)
+        . Text.decodeUtf8'
```
Jimbo4350, line 143: We are still throwing away the exception error here.

cardano-api/src/Cardano/Api/Serialise/DeserialiseAnyOf.hs
```diff
@@ -208,13 +207,19 @@ deserialiseInputAnyOf bech32Types textEnvTypes inputBs =
...
-    case deserialiseAnyOfFromBech32 bech32Types inputText of
-      Right res -> DeserialiseInputSuccess res
-      -- The input was not valid Bech32.
-      Left (Bech32DecodingError _) -> DeserialiseInputErrorFormatMismatch
-      -- The input was valid Bech32, but some other error occurred.
-      Left err -> DeserialiseInputError $ InputBech32DecodeError err
+  deserialiseBech32 = either id fromBech32 $ decodeText inputBs
+   where
+    -- The input was not valid UTF-8, so it cannot be valid Bech32.
+    decodeText =
```
Jimbo4350, line 213: We are still throwing away the exception error here.

cardano-api/src/Cardano/Api/Serialise/Bech32.hs
```diff
@@ -149,6 +149,9 @@ data Bech32DecodeError
       -- ^ Expected header
       !Text
       -- ^ Unexpected header
+  | -- | The input is not valid UTF-8, so it cannot be a Bech32-encoded
+    -- string. The field contains the rendered UTF-8 decoding error.
+    Bech32InvalidUtf8 !Text
```
carbolymer: Why the field is `Text` and not the exception itself? I'd suggest moving of rendering of the exception to `Pretty` instance instead of keeping it in two places.

Jimbo4350 (approved):
2 comments to address.

cardano-api/src/Cardano/Api/Serialise/DeserialiseAnyOf.hs
palas: It is because the error is dropped anyway in line 182 as part of `deserialiseInputAnyOf`, it is just used to discard a format and try another until the right one is found

cardano-api/src/Cardano/Api/Serialise/DeserialiseAnyOf.hs
palas: Same with line 120 here.

cardano-api/src/Cardano/Api/Serialise/Bech32.hs
palas: It is because [UnicodeException](https://hackage-content.haskell.org/package/text-2.1.4/docs/Data-Text-Encoding-Error.html#t:UnicodeException) doesn't implement `Data` but `Bech32DecodeError` derives it. But I can do it with an orphan

palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/1249/changes/b824fdbdcabcc64ab4c87422bbc7a5a0cf70ea5b

PR #1258, opened 2026-07-20
cardano-rpc: Fix TxOutput.address and Datum.hash wire encoding

### LINKED ISSUE #1246: gRPC: `bytes` fields encoded as text or wrong payload (`TxOutput.address`, `Datum.hash`)

carbolymer:
# Context

An audit of the proto `bytes` fields populated by the UTxO RPC server found two fields carrying the wrong encoding.
`TxOutput.address` was populated with human-readable address text (bech32 for Shelley, base58 for Byron) instead of raw ledger address bytes.
`Datum.hash` was populated with the datum's CBOR instead of its blake2b-256 hash when the datum was inline.
Both diverged from the other UTxO RPC implementations (Dolos via pallas-utxorpc, Blink Labs cardano-node-api) and, for the address, from our own predicate and reward-account encoding.
The server is unreleased, so the wire format can still change without breaking clients.

Fixes #1246

Stacked on #1222 - only the last two commits belong to this PR.

# How to trust this PR

`txOutToUtxoRpcTxOutput` now uses `serialiseToRawBytes` for the address (raw header plus payload bytes, Byron included) and `hashScriptDataBytes` for the inline datum hash, and `utxoRpcTxOutputToTxOut` decodes the address with `deserialiseFromRawBytes` in lockstep.
Address serialisation is now uniform across cardano-rpc: transaction outputs, `exact_address` predicates and reward accounts all use raw on-chain bytes in both directions, so a client can echo `TxOutput.address` into a `SearchUtxos` predicate and get matches.
The new `hprop_tx_output_wire_format` property pins the wire format directly; the existing round-trip property cannot catch the inline datum bug because decoding never reads `Datum.hash` when `original_cbor` is present.
Run `cabal test cardano-rpc:cardano-rpc-test` to verify; all 78 tests pass.
The cardano-node E2E test `Cardano.Testnet.Test.Rpc.FetchBlock` asserts the same encoding and changes in lockstep (branch `mgalazyn/fix-grpc-bytes-fields` in cardano-node).

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/.changes/20260720_cardano_rpc_txoutput_raw_bytes.yml
+++ b/.changes/20260720_cardano_rpc_txoutput_raw_bytes.yml
@@ -1,0 +1,6 @@
+project: cardano-rpc
+pr: 1258
+kind:
+  - bugfix
+description: |
+  Fix wire encoding of two TxOutput fields: address now carries raw ledger address bytes instead of bech32/base58 text, and Datum.hash now carries the 32-byte datum hash instead of the datum CBOR for inline datums, matching other UTxO RPC implementations.
--- a/cardano-rpc/cardano-rpc.cabal
+++ b/cardano-rpc/cardano-rpc.cabal
@@ -90,6 +90,7 @@ library
   build-depends:
     aeson,
     base,
+    base16-bytestring,
     bytestring,
     cardano-api >=11.2,
     cardano-binary,
--- a/cardano-rpc/src/Cardano/Rpc/Server/Internal/UtxoRpc/Type/TxOutput.hs
+++ b/cardano-rpc/src/Cardano/Rpc/Server/Internal/UtxoRpc/Type/TxOutput.hs
@@ -14,7 +14,6 @@ module Cardano.Rpc.Server.Internal.UtxoRpc.Type.TxOutput
   )
 where
 
-import Cardano.Api.Address
 import Cardano.Api.Era
 import Cardano.Api.Error
 import Cardano.Api.Experimental.Era
@@ -36,9 +35,9 @@ import Cardano.Binary qualified as CBOR
 
 import RIO hiding (toList)
 
+import Data.ByteString.Base16 qualified as Base16
+import Data.ByteString.Char8 qualified as BSC
 import Data.ProtoLens (defMessage)
-import Data.Text qualified as T
-import Data.Text.Encoding qualified as T
 import GHC.IsList
 import Network.GRPC.Spec
 
@@ -123,12 +122,12 @@ txOutToUtxoRpcTxOutput sbe (TxOut addressInEra txOutValue datum script) = do
         TxOutDatumInline _ hashableScriptData ->
           Just $
             defMessage
-              & U5c.hash .~ serialiseToCBOR hashableScriptData
+              & U5c.hash .~ serialiseToRawBytes (hashScriptDataBytes hashableScriptData)
               & U5c.payload .~ scriptDataToUtxoRpcPlutusData (getScriptData hashableScriptData)
               & U5c.originalCbor .~ getOriginalScriptDataBytes hashableScriptData
 
   defMessage
-    & U5c.address .~ T.encodeUtf8 (shelleyBasedEraConstraints sbe $ serialiseAddress addressInEra)
+    & U5c.address .~ shelleyBasedEraConstraints sbe (serialiseToRawBytes addressInEra)
     & U5c.coin .~ inject (L.unCoin (txOutValueToLovelace txOutValue))
     & U5c.assets .~ multiAsset
     & U5c.maybe'datum .~ datumRpc
@@ -143,11 +142,15 @@ utxoRpcTxOutputToTxOut
   -> m (TxOut CtxUTxO era)
 utxoRpcTxOutputToTxOut txOutput = do
   let era = useEra @era
-  addrUtf8 <- liftEitherError $ T.decodeUtf8' (txOutput ^. U5c.address)
+  let addressBytes = txOutput ^. U5c.address
+      annotateError (SerialiseAsRawBytesError msg) =
+        SerialiseAsRawBytesError $
+          msg <> ", address (hex): " <> BSC.unpack (Base16.encode addressBytes)
   address <-
-    maybe (throwM . stringException $ "Cannot decode address: " <> T.unpack addrUtf8) pure $
-      obtainCommonConstraints era $
-        deserialiseAddress asType addrUtf8
+    obtainCommonConstraints era $
+      liftEitherError $
+        first annotateError $
+          deserialiseFromRawBytes asType addressBytes
   datum <-
     case txOutput ^. U5c.maybe'datum of
       Just datumRpc ->
--- a/cardano-rpc/test/cardano-rpc-test/Test/Cardano/Rpc/FetchBlockTx.hs
+++ b/cardano-rpc/test/cardano-rpc-test/Test/Cardano/Rpc/FetchBlockTx.hs
@@ -8,8 +8,7 @@ module Test.Cardano.Rpc.FetchBlockTx where
 
 import Cardano.Api (SlotNo (..))
 import Cardano.Api.Address
-  ( serialiseAddress
-  , toShelleyAddr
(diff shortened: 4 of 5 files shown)


Jimbo4350 (approved):
LGTM

PR #1266, opened 2026-07-24
Fail on invalid computed collateral instead of building a rejected transaction

### LINKED ISSUE #1261: [BUG] - `transaction build` does not validate min UTxO value of the collateral return output
**Internal/External**

**Area**

**Summary**
`makeTransactionBodyAutoBalance` (used by `cardano-cli transaction build`)
does not check that the collateral return output meets the minimum UTxO
value. It builds transactions with an arbitrarily small (even 1 lovelace)
return collateral output. Such a transaction is guaranteed to be rejected
by the ledger at submission with `BabbageOutputTooSmallUTxO`, so the error
surfaces at submit time instead of build time. The same gap exists in
`estimateBalancedTxBody`.

**Steps to reproduce**
1. Create a transaction that uses a Plutus script and provide collateral
   input(s) whose total value is only slightly above the required
   collateral (`ceil(fee * collateralPercentage / 100)`), so that the
   difference is greater than zero but below the min UTxO value of the
   return collateral output.
2. Build the transaction with `cardano-cli transaction build` (with a
   change address, no explicit `--tx-total-collateral` /
   `--tx-out-return-collateral`).
3. Building succeeds and produces a tx body with a return collateral
   output below the min UTxO value.
4. Submit the transaction.

**Actual behavior:** build succeeds; submission fails with `BabbageOutputTooSmallUTxO`.

The same applies to an explicitly provided return collateral output
(`--tx-out-return-collateral`): it is passed through to the tx body with
no validation.

**Expected behavior**
`transaction build` fails with an error analogous to
`TxBodyErrorMinUTxONotMet`, the same way it fails for regular outputs
below the min UTxO value.

**System info (please complete the following information):**
- OS Name: Fedora
- OS Version 42

Observed with cardano-cli 11.1.0.0 / cardano-node 11.x on a local testnet
(protocol version 11, Conway era), during IntersectMBO/cardano-node-tests
system-test runs.

**Additional context**
One design question: instead of failing, the balancer could fold a
too-small return amount into the total collateral (i.e. burn the dust and
omit the return output), which would make more transactions buildable at
the cost of slightly higher potential collateral loss.

`estimateBalancedTxBody` uses the same `calcReturnAndTotalCollateral` and
likewise runs its min UTxO check only on the regular outputs, so the
fee-estimation-based balancing path has the identical gap.

palas:
# Context

This is the second of three stacked PRs fixing #1261: the balancing functions
(`makeTransactionBodyAutoBalance` and `estimateBalancedTxBody`, in both the traditional and the
experimental API) could build transactions that the ledger rejects at submission time because of
their collateral. Each PR contains a pair of commits — regression tests that fail, then the fix
that makes them pass — and each PR is based on the previous one, so every diff shows only its own
two commits.

🗺️ **Map** — you are here: 2️⃣

|  |  |  |
|---|---|---|
| 🚪 | [#1261](https://github.com/IntersectMBO/cardano-api/issues/1261) | the issue: `transaction build` does not validate min UTxO value of the collateral return output |
| 1️⃣ | #1265 | remove collateral from transactions without Plutus scripts |
| 2️⃣ | **(this PR)** | fail on invalid computed collateral instead of building a rejected transaction |
| 3️⃣ | #1267 | fold return collateral dust into the total collateral |

**This PR:** when a transaction with Plutus scripts provides collateral inputs but no explicit
collateral fields, `calcReturnAndTotalCollateral` (traditional and experimental API) computes the
return collateral output and the total collateral. It could compute collateral that the ledger
rejects, in two ways: the ada left for the return collateral output after covering the required
collateral (150% of the fee) could be below the output's own minimum UTxO value — the direct
#1261 report — and insufficient collateral was silently encoded as "no collateral fields", the
same value returned for transactions that use no collateral at all. Both copies of
`calcReturnAndTotalCollateral` now return the collateral fields in an `Either` with the new
`CollateralError` type, which makes these cases explicit; the computed return collateral output
is checked against its minimum UTxO value; and the balancing functions fail with the new
`TxBodyErrorCollateral` error instead of building a rejected transaction. (For
ada-only collateral, failing on the minimum UTxO value is refined into folding by the next PR.)

# How to trust this PR

- The first commit adds two regression tests (traditional and experimental
  `makeTransactionBodyAutoBalance`) with a token-carrying collateral input holding exactly its
  minimum UTxO value, so the leftover ada is necessarily below the minimum UTxO value of the
  return collateral output. On the base branch they fail; the second commit makes them pass. The
  tests check the actually-produced return collateral output against `calculateMinimumUTxO`
  rather than just expecting failure, so they stay truthful if protocol parameters change.
- `estimateBalancedTxBody` is not covered by these two tests because it receives the potential
  collateral as a plain lovelace amount and cannot see tokens (this assumption is documented in
  the next PR); it gets the same new error paths through the shared `calcReturnAndTotalCollateral`.
- The pre-existing property test `prop_calcReturnAndTotalCollateral` is adapted to the new result
  type (including an assertion that both collateral fields come back empty if and only if the
  transaction has no collateral inputs), and golden files cover the two new error messages.
- This PR is breaking: `calcReturnAndTotalCollateral` returns the collateral fields in an
  `Either CollateralError` instead of a plain pair, and `TxBodyErrorAutoBalance` gains one
  constructor.
- The third commit addresses the review feedback: it replaces the `CollateralComputation` sum
  type with `Either CollateralError`, so the reviewed commits stay intact and the delta is
  visible on its own.

# Checklist

- [x] Self-reviewed the diff


cardano-api/src/Cardano/Api/Tx/Internal/Fee.hs
```diff
@@ -1239,6 +1290,36 @@ hasPlutusScriptWitnesses sbe txbodycontent =
           collectTxBodyScriptWitnesses sbe txbodycontent
       ]
 
+-- | Result of 'calcReturnAndTotalCollateral'.
+data CollateralComputation ret tot
```
carbolymer: I think this type is a hand-rolled `Either` and would be simpler as one.

Two of the four constructors are failures that every caller immediately converts to `Left`/`throwError`, and the other two are successes that every caller immediately converts to a pair.
All four call sites perform the same mechanical 4-way unbundling.

Proposal:

```haskell
-- | Why the collateral fields could not be computed: the ledger would
-- reject a transaction built with them.
data CollateralError
  = -- | The ada in the collateral inputs does not cover the required
    -- collateral (fee * collateralPercentage / 100).
    InsufficientCollateral
      L.Coin
      -- ^ Ada available in the collateral inputs
      L.Coin
      -- ^ Required collateral
  | -- | The ada left over for the return collateral output after covering
    -- the required collateral is below the output's minimum UTxO value.
    ReturnCollateralBelowMinimumUTxO
      L.Coin
      -- ^ Ada left for the return collateral output
      L.Coin
      -- ^ Minimum UTxO value of the return collateral output
  deriving (Eq, Show)
```

with

```haskell
calcReturnAndTotalCollateral
  :: ... -> Either CollateralError (TxReturnCollateral CtxTx era, TxTotalCollateral era)
```

and on the experimental side `Either CollateralError (Maybe (TxReturnCollateral (LedgerEra era)), Maybe TxTotalCollateral)`.

What this buys:

- The type parameters disappear.
  Each API writes its own success pair in its own signature, and only the monomorphic error type (bare `Coin`s) is shared between the two modules.
  As defined here, `ret`/`tot` are completely unconstrained (`CollateralComputation Int String` typechecks), which is uncomfortable for a type exported from `Cardano.Api.Tx`.
- `TxBodyErrorAutoBalance` gains one constructor per API (`TxBodyErrorCollateral CollateralError`) instead of two, and the currently byte-for-byte duplicated constructor docs and `Error` rendering collapse into a single function on `CollateralError` that both instances delegate to.
- Every call site shrinks to `first TxBodyErrorCollateral $ calcReturnAndTotalCollateral ...`.
- Since the experimental `calcReturnAndTotalCollateral` is not exported, this is strictly less public API breakage than the current version.

On `NoCollateralNeeded`: it becomes `Right (TxReturnCollateralNone, TxTotalCollateralNone)` (`Right (Nothing, Nothing)` in the experimental copy), which is exactly what every caller already turns it into.
No ambiguity arises: the computed branch always produces an actual return output and total, and the pass-through branches only fire when the user set at least one field, so "both fields empty" is produced by exactly one path - the no-collateral-inputs clause.
A Haddock line on the function ("returns both fields as `None` when the transaction has no collateral inputs") preserves the documentation.
The old encoding was only broken because insufficient collateral *also* mapped to `(None, None)` - and that case now lives in `Left`, which was the actual bug.

This also composes with #1267 unchanged: the dust-folding case is just another `Right` with no return output.

carbolymer (changes requested):
Let's discuss `CollateralComputation` https://github.com/IntersectMBO/cardano-api/pull/1266/changes#r3646342535

cardano-api/test/cardano-api-test/Test/Cardano/Api/Experimental/Fee.hs
```diff
@@ -885,6 +892,103 @@ prop_estimateBalancedTxBody_no_collateral_without_plutus = H.propertyOnce $ do
...
+    epochInfo
+    ledgerPParams
+    mempty
+    mempty
+    mempty
+    utxo
+    txBodyContent
+    (Api.fromShelleyAddr sbe addr)
+    Nothing of
+    Left _ -> pure ()
```
carbolymer: Shouldn't `Left` make the test fail?

cardano-api/test/cardano-api-test/Test/Cardano/Api/Experimental/Fee.hs
```diff
@@ -885,6 +892,103 @@ prop_estimateBalancedTxBody_no_collateral_without_plutus = H.propertyOnce $ do
...
+    Right (_, balancedContent) ->
+      -- The ledger requires the ada in the return collateral output to cover
+      -- the minimum UTxO value of the output.
+      case Exp.txReturnCollateral balancedContent of
```
carbolymer: suggestion: use `forM_` instead

cardano-api/test/cardano-api-test/Test/Cardano/Api/Transaction/Autobalance.hs
```diff
@@ -506,43 +506,43 @@ prop_calcReturnAndTotalCollateral = H.withTests 400 . H.property $ do
...
+    | txRetColl /= TxReturnCollateralNone || txTotColl /= TxTotalCollateralNone ->
         -- got collateral values as function arguments - not calculating anything
-        txRetColl === resRetColl
-        txTotColl === resTotColl
-    | totalCollateralAda < requiredCollateralAda -> do
-        -- provided collateral not enough, not calculating anything
-        TxReturnCollateralNone === resRetColl
-        TxTotalCollateralNone === resTotColl
-    | otherwise -> do
+        CollateralComputed txRetColl txTotColl === result
```
carbolymer: so it's computed (vide constructor name) and not-computed (vide comment) at the same time?

cardano-api/src/Cardano/Api/Tx/Internal/Fee.hs
```diff
@@ -878,6 +907,22 @@ instance Error (TxBodyErrorAutoBalance era) where
         [ "ScriptWitnessIndex (redeemer pointer): " <> pshow sIndex <> " is missing from the execution "
         , "units (redeemer pointer) map: " <> pshow eUnitsMap
         ]
+    TxBodyErrorInsufficientCollateral available required ->
+      mconcat
+        [ "The ada in the collateral inputs does not cover the required collateral. \n"
+        , "Ada in collateral inputs: " <> pretty available <> "\n"
+        , "Required collateral: "
+            <> pretty required
+            <> " (the transaction fee multiplied by the collateralPercentage protocol parameter). \n"
```
carbolymer:
```suggestion
            <> " (the transaction fee multiplied by the collateralPercentage / 100 protocol parameter). \n"
```

cardano-api/src/Cardano/Api/Tx/Internal/Fee.hs
```diff
@@ -1279,31 +1360,34 @@ calcReturnAndTotalCollateral w fee pp' TxInsCollateral{} txReturnCollateral txTo
...
       | returnCollateralAmount < 0 ->
-          (TxReturnCollateralNone, TxTotalCollateralNone)
+          InsufficientCollateral
+            totalCollateralLovelace
+            (L.rationalToCoinViaCeiling $ reqAmt % 100)
       | otherwise ->
-          ( TxReturnCollateral
-              w
-              ( TxOut
+          let returnCollateralTxOut =
```
carbolymer, line 1359: suggestion: use `do` instead of `let-in`, like everywhere around

cardano-api/test/cardano-api-test/Test/Cardano/Api/Transaction/Autobalance.hs
```diff
@@ -688,6 +688,97 @@ prop_estimate_balanced_tx_body_no_collateral_without_plutus = H.propertyOnce $ d
...
+    pparams
+    mempty
+    mempty
+    mempty
+    utxos
+    content
+    address
+    Nothing of
+    Left _ -> pure ()
+    Right (BalancedTxBody balancedContent _ _ _) ->
```
carbolymer: same remarks from the other test apply here

cardano-api/test/cardano-api-test/Test/Cardano/Api/Experimental/Fee.hs
palas: No, because there is not enough collateral. Input collateral is minUTxO, and the transaction has a Plutus script, so the remaining collateral to return would be less than minUTxO, so you cannot have a valid colateral return.

You could just trash the remaining collateral (as in PR https://github.com/IntersectMBO/cardano-api/pull/1267), but you can't because in this case the collateral also has tokens, and you cannot trash the tokens.

So failure is what is expected. It is a bit confusing that `Right` doesn't simply fail. But the reason is I want to make sure that it actually generated a transaction with a less than minUTxO output, which is invalid

cardano-api/src/Cardano/Api/Tx/Internal/Fee.hs
palas: Yes, good idea. I'll use your proposed type structure

palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/1266/changes/9775a75592d82df7dbda77bd28b57ab2d5fb215a

cardano-api/test/cardano-api-test/Test/Cardano/Api/Experimental/Fee.hs
palas: Addressed here for the `Maybe`: https://github.com/IntersectMBO/cardano-api/pull/1266/changes/9775a75592d82df7dbda77bd28b57ab2d5fb215a

I am not doing it for the `Either` in principle since it is already confusing enough

cardano-api/test/cardano-api-test/Test/Cardano/Api/Transaction/Autobalance.hs
palas: Addressed as I side effect of: https://github.com/IntersectMBO/cardano-api/pull/1266/changes/9775a75592d82df7dbda77bd28b57ab2d5fb215a

cardano-api/test/cardano-api-test/Test/Cardano/Api/Transaction/Autobalance.hs
palas: The only thing left to do was maybe commenting on the `Left` branch, on why it passes. And that is addressed in: https://github.com/IntersectMBO/cardano-api/commit/9775a75592d82df7dbda77bd28b57ab2d5fb215a

cardano-api/src/Cardano/Api/Tx/Internal/Fee.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-api/commit/9775a75592d82df7dbda77bd28b57ab2d5fb215a

cardano-api/src/Cardano/Api/Tx/Internal/Fee.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-api/commit/9775a75592d82df7dbda77bd28b57ab2d5fb215a

.changes/20260723_cardano_api_collateral_min_utxo.yml
```diff
@@ -0,0 +1,8 @@
+project: cardano-api
+pr: 1266
+kind:
+  - bugfix
+  - breaking
+description: |
+  Issue: `makeTransactionBodyAutoBalance` and `estimateBalancedTxBody` (both the traditional and the experimental versions) could build transactions that the ledger rejects: with a return collateral output below the minimum UTxO value, or with insufficient collateral.
+  Solution: balancing now fails with the new `TxBodyErrorCollateral` error instead of building a rejected transaction. `calcReturnAndTotalCollateral` now returns the collateral fields in an `Either CollateralError` instead of a plain pair.
```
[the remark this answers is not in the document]
palas: Already addressed

cardano-api/test/cardano-api-test/Test/Cardano/Api/Transaction/Autobalance.hs
```diff
@@ -688,6 +690,100 @@ prop_estimate_balanced_tx_body_no_collateral_without_plutus = H.propertyOnce $ d
...
+    mempty
+    mempty
+    utxos
+    content
+    address
+    Nothing of
+    -- Any failure passes here: this is a regression test against
+    -- successfully building an invalid transaction, and it predates the fix,
+    -- so it cannot name the specific error the fix introduces.
+    Left _ -> pure ()
```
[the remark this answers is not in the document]
palas: Addressed in: https://github.com/IntersectMBO/cardano-api/pull/1266/changes/e85db1fb3ccf08b151b78dbdff9e68bbb927f7fb

cardano-api/test/cardano-api-test/Test/Cardano/Api/Experimental/Fee.hs
```diff
@@ -885,6 +893,104 @@ prop_estimateBalancedTxBody_no_collateral_without_plutus = H.propertyOnce $ do
...
+    mempty
+    mempty
+    utxo
+    txBodyContent
+    (Api.fromShelleyAddr sbe addr)
+    Nothing of
+    -- Any failure passes here: this is a regression test against
+    -- successfully building an invalid transaction, and it predates the fix,
+    -- so it cannot name the specific error the fix introduces.
+    Left _ -> pure ()
```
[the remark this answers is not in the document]
palas: Addressed in: https://github.com/IntersectMBO/cardano-api/pull/1266/changes/e85db1fb3ccf08b151b78dbdff9e68bbb927f7fb

cardano-api/src/Cardano/Api/Tx/Internal/Fee.hs
```diff
@@ -1239,6 +1248,51 @@ hasPlutusScriptWitnesses sbe txbodycontent =
...
+        , "Ada left for the return collateral output: " <> pretty returnAda <> "\n"
+        , "Minimum required UTxO: " <> pretty minUTxO <> "\n"
+        , "The usual solution is to provide collateral inputs with more ada."
+        ]
+
+-- | Compute the return and total collateral fields of the transaction.
+-- Both fields are returned as 'TxReturnCollateralNone' and
+-- 'TxTotalCollateralNone' if and only if the transaction uses no collateral
+-- inputs. Fields provided by the user are passed through unvalidated:
+-- 'Left' can only arise when the function computes the fields itself.
```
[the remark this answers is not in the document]
palas: Addressed in: https://github.com/IntersectMBO/cardano-api/pull/1266/changes/e85db1fb3ccf08b151b78dbdff9e68bbb927f7fb

cardano-api/src/Cardano/Api/Experimental/Tx/Internal/Fee.hs
```diff
@@ -696,6 +704,13 @@ hasPlutusScriptWitnesses txbodycontent =
 
 -- Calculation taken from validateInsufficientCollateral:
 -- https://github.com/input-output-hk/cardano-ledger/blob/389b266d6226dedf3d2aec7af640b3ca4984c5ea/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs#L335
+
+-- | Compute the return and total collateral fields of the transaction.
+-- Both fields are returned as 'Nothing' if and only if the transaction uses
+-- no collateral inputs. Fields provided by the user are passed through
+-- unvalidated: 'Left' can only arise when the function computes the fields
+-- itself.
```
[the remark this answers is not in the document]
palas: Addressed in: https://github.com/IntersectMBO/cardano-api/pull/1266/changes/e85db1fb3ccf08b151b78dbdff9e68bbb927f7fb

cardano-api/src/Cardano/Api/Tx/Internal/Fee.hs
```diff
@@ -1279,31 +1335,39 @@ calcReturnAndTotalCollateral w fee pp' TxInsCollateral{} txReturnCollateral txTo
...
       | returnCollateralAmount < 0 ->
-          (TxReturnCollateralNone, TxTotalCollateralNone)
-      | otherwise ->
-          ( TxReturnCollateral
-              w
-              ( TxOut
+          Left $
+            InsufficientCollateral
+              totalCollateralLovelace
+              (L.rationalToCoinViaCeiling $ reqAmt % 100)
```
carbolymer: This is calculated twice (in line 1331)

cardano-api/src/Cardano/Api/Tx/Internal/Fee.hs
```diff
@@ -1239,6 +1248,53 @@ hasPlutusScriptWitnesses sbe txbodycontent =
...
+    InsufficientCollateral
+      L.Coin
+      -- ^ Ada available in the collateral inputs
+      L.Coin
+      -- ^ Required collateral
+  | -- | The ada left over for the return collateral output after covering
+    -- the required collateral is below the output's minimum UTxO value.
+    ReturnCollateralBelowMinimumUTxO
+      L.Coin
+      -- ^ Ada left for the return collateral output
```
carbolymer: This constructor now has two producers reporting values with different provenance: the computed arm passes the ada left over after subtracting the required collateral (line 1369 here, 773 in the experimental copy), but the user-provided arm added in the last commit passes the ada in the caller's own output (line 1350 / 763), where no subtraction is involved.
The Haddock describes only the first, so for a provided output it can mislead: with 100 ada of collateral inputs, a required collateral of 1 ada and an explicitly provided 0.5 ada return collateral output, the error reads as "the ada left over after covering the required collateral is 0.5" - even though ~99 ada was left over and the actual problem is the output being built too small.
`prettyError` already covers both cases ("provide collateral inputs with more ada, or increase the ada in the explicitly provided return collateral output"); the Haddock should match:

```suggestion
  | -- | The ada in the return collateral output is below the output's
    -- minimum UTxO value. The output is either computed from the collateral
    -- inputs (the ada left over after covering the required collateral) or
    -- provided explicitly by the user.
    ReturnCollateralBelowMinimumUTxO
      L.Coin
      -- ^ Ada in the return collateral output
```

cardano-api/test/cardano-api-test/Test/Cardano/Api/Experimental/Fee.hs
```diff
@@ -885,6 +893,107 @@ prop_estimateBalancedTxBody_no_collateral_without_plutus = H.propertyOnce $ do
...
+      H.annotateShow err
+      H.annotate "Expected balancing to fail with a collateral error"
+      H.failure
+    Right (_, balancedContent) ->
+      -- The ledger requires the ada in the return collateral output to cover
+      -- the minimum UTxO value of the output.
+      forM_ (Exp.txReturnCollateral balancedContent) $ \(Exp.TxReturnCollateral returnCollateralTxOut) -> do
+        let minUTxO = Exp.calculateMinimumUTxO ledgerPParams (Exp.TxOut returnCollateralTxOut)
+        H.note_ "Check that the return collateral output meets the minimum UTxO value"
+        H.assertWith (returnCollateralTxOut ^. L.coinTxOutL, minUTxO) $ uncurry (>=)
```
carbolymer: With the fix in place this test is deterministic - `propertyOnce`, pinned fixture, pinned `conway.json` - and only this `Left` branch is reachable.
The `Right` branch went dead the moment the fix commit landed; its purpose was to make the test commit fail on its parent commit, which the commit ordering already demonstrates.

Keeping the two-outcome structure weakens the test in both directions:

- `Exp.TxBodyErrorCollateral _` also accepts `InsufficientCollateral`, which would be a wrong outcome for this fixture (the collateral covers the requirement comfortably; only `ReturnCollateralBelowMinimumUTxO` is correct here).
- The `Right` branch is a weaker net than it looks: a regression that succeeds while dropping the token from the return collateral output (ledger-invalid - non-ada collateral must be fully returned), or while omitting the return output entirely (`forM_` over `Nothing` is `pure ()`), passes this test.

Since exactly one outcome is correct, suggest asserting it and failing on everything else:

```haskell
    -- The leftover collateral ada cannot cover the token-carrying return
    -- collateral output's minimum UTxO value, so balancing must fail with
    -- exactly this error.
    Left (Exp.TxBodyErrorCollateral (ReturnCollateralBelowMinimumUTxO _ _)) -> H.success
    Left err -> do
      H.annotateShow err
      H.annotate "Expected balancing to fail with ReturnCollateralBelowMinimumUTxO"
      H.failure
    Right _ -> do
      H.annotate "Expected balancing to fail with ReturnCollateralBelowMinimumUTxO, but it succeeded"
      H.failure
```

(`CollateralError (..)` is exported from `Cardano.Api.Tx`.)
Any unexpected success then fails immediately instead of being waved through if it happens to clear one min-UTxO check.
The same applies to the traditional twin at `Test/Cardano/Api/Transaction/Autobalance.hs:786`.
If the concern is protocol-parameter drift: the parameters are a pinned fixture file in this repo, so updating the assertion alongside the fixture is the normal cost, not a reason to keep the assertion loose.

cardano-api/test/cardano-api-test/Test/Cardano/Api/Transaction/Autobalance.hs
```diff
@@ -506,43 +506,56 @@ prop_calcReturnAndTotalCollateral = H.withTests 400 . H.property $ do
...
+            H.annotateShow collBalance
+            H.note_ "Check if collateral balance is positive"
+            H.assertWith collBalance $ L.pointwise (<=) mempty
+            H.note_ "Check if collateral balance contains only ada"
+            H.assertWith collBalance L.isAdaOnly
+            H.note_ "Check if collateral balance is at least minimum required"
+            H.assertWith collBalance $ L.pointwise (<=) (L.inject requiredCollateralAda)
+            H.note_ "Check that collateral balance is equal to collateral in tx body"
+            resTotCollValue === collBalance
+          _ -> H.failure
```
carbolymer: This wildcard catches exactly one constructor: `Left (InsufficientCollateral _ _)`.
It should be unreachable - the `totalCollateralAda < requiredCollateralAda` guard above mirrors the implementation's `returnCollateralAmount < 0` check (equivalent at every boundary, since `T < ceiling(R/100)` ⟺ `T * 100 < R` for integers).
Naming it makes the failure self-explanatory and keeps the case exhaustive, so a future `CollateralError` constructor trips `-Wincomplete-patterns` here instead of being silently absorbed into an unexplained failure:

```suggestion
          Left err@InsufficientCollateral{} -> do
            H.annotateShow err
            H.annotate "Unreachable: the guard above ensures the collateral covers the requirement"
            H.failure
```

cardano-api/test/cardano-api-test/Test/Cardano/Api/Transaction/Autobalance.hs
```diff
@@ -688,6 +701,103 @@ prop_estimate_balanced_tx_body_no_collateral_without_plutus = H.propertyOnce $ d
...
+      H.failure
+    Right (BalancedTxBody balancedContent _ _ _) ->
+      -- The ledger requires the ada in the return collateral output to cover
+      -- the minimum UTxO value of the output.
+      case txReturnCollateral balancedContent of
+        TxReturnCollateralNone -> pure ()
+        TxReturnCollateral _ txOut@(TxOut _ txOutValue _ _) -> do
+          let minUTxO = calculateMinimumUTxO sbe ledgerPParams txOut
+          H.note_ "Check that the return collateral output meets the minimum UTxO value"
+          H.assertWith (txOutValueToLovelace txOutValue, minUTxO) $ uncurry (>=)
```
carbolymer: Same issue as the experimental twin (https://github.com/IntersectMBO/cardano-api/pull/1266#discussion_r3657909624): with the fix in place this `propertyOnce` test is deterministic, so only the `Left (TxBodyErrorCollateral _)` branch is reachable and the `Right` branch is dead.
Here the weak success net is even explicit: `TxReturnCollateralNone -> pure ()` waves through a balancing result that omits the return output entirely, which the ledger would reject while tokens sit in the collateral inputs.
Suggest asserting the single correct outcome:

```suggestion
    -- The leftover collateral ada cannot cover the token-carrying return
    -- collateral output's minimum UTxO value, so balancing must fail with
    -- exactly this error.
    Left (Api.TxBodyErrorCollateral (ReturnCollateralBelowMinimumUTxO _ _)) -> H.success
    Left err -> do
      H.annotateShow err
      H.annotate "Expected balancing to fail with ReturnCollateralBelowMinimumUTxO"
      H.failure
    Right _ -> do
      H.annotate "Expected balancing to fail with ReturnCollateralBelowMinimumUTxO, but it succeeded"
      H.failure
```

carbolymer (approved):

cardano-api/test/cardano-api-test/Test/Cardano/Api/Experimental/Fee.hs
```diff
@@ -978,16 +979,17 @@ prop_makeTransactionBodyAutoBalance_return_collateral_with_tokens_below_min_utxo
...
-      case Exp.txReturnCollateral balancedContent of
-        Nothing -> pure ()
-        Just (Exp.TxReturnCollateral returnCollateralTxOut) -> do
-          let minUTxO = Exp.calculateMinimumUTxO ledgerPParams (Exp.TxOut returnCollateralTxOut)
-          H.note_ "Check that the return collateral output meets the minimum UTxO value"
-          H.assertWith (returnCollateralTxOut ^. L.coinTxOutL, minUTxO) $ uncurry (>=)
+      forM_ (Exp.txReturnCollateral balancedContent) $ \(Exp.TxReturnCollateral returnCollateralTxOut) -> do
+        let minUTxO = Exp.calculateMinimumUTxO ledgerPParams (Exp.TxOut returnCollateralTxOut)
+        H.note_ "Check that the return collateral output meets the minimum UTxO value"
+        H.assertWith (returnCollateralTxOut ^. L.coinTxOutL, minUTxO) $ uncurry (>=)
```
Jimbo4350: This is a bizarre way to represent the assertion. `uncurry` makes things less readable here.

Jimbo4350 (approved):

cardano-api/test/cardano-api-test/Test/Cardano/Api/Experimental/Fee.hs
palas: Yes, I was thinking that it would be good to ensure that the reason the test fails is valid, that is: the collateral is actually below minUTxO. But you are right that it is a weakening of the test, and that is not desirable, because it is the wrong type of failure, a false negative. So I'll remove it

cardano-api/test/cardano-api-test/Test/Cardano/Api/Experimental/Fee.hs
palas: This got removed altogether when addressing the previous reviews

cardano-api/src/Cardano/Api/Tx/Internal/Fee.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/1266/changes/229fc8f2a9cc988cd0b998ac61b80e397b006da5

cardano-api/src/Cardano/Api/Tx/Internal/Fee.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/1266/changes/229fc8f2a9cc988cd0b998ac61b80e397b006da5

cardano-api/test/cardano-api-test/Test/Cardano/Api/Transaction/Autobalance.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/1266/changes/229fc8f2a9cc988cd0b998ac61b80e397b006da5

cardano-api/test/cardano-api-test/Test/Cardano/Api/Transaction/Autobalance.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/1266/changes/229fc8f2a9cc988cd0b998ac61b80e397b006da5

cardano-api/test/cardano-api-test/Test/Cardano/Api/Experimental/Fee.hs
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/1266/changes/229fc8f2a9cc988cd0b998ac61b80e397b006da5

PR #1267, opened 2026-07-24
Fold return collateral dust into the total collateral

### LINKED ISSUE #1261: [BUG] - `transaction build` does not validate min UTxO value of the collateral return output

palas:
# Context

This is the third of three stacked PRs fixing #1261: the balancing functions
(`makeTransactionBodyAutoBalance` and `estimateBalancedTxBody`, in both the traditional and the
experimental API) could build transactions that the ledger rejects at submission time because of
their collateral. Each PR contains a pair of commits — regression tests that fail, then the fix
that makes them pass — and each PR is based on the previous one, so every diff shows only its own
two commits.

🗺️ **Map** — you are here: 3️⃣

|  |  |  |
|---|---|---|
| 🚪 | [#1261](https://github.com/IntersectMBO/cardano-api/issues/1261) | the issue: `transaction build` does not validate min UTxO value of the collateral return output |
| 1️⃣ | #1265 | remove collateral from transactions without Plutus scripts |
| 2️⃣ | #1266 | fail on invalid computed collateral instead of building a rejected transaction |
| 3️⃣ | **(this PR)** | fold return collateral dust into the total collateral |

**This PR:** after the previous PR, balancing fails when the ada left over for the return
collateral output is below its minimum UTxO value — even when the collateral is ada-only and a
valid transaction exists. `calcReturnAndTotalCollateral` (both copies) now folds that ada into
the total collateral and omits the return collateral output instead of failing; the extra ada is
only lost if a Plutus script fails on chain. `estimateBalancedTxBody` cannot see the content of
the collateral inputs, so it computes collateral under the assumption that they contain no native
tokens — now documented on its `totalPotentialCollateral` parameters. A transaction built under a
wrong assumption is rejected by the ledger, and tokens are never lost: a valid transaction must
always give the tokens back in the return collateral output. Fixes #1261.

# How to trust this PR

- The first commit adds four regression tests (one per balancing function) with an ada-only
  collateral input holding exactly its minimum UTxO value: balancing must succeed, with no return
  collateral output and the whole input as total collateral. On the base branch they fail with
  the `ReturnCollateralBelowMinimumUTxO` collateral error; the second commit makes them pass.
- The fold only happens when the collateral value is ada-only (`isZero` on the non-ada part):
  token-carrying collateral keeps failing, and the tests from the previous PR still pass.
- The pre-existing property test `prop_calcReturnAndTotalCollateral` needs no change: its
  computed-case assertions hold for folded results.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-api/src/Cardano/Api/Experimental/Tx/Internal/Fee.hs
+++ b/cardano-api/src/Cardano/Api/Experimental/Tx/Internal/Fee.hs
@@ -785,8 +785,8 @@ calcReturnAndTotalCollateral fee pp' _ mTxReturnCollateral mTxTotalCollateral cA
                   )
             | L.isZero nonAdaCollateral ->
                 -- The ada left over is too small for a return collateral output, so
-                -- use the whole collateral as total collateral instead. The extra ada
-                -- is only lost if a Plutus script fails on chain.
+                -- use all of the collateral inputs as total collateral instead. The
+                -- extra ada is only lost if a Plutus script fails on chain.
                 Right (Nothing, Just $ TxTotalCollateral totalCollateralLovelace)
             | otherwise ->
                 Left $ ReturnCollateralBelowMinimumUTxO returnCollateralAda minReturnUTxO
--- a/cardano-api/src/Cardano/Api/Experimental/Tx/Internal/Fee.hs
+++ b/cardano-api/src/Cardano/Api/Experimental/Tx/Internal/Fee.hs
@@ -232,7 +232,11 @@ estimateBalancedTxBody
   -> Map (Ledger.PlutusPurpose Ledger.AsIx (LedgerEra era)) ExecutionUnits
   -- ^ Plutus script execution units.
   -> Coin
-  -- ^ Total potential collateral amount.
+  -- ^ Total potential collateral amount. The content of the collateral
+  --   inputs is not visible to this function, so they are assumed to
+  --   contain no native tokens. To use collateral inputs that carry
+  --   native tokens, provide 'txReturnCollateral' and 'txTotalCollateral'
+  --   explicitly.
   -> Int
   -- ^ The number of key witnesses to be added to the transaction.
   -> Int
@@ -292,7 +296,11 @@ estimateBalancedTxBody'
   -> Map ScriptWitnessIndex ExecutionUnits
   -- ^ Plutus script execution units.
   -> Coin
-  -- ^ Total potential collateral amount.
+  -- ^ Total potential collateral amount. The content of the collateral
+  --   inputs is not visible to this function, so they are assumed to
+  --   contain no native tokens. To use collateral inputs that carry
+  --   native tokens, provide 'txReturnCollateral' and 'txTotalCollateral'
+  --   explicitly.
   -> Int
   -- ^ The number of key witnesses to be added to the transaction.
   -> Int
@@ -769,13 +777,19 @@ calcReturnAndTotalCollateral fee pp' _ mTxReturnCollateral mTxTotalCollateral cA
       | otherwise -> obtainCommonConstraints (useEra @era) $ do
           let returnCollateralTxOut = L.mkBasicTxOut (toShelleyAddr cAddr) returnCollateral
               minReturnUTxO = calculateMinimumUTxO pp' (TxOut returnCollateralTxOut)
-          if returnCollateralAda < minReturnUTxO
-            then Left $ ReturnCollateralBelowMinimumUTxO returnCollateralAda minReturnUTxO
-            else
-              Right
-                ( Just $ TxReturnCollateral returnCollateralTxOut
-                , Just $ TxTotalCollateral totalCollateral
-                )
+          if
+            | returnCollateralAda >= minReturnUTxO ->
+                Right
+                  ( Just $ TxReturnCollateral returnCollateralTxOut
+                  , Just $ TxTotalCollateral totalCollateral
+                  )
+            | L.isZero nonAdaCollateral ->
+                -- The ada left over is too small for a return collateral output, so
+                -- use the whole collateral as total collateral instead. The extra ada
+                -- is only lost if a Plutus script fails on chain.
+                Right (Nothing, Just $ TxTotalCollateral totalCollateralLovelace)
+            | otherwise ->
+                Left $ ReturnCollateralBelowMinimumUTxO returnCollateralAda minReturnUTxO
 
 -- | Transaction fees can be computed for a proposed transaction based on the
 -- expected number of key witnesses (i.e. signatures).
--- a/cardano-api/src/Cardano/Api/Tx/Internal/Fee.hs
+++ b/cardano-api/src/Cardano/Api/Tx/Internal/Fee.hs
@@ -1373,8 +1373,8 @@ calcReturnAndTotalCollateral w fee pp' TxInsCollateral{} txReturnCollateral txTo
                 Right (TxReturnCollateral w returnCollateralTxOut, totalCollateral)
             | L.isZero nonAdaCollateral ->
                 -- The ada left over is too small for a return collateral output, so
-                -- use the whole collateral as total collateral instead. The extra ada
-                -- is only lost if a Plutus script fails on chain.
+                -- use all of the collateral inputs as total collateral instead. The
+                -- extra ada is only lost if a Plutus script fails on chain.
                 Right (TxReturnCollateralNone, TxTotalCollateral w totalCollateralLovelace)
             | otherwise ->
                 Left $ ReturnCollateralBelowMinimumUTxO returnCollateralAda minReturnUTxO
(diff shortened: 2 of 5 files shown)


cardano-api/test/cardano-api-test/Test/Cardano/Api/Experimental/Fee.hs
```diff
@@ -994,6 +1002,182 @@ prop_makeTransactionBodyAutoBalance_return_collateral_with_tokens_below_min_utxo
...
+        0
+        (Api.fromShelleyAddr sbe addr)
+        (L.MaryValue (L.Coin 12_000_000) mempty)
+
+  case (Exp.txReturnCollateral balancedContent, Exp.txTotalCollateral balancedContent) of
+    (Nothing, Just (Exp.TxTotalCollateral totalCollateral)) ->
+      totalCollateral H.=== minUTxOCollateral
+    _ -> do
+      H.annotate "Expected the whole collateral as total collateral, with no return collateral output"
+      H.failure
```
carbolymer: Could using `H.assertWith` be helpful in printing more debuggable test output here?

cardano-api/src/Cardano/Api/Tx/Internal/Fee.hs
```diff
@@ -1365,9 +1370,16 @@ calcReturnAndTotalCollateral w fee pp' TxInsCollateral{} txReturnCollateral txTo
...
+          if
+            | returnCollateralAda >= minReturnUTxO ->
+                Right (TxReturnCollateral w returnCollateralTxOut, totalCollateral)
+            | L.isZero nonAdaCollateral ->
+                -- The ada left over is too small for a return collateral output, so
+                -- use the whole collateral as total collateral instead. The extra ada
+                -- is only lost if a Plutus script fails on chain.
+                Right (TxReturnCollateralNone, TxTotalCollateral w totalCollateralLovelace)
+            | otherwise ->
+                Left $ ReturnCollateralBelowMinimumUTxO returnCollateralAda minReturnUTxO
```
carbolymer, line 1380: The fold's behavioural boundary is currently unpinned by the property test, in both directions.
`prop_calcReturnAndTotalCollateral`'s `Right` branch checks balance positivity, ada-onliness, sufficiency and the declared-total equality, but:

- a regression producing a return collateral output below its own minimum UTxO still passes the property (only the deterministic regression tests catch the one pinned scenario), and
- an over-eager fold - folding ada-only collateral even when the leftover could fund a valid return output, silently putting the user's whole collateral at risk - also passes every current assertion.

Suggest strengthening the `Right` branch with two assertions: a produced return output must meet its own `calculateMinimumUTxO`, and no-return-output results are only valid when the reconstructed candidate output could not have met it.
That pins this `>= minReturnUTxO` boundary at property strength (400 runs, shrinking straight to the boundary) without adding end-to-end tests.
Related and pre-existing: the experimental `calcReturnAndTotalCollateral` still has no property test at all.

carbolymer (approved):

cardano-api/src/Cardano/Api/Experimental/Tx/Internal/Fee.hs
```diff
@@ -769,13 +777,19 @@ calcReturnAndTotalCollateral fee pp' _ mTxReturnCollateral mTxTotalCollateral cA
...
-                )
+          if
+            | returnCollateralAda >= minReturnUTxO ->
+                Right
+                  ( Just $ TxReturnCollateral returnCollateralTxOut
+                  , Just $ TxTotalCollateral totalCollateral
+                  )
+            | L.isZero nonAdaCollateral ->
+                -- The ada left over is too small for a return collateral output, so
+                -- use the whole collateral as total collateral instead. The extra ada
```
Jimbo4350: "total collateral" should be renamed to "all of the collateral inputs".

cardano-api/src/Cardano/Api/Experimental/Tx/Internal/Fee.hs
```diff
@@ -769,13 +777,19 @@ calcReturnAndTotalCollateral fee pp' _ mTxReturnCollateral mTxTotalCollateral cA
...
-                ( Just $ TxReturnCollateral returnCollateralTxOut
-                , Just $ TxTotalCollateral totalCollateral
-                )
+          if
+            | returnCollateralAda >= minReturnUTxO ->
+                Right
+                  ( Just $ TxReturnCollateral returnCollateralTxOut
+                  , Just $ TxTotalCollateral totalCollateral
+                  )
+            | L.isZero nonAdaCollateral ->
```
Jimbo4350, line 786: Are you sure we want this? For ada-only collateral we silently fold a sub-minUTxO surplus into total collateral, so the user forfeits that ada if a Plutus script fails on chain instead of failing with an actionable error telling them to add a collateral input.

palas: It was one of the difficult choices, here are the reasons why I thought it was a good idea:
- Collateral should never really be lost to begin with, it gets validated when constructing the transaction and transactions are deterministic, as far as I understand. (Turns out this is not 100% true, it seems it could happen in the middle of protocol parameters change.)
- Even if the collateral was lost, for some reason, the amount we are adding to the collateral is less than minUTxO, so it shouldn't be a lot. (Turns out this is not negligible either, apparently this can be more than the collateral even if it is currently less than $1.)
- The purpose of the function can be thought of as minimising the amount of collateral if it is possible to make a transaction with the provided input collateral. And trashing the dust satisfies this ask: it is possible to make a transaction, we provide it, and there is no transaction possible with those inputs and less collateral.

On the other hand, an error is not a bad response either, although it may annoy users more.

Maybe the most pragmatic thing would be to give a warning.

If not we can just close the PR

cardano-api/test/cardano-api-test/Test/Cardano/Api/Experimental/Fee.hs
palas: It doesn't work because there are `Show` instances missing:
```
test/cardano-api-test/Test/Cardano/Api/Experimental/Fee.hs:1086:3: error: [GHC-39999]
    • Could not deduce ‘Show Exp.TxTotalCollateral’
        arising from a use of ‘H.assertWith’
```

palas: Instead I've removed the case and used `H.===` for both assertions in: https://github.com/IntersectMBO/cardano-api/pull/1267/changes/c44b25f0a730a4b0ede3eb7f4bb9d929a3354a49.
So we don't have the `H.failure` without info.

cardano-api/src/Cardano/Api/Tx/Internal/Fee.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-api/pull/1267/changes/c44b25f0a730a4b0ede3eb7f4bb9d929a3354a49
Also tweaked the generator to exercise them

cardano-api/src/Cardano/Api/Experimental/Tx/Internal/Fee.hs
palas: I am guessing you meant "whole collateral". Addressed in: https://github.com/IntersectMBO/cardano-api/pull/1267/changes/c44b25f0a730a4b0ede3eb7f4bb9d929a3354a49

Jimbo4350 (approved):

Issue #1229, opened 2026-05-27
gRPC: Replace Ogmios as Kupo's chain data source with cardano-rpc gRPC


carbolymer:
## Goal

Enable [Kupo](https://github.com/IntersectMBO/kupo) to use cardano-rpc as a drop-in replacement for its [Ogmios](https://github.com/IntersectMBO/ogmios) backend.
Kupo currently uses Ogmios for chain synchronisation via WebSocket/JSON-RPC.
The same capability can be provided natively by cardano-rpc over gRPC, eliminating the need for a separate Ogmios process.

## Kupo's Ogmios usage

Kupo depends on exactly three Ogmios methods:

| Ogmios method | Mini-protocol | Purpose |
|---|---|---|
| `findIntersection` | ChainSync | Resume sync from a known chain point |
| `nextBlock` | ChainSync | Stream blocks (RollForward/RollBackward) with full block data |
| `queryNetwork/genesisConfiguration` | Network query | One-shot Byron genesis bootstrap (optional if local genesis files available) |

Additionally, Kupo's HTTP API uses a separate Ogmios connection for ad-hoc historical block lookups (e.g. the `/metadata` endpoint fetches a block by slot on demand).

## Required gRPC services

1. **`SyncService.FollowTip`** - #1219
   Server-streaming RPC delivering blocks as applied/rolled back.
   Must carry full block data (transactions, inputs, outputs, datums, scripts, redeemers, metadata).
   This is the primary blocker - replaces Ogmios's `findIntersection` + `nextBlock` streaming.

2. **`SyncService.FetchBlock`** - #1216
   Retrieve a full block by chain point.
   Required for Kupo's HTTP API endpoints that perform ad-hoc historical block lookups (e.g. `/metadata`).

3. **`QueryService.ReadGenesis`** - #1217
   Return genesis configuration for all eras.
   Only required when Kupo cannot read genesis files from disk.

## Prerequisites

- ADR-019 (node kernel access) - direct ChainDB access is needed for efficient block streaming and block-by-point retrieval.
  Current N2C IPC (one connection per request, double serialisation) cannot support this.

## Not required for this goal

- `ReadTip` (#1218) - same SyncService cluster, likely lands alongside FollowTip, but Kupo does not use it (tip comes back in FollowTip responses)
- `SubmitTx`, `EvalTx` - already implemented, Kupo does not use them
- Mempool watching, stake/governance queries - Kupo does not use them

## Downstream integration

[...]

PR #1268, opened 2026-07-27
cardano-rpc: Implement FollowTip SyncService method

### LINKED ISSUE #1219: gRPC: Implement FollowTip (UTxO RPC SyncService) in cardano-rpc
## Summary

Implement the `FollowTip` RPC method from the UTxO RPC `SyncService` spec, providing a server-streaming gRPC endpoint that delivers blocks as they are applied or rolled back on the chain.
This is the gRPC equivalent of the Ouroboros ChainSync mini-protocol and the most architecturally significant method in the UTxO RPC spec.

### Replacing ChainSync for remote clients

Chain-following clients (Kupo, Scrolls, Oura, custom indexers) currently connect to the cardano-node Unix socket and run the ChainSync mini-protocol directly.
This requires:
- Local filesystem access to the node socket
- A Haskell (or compatible) implementation of the Ouroboros N2C handshake and ChainSync state machine
- Version negotiation with the node

`FollowTip` exposes the same block stream over gRPC, which:
- Works over TCP (no filesystem coupling)
- Uses standard gRPC client libraries available in every language
- Handles versioning at the proto/gRPC level rather than the Ouroboros protocol level

### Kupo's primary data path

Kupo's entire indexing pipeline runs on a single ChainSync connection.
It uses `MsgFindIntersect` to establish a starting point, then streams blocks via `MsgRequestNext` with adaptive pipelining.
`FollowTip` maps directly onto this: `intersect` for the starting point, `apply` for roll-forward, `undo` for roll-backward.

### What the spec requires

`FollowTipRequest` takes:
- `repeated BlockRef intersect` - candidate block references for finding a common point (analogous to Ouroboros `MsgFindIntersect`)
- Optional `FieldMask` - to select which parsed fields to materialise in the response

`FollowTipResponse` is a server-streaming response where each message carries:
- `oneof action`:
  - `apply` (`AnyChainBlock`) - a new block to apply (= Ouroboros RollForward)
  - `undo` (`AnyChainBlock`) - a block to undo (= Ouroboros RollBackward). Notably, the full block data is included, so the client does not need to maintain its own block cache for rollback processing
  - `reset` (`BlockRef`) - server-initiated reset to a point. No direct Ouroboros analogue; used when no intersect matched or on reinitialisation
- `tip` (`BlockRef`) - current chain tip for lag measurement

### Key differences from Ouroboros ChainSync

| Aspect | ChainSync | FollowTip |
|--------|-----------|-----------|
| Transport | Unix socket (N2C) | TCP/gRPC |
| Rollback data | Only the rollback point | Full block included in `undo` |
| No-intersect | `MsgIntersectNotFound` | `reset` action |
| Pipelining | Client-driven (`MsgRequestNextPipelined`) | Server-driven (gRPC stream backpressure) |
| Block format | CBOR | Proto (`AnyChainBlock` with `native_bytes` and/or parsed `cardano.Block`) |

### Why this depends on node kernel access

The existing N2C path is insufficient for a production-quality `FollowTip`:

- **N2C ChainSync is per-connection**: each gRPC client would require a separate N2C socket connection to the node, each running its own ChainSync follower. This does not scale.
- **No shared follower**: the node's ChainSync server creates a fresh follower per N2C connection. There is no way to multiplex multiple gRPC clients onto a single follower.
- **In-process access**: ADR-019's `NodeAccess` record can expose a `ChainDB.newFollower` callback, letting cardano-rpc create followers directly against the consensus `ChainDB`. Multiple gRPC clients can each get their own follower without socket overhead.

The in-process path also enables zero-copy block delivery: `ChainDB.getBlockComponent GetRawBlock` returns the raw CBOR directly from the ImmutableDB/VolatileDB without serialisation.

### NodeAccess extension

Extend the `NodeAccess` record (ADR-019) with:
- A follower creation callback (wrapping `ChainDB.newFollower`)
- Follower operations: read next (with rollback support), find intersection

The exact API depends on how `ChainDB.Follower` is surfaced through the consensus interface.

### Follower lifecycle

Each `FollowTip` gRPC stream creates a new `ChainDB` follower.
On stream setup:
1. Create a follower via `NodeAccess`.
2. Find intersection using the client's `repeated BlockRef intersect`.
3. If no intersection found, send a `reset` action to genesis (or the oldest available point).

On each iteration:
1. Read the next change from the follower (blocking if at tip).
2. Translate to `apply`, `undo`, or `reset` action.
3. Send the `FollowTipResponse` on the gRPC stream.

On stream teardown (client disconnect or error):
1. Close the follower to free resources.

### Block encoding

For the initial implementation, populate `native_bytes` (raw block CBOR) only.
Populating the parsed `cardano.Block` oneof requires mapping the full block structure to proto, which is a large effort and can follow separately.
Clients that need parsed data can deserialise the CBOR themselves.

### Backpressure

gRPC server-streaming has built-in flow control via HTTP/2 window management.
The follower read loop should respect this: if the client is slow to consume, the server blocks on the gRPC send rather than buffering unboundedly.

### Server wiring

Register `FollowTip` in the `SyncService` alongside `FetchBlock` and `ReadTip`.

### Scalability considerations

Each follower maintains a pointer into the `ChainDB`.
The node already supports multiple followers (one per N2C ChainSync connection), so this is not a new resource type.
However, gRPC makes it easier to open many connections from remote clients.
Consider:
- A configurable maximum number of concurrent followers
- Logging/metrics for active follower count
- Idle timeout for followers that stop consuming

### Parsed block support (future)

The `FollowTipResponse` can carry a parsed `cardano.Block` in the `AnyChainBlock`.
This requires mapping the full Cardano block structure (header, body, witnesses, auxiliary data) to the proto model.
This is significant additional work and should be a follow-up issue, not a blocker for the initial `FollowTip` implementation.

### FieldMask

`FollowTipRequest` includes a `FieldMask`.
As with other methods, this can be ignored initially (always return `native_bytes`).
Once parsed block support exists, `FieldMask` becomes valuable for letting clients request only headers, or only transaction hashes, reducing bandwidth.

## Acceptance criteria

- [ ] `NodeAccess` (ADR-019) exposes follower creation and intersection-finding.
- [ ] `followTipMethod` handler implemented as a gRPC server-streaming endpoint.
- [ ] Intersection logic: client sends `BlockRef` list, server finds most recent common point.
- [ ] Stream delivers `apply` (with raw block CBOR), `undo` (with raw block CBOR), and `reset` actions.
- [ ] Each response includes the current `tip`.
- [ ] Follower is cleaned up on client disconnect.
- [ ] Integration test in cardano-testnet: connect via `FollowTip`, submit a transaction, verify the block containing it arrives as an `apply` action with valid CBOR.
- [ ] `cardano-rpc/README.md` updated to mark `FollowTip` as supported in the coverage table.

## Out of scope

- Parsed `cardano.Block` population (native_bytes only for initial implementation).
- `FieldMask` support.
- Concurrent follower limits (can be added as a follow-up).
- `DumpHistory` (separate issue).

## Dependencies

- [ADR-019: Node Kernel Access for cardano-rpc](https://github.com/input-output-hk/cardano-node-wiki/pull/94) and its implementation.
- #1216 (FetchBlock) for the sync proto file and `NodeAccess` block-lookup foundation.

## References

- UTxO RPC SyncService spec: https://utxorpc.org/sync/spec/
- Upstream proto: `utxorpc/spec` repo, `proto/utxorpc/v1beta/sync/sync.proto`
- Ouroboros ChainSync mini-protocol: network-protocols package
- Related: #1216 (FetchBlock), #1217 (ReadGenesis)

carbolymer:
# Context

Implements the `SyncService.FollowTip` method (#1219), the primary blocker of the Kupo/Ogmios replacement effort (#1229). It is the gRPC analogue of the Ouroboros ChainSync mini-protocol, built on in-process ChainDB followers (ADR-019).

Stacked on #1259; review the commits after #1259's.

`FollowTip` streams fully parsed blocks as the chain advances, reusing the same block-to-proto conversion `FetchBlock` uses. The client's intersect list is honoured in preference order, and the first streamed message is always a reset announcing the resolved start point. An empty-hash block ref denotes origin, handled as an infallible catch-all rather than a special case. An empty intersect list follows from the current tip, matching Dolos's behaviour; an intersect list that matches nothing on the chain fails with `NOT_FOUND` before any message is sent. Every response carries the current tip alongside the action, for lag measurement.

Rollbacks are delivered as `undo` messages carrying the rolled-back blocks, newest first. Consensus's `RollBack` is point-only, it never carries block data, so each stream tracks the points of its last *k* applied blocks (the security parameter, about 86 KB per stream on mainnet) and re-fetches the undone blocks from the ChainDB on rollback. The stream falls back to a `reset` at the rollback point in the two cases where `undo` is impossible: the rollback target is outside the tracked window, or a re-fetch misses because the block was garbage-collected first.

## Implementation notes

- `NodeKernelAccess` gains `withFollower`, a bracket over `ChainDB.newFollower` that closes the follower and its resource registry on every exit path, including client disconnect, plus the `ChainFollower`/`ChainChange` types wrapping the raw follower API.
- `Cardano.Api.Consensus` gains the follower re-exports (`newFollower`, `Follower`, `ChainType`, `ChainUpdate`, `withRegistry`, `ResourceRegistry`).
- `FollowTip` is the first server-streaming handler in cardano-rpc, wired via grapesy's `mkServerStreaming`.
- A `TraceRpcFollowTipSpan` trace wraps the handler; the node-side rendering and the `rpc.request.SyncService.FollowTip` prometheus counter land in IntersectMBO/cardano-node#6579, the fetchblock-series node PR that already carries the earlier tracing and E2E work (there is no dedicated FollowTip node PR).
- The dead `TraceRpcForkerError` constructor is removed.

# How to trust this PR

The unit suite (83 tests) covers the conversion layer.

The E2E test in IntersectMBO/cardano-node#6579 exercises the streaming semantics live: full sync from origin, resuming from a point, an empty intersect list resolving from the current tip, and the `NOT_FOUND` path for an unmatched intersect list. It also adds a live-tail scenario, being added right now in the node working tree, that submits a transaction over gRPC and asserts it arrives via `FollowTip` as a parsed apply action.

Run it with:

```
cabal test cardano-rpc-test
TASTY_PATTERN='/RPC FollowTip/' cabal test cardano-testnet-test
```

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-rpc/src/Cardano/Rpc/Server/NodeKernelAccess.hs
+++ b/cardano-rpc/src/Cardano/Rpc/Server/NodeKernelAccess.hs
@@ -2,12 +2,17 @@
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE NoFieldSelectors #-}
 
 module Cardano.Rpc.Server.NodeKernelAccess
   ( NodeKernelAccess (..)
   , mkNodeKernelAccess
   , fetchBlock
   , grabNodeKernelAccess
+  , ChainChange (..)
+  , ChainFollower (..)
+  , withFollower
   )
 where
 
@@ -16,7 +21,7 @@ import Cardano.Api.Consensus qualified as Consensus
 import Cardano.Rpc.Server.Internal.Monad (MonadRpc, grab)
 import Cardano.Rpc.Server.NodeKernelAccess.Type
 
-import RIO (atomically, throwIO)
+import RIO (MonadUnliftIO, atomically, bracket, throwIO, withRunInIO)
 
 import Control.Tracer (Tracer, traceWith)
 import Data.ByteString (ByteString)
@@ -34,17 +39,19 @@ mkNodeKernelAccess
   -> Consensus.BlockType blk
   -- ^ Block type witness
   -> Consensus.TopLevelConfig blk
-  -- ^ Top-level consensus config (for system start and era history)
+  -- ^ Top-level consensus config (for system start, era history and the
+  -- security parameter)
   -> Consensus.NodeKernel IO addrNTN addrNTC blk
   -- ^ Consensus node kernel
   -> m (Maybe NodeKernelAccess)
 mkNodeKernelAccess tracer blockType topLevelConfig kernel = case blockType of
   Consensus.CardanoBlockType ->
-    pure $ Just NodeKernelAccess{chainDb, systemStart, readEraHistory}
+    pure $ Just NodeKernelAccess{chainDb, systemStart, readEraHistory, securityParam}
    where
     chainDb = Consensus.getChainDB kernel
     ledgerConfig = Consensus.configLedger topLevelConfig
     systemStart = Consensus.nodeSystemStart topLevelConfig
+    securityParam = Consensus.configSecurityParam topLevelConfig
     -- Read the current ledger state (cheap STM TVar read) and recompute
     -- the era summary on every call - O(number_of_eras).
     -- This is the same approach consensus uses for GetInterpreter queries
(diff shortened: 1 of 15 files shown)


cardano-rpc/src/Cardano/Rpc/Server/NodeKernelAccess.hs
```diff
@@ -94,3 +101,81 @@ fetchBlock NodeKernelAccess{chainDb} slot (HeaderHash shortHash) = do
...
+        (Consensus.newFollower chainDb registry Consensus.SelectedChain component)
+        Consensus.followerClose
+        (runInIO . action . toChainFollower)
+ where
+  component
+    :: Consensus.BlockComponent
+         (Consensus.CardanoBlock Consensus.StandardCrypto)
+         (ByteString, BlockInMode)
+  component =
+    (,) <$> fmap BSL.toStrict Consensus.GetRawBlock <*> fmap fromConsensusBlock Consensus.GetBlock
```
Jimbo4350, line 157: What about clients who just want the raw bytes and don't want to pay the decode cost at this point?

Jimbo4350 (approved):
LGTM. Some of the comments on your functions are difficult to decipher however I wouldn't block the PR on that. One comment for you to consider.

cardano-rpc/src/Cardano/Rpc/Server/NodeKernelAccess.hs
carbolymer: Great question - and I'm glad you asked, because this is exactly where I want to take this next!

Right now they can't avoid it: the `AnyChainBlock` response message carries both `native_bytes` and the parsed `cardano` block, so the handler has to decode every block just to populate the response, regardless of what the client wants.

The exciting part: the spec already has the answer built in - the `field_mask` in `FollowTipRequest`/`FetchBlockRequest`, which cardano-rpc currently ignores (like every other UTxO RPC method here).
I've just opened #1273 to fix that properly: the mask will drive response construction itself, so a mask excluding the parsed `cardano.*` fields selects a raw-bytes-only `BlockComponent` and the decode never happens at all.
Raw-bytes-only followers get dramatically cheaper streaming, practically for free.
Out of scope for this PR, but very much on the roadmap!

PR #1271, opened 2026-07-28
Fix key witness count underestimation for votes in fee estimation

### LINKED ISSUE #722: [FR] - `TxVotingProcedures` accepts `ScriptWitness` but not broader `Witness` type
*External*

**Area**
Transaction building.

**Describe the feature you'd like**
Constructors such as `TxWithdrawals`, `TxCertificates` accepts `Witness` under `BuildTxWith` but [`TxVotingProcedures`](https://github.com/IntersectMBO/cardano-api/blob/9386533b1c9f5e262287d4cca6dbaad4d4365eec/cardano-api/internal/Cardano/Api/Tx/Body.hs#L1289) utilizes `ScriptWitness` which does not include key witnesses. Is this deliberate? Is it that key witnesses don't make sense in this context? If key witnesses make sense for voting procedures, then not including them may mess up with functions like `estimateTransactionKeyWitnessCount`.
Thanks!

carbolymer:
# Context

Closes #722.

Both variants of `estimateTransactionKeyWitnessCount` (experimental `Cardano.Api.Experimental.Tx.Internal.Fee` and legacy `Cardano.Api.Tx.Internal.Fee`) ignored `txVotingProcedures` entirely, so a transaction carrying votes from key-credentialed voters (key-hash DReps, CC hot keys, SPOs) had its key witness count underestimated.
This understates the fee and can lead to `FeeTooSmallUTxO`.

Both estimators now derive the count from the ledger `L.VotingProcedures` voter credentials, folded into a `Set` of witness key hashes exactly like ledger's [`voterWitnesses`](https://github.com/IntersectMBO/cardano-ledger/blob/cardano-ledger-conway-1.23.0.0/eras/conway/impl/src/Cardano/Ledger/Conway/UTxO.hs#L182): a `CommitteeVoter`/`DRepVoter` needs a key witness only when its credential is a `KeyHashObj`, a `StakePoolVoter` always does, and a voter voting on several governance actions, or a key hash appearing under several governance roles, is counted once.
`estimateTransactionKeyWitnessCount` is also newly exported from `Cardano.Api.Experimental` (it was previously unreachable outside the package).
The ledger requires no key witness for proposal procedures, so the vote fix deliberately adds no proposal term; the experimental estimator keeps its pre-existing conservative charge of one witness per proposal, now documented in its haddock.

Both estimators also deduplicate key witnesses across every role whose key hashes the transaction body itself pins down: certificates, withdrawals, extra key witnesses and votes are folded into a single set of witness key hashes, so a key required by several of those roles is counted once.
Pool registration certificates now count one key witness for the operator and each owner, closing another pre-existing under-estimate (the ledger [requires every owner to sign](https://github.com/IntersectMBO/cardano-ledger/blob/cardano-ledger-conway-1.23.0.0/eras/shelley/impl/src/Cardano/Ledger/Shelley/UTxO.hs#L237-L238), but only the operator was counted).
Inputs and collateral are still counted per input on top of that set, since their key hashes live in the spent outputs and are only known with a `UTxO`; the result stays an upper bound, and `calculateMinTxFee` remains the exact path when a `UTxO` is available.

Note that the type asymmetry the issue asks about (`ScriptWitness` vs `Witness`) is already resolved in the experimental API via `AnyWitness`; this PR closes the remaining fee-estimation gap in both APIs.

# How to trust this PR

Both fixes mirror ledger's own witness collection ([`voterWitnesses`](https://github.com/IntersectMBO/cardano-ledger/blob/cardano-ledger-conway-1.23.0.0/eras/conway/impl/src/Cardano/Ledger/Conway/UTxO.hs#L182) and the [`getWitsVKeyNeeded` union](https://github.com/IntersectMBO/cardano-ledger/blob/cardano-ledger-conway-1.23.0.0/eras/conway/impl/src/Cardano/Ledger/Conway/UTxO.hs#L172) in `Cardano.Ledger.Conway.UTxO`), so the credential-matching and deduplication logic is not new invention, it is the same rule the ledger itself applies when building the witness set.

Two hedgehog property tests cover the fix, each generating a small pool of key-credentialed and script-credentialed voters that vote on random non-empty subsets of several governance action ids, so the same voter recurs across action ids, and asserting the exact expected key witness count:

- `Test.Cardano.Api.Experimental.Fee`: `estimateTransactionKeyWitnessCount counts key witnesses required by key-credentialed voters`, which also exercises the `mkTxVotingProcedures` merge path.
- `Test.Cardano.Api.TxBody`: `vote key witness count`, which isolates the vote contribution by comparing the estimate for a randomly generated tx body with and without the generated votes attached.

Both suites additionally gain a `cross-role key witness dedupe` property: a stake key that both withdraws rewards and is deregistered by a certificate in the same transaction is counted once, while an unrelated key still counts.
A `pool registration counts operator and every owner` property in both suites covers the owner counting, including an owner shared with a withdrawal.

Both tests fail against the pre-fix code with a bare `0 === 1` (or similar), since the old record patterns simply never destructured `txVotingProcedures`.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/src/Cardano/Api/Experimental/Tx/Internal/Fee.hs
+++ b/cardano-api/src/Cardano/Api/Experimental/Tx/Internal/Fee.hs
@@ -20,6 +20,7 @@ module Cardano.Api.Experimental.Tx.Internal.Fee
   , calcMinFeeRecursive
   , collectTxBodyScriptWitnesses
   , estimateBalancedTxBody
+  , estimateTransactionKeyWitnessCount
   , evaluateTransaction
   , TxEvaluationResult (..)
   , evaluateTransactionExecutionUnits
@@ -1789,6 +1790,7 @@ estimateTransactionKeyWitnessCount
     , txWithdrawals
     , txCertificates
     , txProposalProcedures
+    , txVotingProcedures
     } =
     fromIntegral $
       sum (map estimateTxInWitnesses txIns)
@@ -1806,6 +1808,10 @@ estimateTransactionKeyWitnessCount
         + case txProposalProcedures of
           Just (TxProposalProcedures m) ->
             OMap.size m
+          Nothing -> 0
+        + case txVotingProcedures of
+          Just (TxVotingProcedures _ voteWits) ->
+            length [() | AnyKeyWitnessPlaceholder <- Map.elems voteWits]
           Nothing -> 0
    where
     estimateTxInWitnesses :: (TxIn, AnyWitness (LedgerEra era)) -> Int
--- a/cardano-api/src/Cardano/Api/Experimental/Tx/Internal/Fee.hs
+++ b/cardano-api/src/Cardano/Api/Experimental/Tx/Internal/Fee.hs
@@ -20,6 +20,7 @@ module Cardano.Api.Experimental.Tx.Internal.Fee
   , calcMinFeeRecursive
   , collectTxBodyScriptWitnesses
   , estimateBalancedTxBody
+  , estimateTransactionKeyWitnessCount
   , evaluateTransaction
   , TxEvaluationResult (..)
   , evaluateTransactionExecutionUnits
@@ -75,7 +76,8 @@ import Cardano.Ledger.Alonzo.Core qualified as Ledger
 import Cardano.Ledger.Api qualified as L
 import Cardano.Ledger.Coin qualified as L
 import Cardano.Ledger.Conway.Governance qualified as L
-import Cardano.Ledger.Credential as Ledger (Credential)
+import Cardano.Ledger.Credential as Ledger (Credential, credKeyHashWitness)
+import Cardano.Ledger.Keys (asWitness)
 import Cardano.Ledger.Val qualified as L
 
 import Control.Monad
@@ -1744,8 +1746,13 @@ handleExUnitsErrors ScriptInvalid failuresMap exUnitsMap
   | null failuresMap = Left TxBodyScriptBadScriptValidity
   | otherwise = Right $ Map.map (\_ -> ExecutionUnits 0 0) failuresMap <> exUnitsMap
 
--- | Provide and approximate count of the key witnesses (i.e. signatures)
+-- | Provide an approximate count of the key witnesses (i.e. signatures)
 -- required for a transaction.
+--
+-- Certificates, withdrawals, extra key witnesses and votes are deduplicated against
+-- each other, mirroring the key hash set ledger's @getWitsVKeyNeeded@ computes, so a
+-- key acting in several of those roles is counted once. A pool registration
+-- certificate counts one key witness for the operator and one for every owner.
 --
 -- This estimate is not exact and may overestimate the required number of witnesses.
 -- The function makes conservative assumptions, including:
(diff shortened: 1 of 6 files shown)


cardano-api/test/cardano-api-test/Test/Cardano/Api/Experimental/Fee.hs
```diff
@@ -1345,3 +1376,61 @@ genShuffledCertsWithCount = do
...
+      govActionId =
+        L.GovActionId
+          (L.TxId (L.unsafeMakeSafeHash "0000000000000000000000000000000000000000000000000000000000000000"))
+          (L.GovActionIx 0)
+      votingProcedure = L.VotingProcedure{L.vProcVote = L.VoteYes, L.vProcAnchor = L.SNothing}
+      scriptWitness = Exp.AnySimpleScriptWitness (Exp.SReferenceScript refTxIn)
+      allVoters = keyVoters <> scriptVoters
+      votingProcedures =
+        L.VotingProcedures $
+          Map.fromList [(voter, Map.singleton govActionId votingProcedure) | voter <- allVoters]
```
palas, line 1430: We should try with several `govActionId` and also having the same voter vote for several action ids. The easiest way would be generating a relatively small pool of potential voters and reusing them to create collisions.

cardano-api/src/Cardano/Api/Experimental/Tx/Internal/Fee.hs
```diff
@@ -1807,6 +1809,10 @@ estimateTransactionKeyWitnessCount
           Just (TxProposalProcedures m) ->
             OMap.size m
           Nothing -> 0
+        + case txVotingProcedures of
+          Just (TxVotingProcedures _ voteWits) ->
```
palas, line 1813: I think here it would be more reliable to look at the voting procedures (first field) instead of `voteWits`, like in the version of this for the traditional API. And that is what the ledger does [here](https://cardano-ledger.cardano.intersectmbo.org/cardano-ledger-conway/src/Cardano.Ledger.Conway.UTxO.html#voterWitnesses)

palas (commented):
Nice catch! This is a net improvement, but I think there are two things that could be improved, one in the implementation, and one in the tests

cardano-api/src/Cardano/Api/Experimental/Tx/Internal/Fee.hs
```diff
@@ -1807,6 +1809,10 @@ estimateTransactionKeyWitnessCount
...
+            length [() | AnyKeyWitnessPlaceholder <- Map.elems voteWits]
```
Jimbo4350, line 1814: I just realised we may be counting the same key witness several times over. The ledger gathers all required key witnesses into a single set, so a key that appears in more than one role only needs one witness. For example, the key witnessing the spend of a transaction input may also witness a vote and therefore it is counted twice.

Jimbo4350 (changes requested):

cardano-api/src/Cardano/Api/Experimental/Tx/Internal/Fee.hs
carbolymer: You're right about the ledger: `getConwayWitsVKeyNeeded` folds everything into a single `Set (KeyHash Witness)`, so a key acting in several roles needs only one witness on chain. This function can't replicate that fully: it has no `UTxO`, and input key hashes live only in the spent outputs, so it can't detect your input-also-votes example even in principle. That's why its haddock documents it as a conservative overestimate; over-counting only rounds the fee up, while the under-count fixed here is what causes `FeeTooSmallUTxO`. For exact fees, `calculateMinTxFee` already goes through the ledger's `calcMinFeeTx`, which counts the size of the deduplicated `getWitsVKeyNeeded` set, votes included.

What this PR now does about it: both estimators fold every role whose key hashes the body itself pins down - certificates, withdrawals, extra key witnesses and votes - into a single deduplicated `Set (KeyHash Witness)`, mirroring the ledger's union, so a key required by several of those roles counts once. There's a property test covering a stake key that both withdraws and is deregistered in one transaction. Only inputs and collateral are still counted per input on top of that set, which is the part that cannot be deduplicated without a `UTxO`; the haddock documents that remaining limitation.

Jimbo4350 (approved):

PR #1274, opened 2026-07-31
Fail on collateral without Plutus script instead of dropping collateral

palas:
# Context

This is a follow-up PR to address the following two reviews:

- https://github.com/IntersectMBO/cardano-api/pull/1265#discussion_r3657454013
- https://github.com/IntersectMBO/cardano-api/pull/1265#pullrequestreview-4774208788

# How to trust this PR

Look at the two commits independently:
- The first commit only changes the behaviour in case of collateral provided but not Plutus script to a failure.
- The second commit should be a pure refactoring, moved code.

# Checklist

- [x] Self-reviewed the diff


.changes/20260730_cardano_api_collateral_without_plutus_fails.yml
```diff
@@ -0,0 +1,6 @@
+project: cardano-api
+pr: 1274
+kind:
+  - breaking
+description: |
+  Since #1265, `makeTransactionBodyAutoBalance` and `estimateBalancedTxBody` (both the traditional and the experimental versions) silently remove the collateral inputs from transactions without Plutus scripts. From this PR, instead of silently dropping the collateral inputs, it fails with the new `CollateralWithoutPlutusScripts` error. The ledger requires collateral only for transactions that run Plutus scripts and ignores it otherwise: the collateral inputs would only cost fees and require the collateral UTxOs to stay unspent, so they are most likely a mistake.
```
carbolymer: It would be nice to turn the PR number into a link

carbolymer: Needs rewording. "so they are most likely a mistake." and "From this PR" do not fit into changelog

cardano-api/test/cardano-api-test/Test/Cardano/Api/Transaction/Utils.hs
```diff
@@ -0,0 +1,193 @@
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE NumericUnderscores #-}
+{-# LANGUAGE OverloadedLists #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# OPTIONS_GHC -Wno-deprecations #-}
+
+module Test.Cardano.Api.Transaction.Utils
```
carbolymer: [`Utils` modules are an anti-pattern](https://mattilehtinen.com/articles/dunghill-anti-pattern-why-utility-classes-and-modules-smell/). Can you use more-specific module name? Some of the functions exist in similar form already.

cardano-api/test/cardano-api-test/Test/Cardano/Api/Experimental/Collateral.hs
```diff
@@ -0,0 +1,477 @@
...
+import Test.Tasty (TestTree, testGroup)
+import Test.Tasty.Hedgehog (testProperty)
+
+-- | Tests in this module can be run by themselves by writing:
+-- ```bash
+-- cabal test cardano-api-test --test-options="--pattern=Test.Cardano.Api.Experimental.Collateral"
+-- ```
+tests :: TestTree
+tests =
+  testGroup
```
carbolymer: in some tests `tests` is at the bottom of the file, in the other ones its at the top. We should stick to a single convention.

cardano-api/src/Cardano/Api/Tx/Internal/Fee.hs
```diff
@@ -1253,8 +1251,21 @@ hasPlutusScriptWitnesses sbe txbodycontent =
           collectTxBodyScriptWitnesses sbe txbodycontent
       ]
 
--- | Why 'calcReturnAndTotalCollateral' could not compute the collateral
--- fields: the ledger would reject a transaction built with them.
+-- | Fail with 'CollateralWithoutPlutusScripts' when the transaction has
+-- collateral inputs but no Plutus script witnesses.
+checkCollateralOnlyWithPlutusScripts
+  :: ShelleyBasedEra era -> TxBodyContent BuildTx era -> Either CollateralError ()
+checkCollateralOnlyWithPlutusScripts sbe txbodycontent =
```
carbolymer, line 1258: The new `checkCollateralOnlyWithPlutusScripts` guard covers `makeTransactionBodyAutoBalance` and `estimateBalancedTxBody`, but `calcReturnAndTotalCollateral` is still exported from this module and re-exported all the way up (`Cardano.Api.Tx` -> `Cardano.Api`), with no validation of its own.
So an external caller using `calcReturnAndTotalCollateral` directly with collateral inputs and no Plutus scripts still gets silently computed return/total collateral - the behaviour this PR removes from the balancing entry points.
The new property test even documents the gap: "Unreachable: calcReturnAndTotalCollateral does not check for Plutus scripts, the balancing functions do before calling it".
Would it make sense to either move the check into `calcReturnAndTotalCollateral` itself, or stop exporting it (the experimental module already keeps its copy internal)?
Not blocking - its haddock has always described it as an unvalidated low-level helper - but worth deciding deliberately.

cardano-api/test/cardano-api-test/Test/Cardano/Api/Experimental/Collateral.hs
```diff
@@ -0,0 +1,477 @@
...
+  let era = Exp.ConwayEra
+      sbe = convert era
+      systemStart = Api.SystemStart $ Time.posixSecondsToUTCTime 0
+      epochInfo =
+        Api.LedgerEpochInfo $
+          Slotting.fixedEpochInfo (Slotting.EpochSize 100) (Slotting.mkSlotLength 1000)
+
+  fundingTxIn <-
+    H.evalEither $
+      Api.runParser Api.parseTxIn "01f4b788593d4f70de2a45c2e1e87088bfbdfa29577ae1b62aba60e095e3ab53#0"
```
carbolymer: This module inlines `H.evalEither $ Api.runParser Api.parseTxIn "..."` for the same two tx-in literals at five sites.
The same commit adds `mkTxIn` to `Test.Cardano.Api.Transaction.Utils` for exactly this purpose and uses it in the sibling `Test.Cardano.Api.Transaction.Collateral`.
Importing it here removes ~15 lines of boilerplate; the helper is API-flavour-agnostic.

carbolymer (approved):

.changes/20260730_cardano_api_collateral_without_plutus_fails.yml
palas: I've removed the reference to the PR and, instead, I've also removed the change log entry for the PR that caused the collateral inputs to be removed in the first place. Because there hasn't been a release since then, this will make the narrative easier in the change log

cardano-api/src/Cardano/Api/Tx/Internal/Fee.hs
palas: I was almost convinced, but now I think that that would be out of scope for `calcReturnAndTotalCollateral`. For one main reason: It doesn't have enough info to make the call

And we could give it the information, but it seems overkill.

About not exporting it: I agree, let's not export it. But, like you say, if experimental module already doesn't, we may as well just wait until we deprecate the other one and then it gets un-exported naturally

.changes/20260730_cardano_api_collateral_without_plutus_fails.yml
palas: Addressed in: https://github.com/IntersectMBO/cardano-api/pull/1274/changes/2258e85e5f27ad0a9e9f630cb37958a6f390dbf8

palas: Addressed in: https://github.com/IntersectMBO/cardano-api/pull/1274/changes/2258e85e5f27ad0a9e9f630cb37958a6f390dbf8

cardano-api/test/cardano-api-test/Test/Cardano/Api/Transaction/Utils.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-api/pull/1274/changes/2258e85e5f27ad0a9e9f630cb37958a6f390dbf8

cardano-api/test/cardano-api-test/Test/Cardano/Api/Experimental/Collateral.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-api/pull/1274/changes/2258e85e5f27ad0a9e9f630cb37958a6f390dbf8

cardano-api/test/cardano-api-test/Test/Cardano/Api/Experimental/Collateral.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-api/pull/1274/changes/2258e85e5f27ad0a9e9f630cb37958a6f390dbf8

PR #1282, opened 2026-08-07
Add support for Plutus scripts on compatible Tx

carbolymer:
## Description

Adds plutus spending script support to the compatible transaction API, so pre-Conway script transactions (PlutusV1/V2, Alonzo/Babbage) can be built without the deprecated API.
Motivated by the tx-generator migration in IntersectMBO/cardano-node#6602.

Changes:

- `createCompatibleTx` supports script-witnessed inputs: spending redeemer pointers, witness datums, script integrity hash, collateral inputs, transaction metadata and validity upper bound.
- Breaking: all transaction content is now passed as a single `CompatibleTxBodyContent` record, and failures are reported via the new `CompatibleTxError`.
- Breaking: `mkSpendingScriptDatum` is exported for building spending-purpose datums.
- Fixes: certificate, proposal and vote script witnesses were missing from the witness set; proposal reference inputs were overwritten.

Review follow-ups included:

- Vote script witnesses are wired through fully (redeemers, scripts, script integrity hash, reference inputs), with a regression test.
- Removed the unreachable `CompatibleTxProtocolParametersConversionError` constructor.
- Readability: the pure bindings moved out of `createCompatibleTx`'s `Either` block, `setRefInputs` collapsed to a single comprehension, and `setScriptWitnesses` rewritten without `Endo`.
- Duplicate-`TxIn` validation is split out into #1324, since it equally affects the legacy and experimental builders.

## Checklist

- [x] New tests: plutus witness properties for the compatible builder in `Test.Cardano.Api.Experimental.Fee`
- [x] Changelog fragment (`.changes/20260807_cardano_api_compatible_plutus.yml`)
- [x] Self-reviewed

### DIFF
--- a/.changes/20260807_cardano_api_compatible_plutus.yml
+++ b/.changes/20260807_cardano_api_compatible_plutus.yml
@@ -5,4 +5,4 @@ kind:
   - breaking
   - bugfix
 description: |
-  Add plutus spending script support to createCompatibleTx. Breaking: it now takes all transaction content as a single CompatibleTxBodyContent record and fails with the new CompatibleTxError. Export mkSpendingScriptDatum. Fix certificate and proposal script witnesses missing from the witness set, and proposal reference inputs being overwritten. Remove the redundant StakeCredential field from WitTxCert.
+  Add plutus spending script support to createCompatibleTx. Breaking: it now takes all transaction content as a single CompatibleTxBodyContent record and fails with the new CompatibleTxError. Export mkSpendingScriptDatum. Fix certificate, proposal and vote script witnesses missing from the witness set, and proposal reference inputs being overwritten. Remove the redundant StakeCredential field from WitTxCert.
--- a/.changes/20260807_cardano_api_compatible_plutus.yml
+++ b/.changes/20260807_cardano_api_compatible_plutus.yml
@@ -1,0 +1,8 @@
+project: cardano-api
+pr: 1282
+kind:
+  - feature
+  - breaking
+  - bugfix
+description: |
+  Add plutus spending script support to createCompatibleTx. Breaking: it now takes all transaction content as a single CompatibleTxBodyContent record and fails with the new CompatibleTxError. Export mkSpendingScriptDatum. Fix certificate and proposal script witnesses missing from the witness set, and proposal reference inputs being overwritten. Remove the redundant StakeCredential field from WitTxCert.
--- a/cardano-api/src/Cardano/Api/Compatible/Tx.hs
+++ b/cardano-api/src/Cardano/Api/Compatible/Tx.hs
@@ -5,30 +5,41 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeOperators #-}
 
--- | This module provides a way to construct a simple transaction over all eras.
--- It is exposed for testing purposes only.
+-- | This module provides a transaction builder covering all Shelley-based
+-- eras, for consumers that still need pre-Conway support (e.g.
+-- tx-generator). The experimental API only covers the current and next
+-- era.
 module Cardano.Api.Compatible.Tx
   ( AnyProtocolUpdate (..)
   , AnyVote (..)
+  , CompatibleTxBodyContent (..)
+  , CompatibleTxError (..)
+  , defaultCompatibleTxBodyContent
   , createCompatibleTx
   , addWitnesses
   )
 where
 
 import Cardano.Api.Era
-import Cardano.Api.Experimental.AnyScriptWitness
+import Cardano.Api.Error (Error (..))
 import Cardano.Api.Experimental.Era (obtainCommonConstraints)
+import Cardano.Api.Experimental.Era qualified as Exp
+import Cardano.Api.Experimental.Plutus
+  ( Witnessable (..)
+  , WitnessableItem (..)
+  , getAnyWitnessRedeemerPointerMap
+  , obtainAlonzoScriptPurposeConstraints
+  )
 import Cardano.Api.Experimental.Tx qualified as Exp
 import Cardano.Api.Experimental.Tx.Internal.AnyWitness
 import Cardano.Api.Experimental.Tx.Internal.AnyWitness qualified as Exp
 import Cardano.Api.Experimental.Tx.Internal.Certificate qualified as Exp
-import Cardano.Api.Plutus.Internal.Script
+import Cardano.Api.Monad.Error ((?!))
 import Cardano.Api.ProtocolParameters
 import Cardano.Api.Tx.Internal.Body hiding
   ( convCertificates
-  , indexTxCertificates
-  , indexWitnessedTxProposalProcedures
   )
+import Cardano.Api.Tx.Internal.Body.Lens qualified as A
 import Cardano.Api.Tx.Internal.Sign
 import Cardano.Api.Value.Internal
 
@@ -36,7 +47,11 @@ import Cardano.Ledger.Alonzo.Tx qualified as L
 import Cardano.Ledger.Alonzo.TxWits qualified as Alonzo
 import Cardano.Ledger.Api qualified as L
 import Cardano.Ledger.Core qualified as L
+import Cardano.Ledger.TxIn qualified as L
+import Cardano.Slotting.Slot (SlotNo)
 
+import Data.List qualified as L
+import Data.Map.Ordered.Strict qualified as OMap
 import Data.Map.Strict (Map)
 import Data.Map.Strict qualified as Map
 import Data.Maybe
@@ -44,7 +59,9 @@ import Data.Maybe.Strict
 import Data.Monoid
 import Data.OSet.Strict (OSet)
 import Data.Sequence.Strict qualified as Seq
-import GHC.Exts (IsList (..))
+import Data.Set (Set)
+import Data.Set qualified as Set
+import GHC.Exts
 import Lens.Micro hiding (ix)
 
 data AnyProtocolUpdate era where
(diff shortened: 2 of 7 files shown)


cardano-api/src/Cardano/Api/Compatible/Tx.hs
```diff
@@ -102,11 +176,14 @@ createCompatibleTx sbe ins outs extraDatums txFee' anyProtocolUpdate anyVote txC
...
+              proposalWitnesses
+                :: [(Witnessable ProposalItem (ShelleyLedgerEra era), AnyWitness (ShelleyLedgerEra era))]
               proposalWitnesses =
-                [ (ix, witness)
-                | (_, (ix, witness)) <-
-                    indexWitnessedTxProposalProcedures conwayOnwards proposalProcedures
-                ]
+                Exp.obtainCommonConstraints (convert conwayOnwards) $
+                  Exp.extractWitnessableProposals $
+                    Just proposalProcedures
```
palas, line 186: I think the function is getting too large (even ignoring the `where` bits).

And `proposalWitnesses` doesn't need to be in the `case`, it could be moved to the `where` (having its own small `case`), and that would allow moving `allWitnesses` and `sData` to the `where` too (`txAuxData` can moved to the `where` already), and I think that would make the body of the function much more manageable.

Actually, `updateTxBody` wouldn't need to be here either, because I think it doesn't make use of the `Either` monad, so it wouldn't need to be monadic, it could be moved to the `where` too, but that is out of scope of the PR

cardano-api/src/Cardano/Api/Compatible/Tx.hs
```diff
@@ -121,49 +198,184 @@ createCompatibleTx sbe ins outs extraDatums txFee' anyProtocolUpdate anyVote txC
...
     let refInputs =
           [ toShelleyTxIn refInput
-          | (_, _, Just (_, wit)) <- indexedTxCerts
+          | wit <- witnessedCertWitnesses
           , refInput <- maybeToList $ getAnyWitnessReferenceInput wit
           ]
+            ++ [ toShelleyTxIn refInput
+               | (_, wit) <- ins
+               , refInput <- maybeToList $ getAnyWitnessReferenceInput wit
+               ]
```
palas, line 297:
```suggestion
  -- certificates and witnessed spending inputs.
  setRefInputs :: Endo (L.TxBody L.TopTx (ShelleyLedgerEra era))
  setRefInputs = do
    let refInputs =
          [ toShelleyTxIn refInput
          | wit <- witnessedCertWitnesses <> map snd ins
          , refInput <- maybeToList $ getAnyWitnessReferenceInput wit
          ]
```
This can be done with a single list comprehension

cardano-api/src/Cardano/Api/Compatible/Tx.hs
```diff
@@ -174,50 +386,58 @@ createCompatibleTx sbe ins outs extraDatums txFee' anyProtocolUpdate anyVote txC
...
+                         ]
+                   )
           )
       ]
 
 convSimpleScripts
   :: ShelleyLedgerEra era ~ ledgerera
   => ShelleyBasedEra era
-  -> [(ScriptWitnessIndex, Exp.AnyWitness (ShelleyLedgerEra era))]
+  -> [Exp.AnyWitness (ShelleyLedgerEra era)]
```
palas, line 435:
```suggestion
  setScriptWitnesses sData scriptWitnesses =
    plutusAdditions . simpleScriptAdditions
   where
    plutusAdditions :: L.TxWits (ShelleyLedgerEra era) -> L.TxWits (ShelleyLedgerEra era)
    plutusAdditions =
      forEraInEon era id $ \aeo ->
        alonzoEraOnwardsConstraints aeo $
          obtainAlonzoScriptPurposeConstraints aeo $
            let (datums, redeemers) =
                  fromMaybe mempty sData
                    :: (L.TxDats (ShelleyLedgerEra era), L.Redeemers (ShelleyLedgerEra era))
                -- 'getAnyWitnessScript' covers both plutus and simple
                -- scripts. The simple ones overlap harmlessly (same
                -- hash, same script) with 'simpleScriptAdditions',
                -- which is still needed for pre-Alonzo eras where this
                -- addition is 'id'.
                plutusAndSimpleScripts =
                  mapMaybe getAnyWitnessScript scriptWitnesses
             in (L.datsTxWitsL .~ datums)
                  . (L.rdmrsTxWitsL %~ (<> redeemers))
                  . ( L.scriptTxWitsL
                        %~ (<> Map.fromList [(L.hashScript sw, sw) | sw <- plutusAndSimpleScripts])
                    )

    simpleScriptAdditions :: L.TxWits (ShelleyLedgerEra era) -> L.TxWits (ShelleyLedgerEra era)
    simpleScriptAdditions =
      forEraInEon era id $ \aeo ->
        allegraEraOnwardsConstraints aeo $
          L.scriptTxWitsL
            %~ ( <>
                   Map.fromList
                     [ (L.hashScript sw, sw)
                     | sw <- convSimpleScripts sbe scriptWitnesses
                     ]
               )

convSimpleScripts
:: ShelleyLedgerEra era ~ ledgerera
=> ShelleyBasedEra era
  -> [Exp.AnyWitness (ShelleyLedgerEra era)]
```
I think `Endo`s in this function only add complexity. They were already here before the PR, but since we are changing most of the function...

cardano-api/src/Cardano/Api/Compatible/Tx.hs
```diff
@@ -121,49 +198,184 @@ createCompatibleTx sbe ins outs extraDatums txFee' anyProtocolUpdate anyVote txC
...
           pure (updateTxBody, proposalWitnesses)
 
+    let
+      -- Flat witnesses from all three script-witnessable categories
+      -- (certificates, proposals, inputs), for 'setScriptIntegrityHash'
+      -- and 'setScriptWitnesses' to collect datums, scripts and plutus
+      -- languages from. Only the redeemer pointer map (built inside
+      -- 'convScriptData'') needs per-category indexing.
+      allWitnesses :: [AnyWitness (ShelleyLedgerEra era)]
+      allWitnesses = witnessedCertWitnesses <> map snd proposalWitnesses <> map snd ins
```
palas, line 208: I think we are missing `TxVotingProcedures` witnesses

cardano-api/src/Cardano/Api/Compatible/Tx.hs
```diff
@@ -228,55 +448,78 @@ convCertificates (Exp.TxCertificates cs) =
...
-            ) <-
-            wits
-        , Just i <- [fromScriptWitnessIndex w idx]
-        ]
+-- | Copy of an experimental extractWitnessableTxIns not requiring 'IsEra era'
+witnessableTxIns
+  :: L.AlonzoEraScript ledgerera
+  => [(TxIn, AnyWitness ledgerera)]
+  -> [(Witnessable TxInItem ledgerera, AnyWitness ledgerera)]
+witnessableTxIns txIns' = L.nub [(WitTxIn txIn, wit) | (txIn, wit) <- txIns']
```
palas, line 487: We should validate there are no repeated `TxIn`s in the list. Or at least that there are no two `TxIn`s with different witnesses, because that would mess up the redeemer pointers

palas (commented):
I think there are a couple of issues that should be resolved before merging, and I also included a couple of potential readability improvement opportunities that I saw.

I still haven't had time to look at the tests yet too.

But it is looking good 👍

cardano-api/test/cardano-api-test/Test/Cardano/Api/Experimental/Fee.hs
```diff
@@ -854,10 +885,183 @@ prop_createCompatibleTx_preserves_all_certs = H.property $ do
...
+-- | 'createCompatibleTx' must index spending redeemer pointers by the
+-- input's position in the final, sorted ledger input set, never by the
+-- order the caller supplied inputs in. The ledger stores tx inputs in a
+-- 'Data.Set.Set', not a list, so swapping the argument order of two
+-- plutus-witnessed inputs must not change which redeemer lands at which
+-- index.
+prop_createCompatibleTx_redeemer_indices_follow_sorted_inputs :: Property
+prop_createCompatibleTx_redeemer_indices_follow_sorted_inputs = H.property $ do
+  txIn1@(Api.TxIn txId (Api.TxIx txIx)) <- H.forAll genTxIn
+  let txIn2 = Api.TxIn txId (Api.TxIx (txIx + 1))
```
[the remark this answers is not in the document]
carbolymer: This cannot overflow in practice: `Api.TxIx` wraps a machine `Word`, not the ledger's `Word16`, and `genTxIn` builds the index from `Gen.word16 Range.constantBounded`, so `txIx` is at most 65535 and `txIx + 1` is at most 65536. A wraparound making `txIn2` equal `txIn1` is impossible for the same reason. Leaving this as is.

cardano-api/src/Cardano/Api/Compatible/Tx.hs
```diff
@@ -67,26 +83,84 @@ data AnyVote era where
...
+    CompatibleTxMissingScriptIntegrityPParams
+  deriving Show
+
+instance Error CompatibleTxError where
+  prettyError err =
+    case err of
+      CompatibleTxProtocolParametersConversionError e -> prettyError e
+      CompatibleTxMissingScriptIntegrityPParams ->
+        "Plutus script witnesses are present but no protocol parameters were supplied "
+          <> "to compute the script integrity hash."
```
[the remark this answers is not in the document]
carbolymer: Removed in d2c152c. The record already takes ledger PParams, so no conversion path could ever construct it.

cardano-api/src/Cardano/Api/Compatible/Tx.hs
carbolymer: Done in d2c152c.

cardano-api/src/Cardano/Api/Compatible/Tx.hs
carbolymer: Done in d2c152c. The whole Alonzo-onwards branch stays inside the alonzoEraOnwardsConstraints continuation, since datsTxWitsL and rdmrsTxWitsL only typecheck there.

cardano-api/src/Cardano/Api/Compatible/Tx.hs
carbolymer: Good catch, fixed in d2c152c. Vote witnesses now contribute redeemers, scripts, the script integrity hash and reference inputs, same as certificates and proposals, with a regression test asserting all three.

cardano-api/src/Cardano/Api/Compatible/Tx.hs
carbolymer: Done in d2c152c. proposalWitnesses, updateTxBody, allWitnesses, sData and txAuxData moved to the where clause, so the Either block is now just the integrity hash step. One subtlety: the ProposalProcedures arm needed its own shelleyBasedEraConstraints wrap once hoisted, since it had been borrowing the constraint from the do block's wrapper.

cardano-api/src/Cardano/Api/Compatible/Tx.hs
carbolymer: Good catch. I checked how this behaves across the builders: witnessableTxIns is a faithful copy of the experimental extractWitnessableTxIns, which is itself identical to the legacy one, so all three share this gap. Two entries with the same TxIn but different witnesses both survive the nub, and the pointer indices shift against the body's deduplicated input set. The result is rejected client-side by transaction build during execution-unit evaluation, or at submission with ExtraRedeemers for build-raw-style paths; in one narrow case it validates with the extra witness silently ignored, provably never misassigned. A proper fix touches the compatible, experimental and legacy APIs, plus arguably cardano-cli, which already dedupes collateral but not spending inputs, so I split it out into #1324. I also fixed the stale dedupInputsByTxIn reference in the comment here.

cardano-api/src/Cardano/Api/Compatible/Tx.hs
```diff
@@ -67,103 +84,309 @@ data AnyVote era where
...
+            pure $
+              Endo $
+                L.scriptIntegrityHashTxBodyL
+                  .~ SJust
+                    ( L.hashScriptIntegrity $
+                        L.ScriptIntegrity
+                          redeemers
+                          datums
+                          (Set.map (L.getLanguageView protocolParams) languages)
+                    )
```
palas, line 389:
```suggestion
            -- Protocol parameters are consumed only by 'L.getLanguageView',
            -- so they are required exactly when plutus witnesses are present.
            -- A datums-only transaction still needs the hash, but it is
            -- computable without them.
            langViews <-
              if null languages
                then pure Set.empty
                else do
                  protocolParams <-
                    compatibleTxProtocolParams bodyContent
                      ?! CompatibleTxMissingScriptIntegrityPParams
                  pure $ Set.map (L.getLanguageView protocolParams) languages
            pure $
              Endo $
                L.scriptIntegrityHashTxBodyL
                  .~ SJust
                    ( L.hashScriptIntegrity $
                        L.ScriptIntegrity
                          redeemers
                          datums
                          langViews
                    )
```
It seems this constraint is unnecessarily conservative. If `languages` is empty, then `getLanguageView` is never applied, so `protocolParams` are never evaluated, we don't have to error out. Same happens in the ledger code: https://github.com/IntersectMBO/cardano-ledger/blob/f649f9751074d2ab3de033fc3912f29c9862c1f5/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs#L420

.changes/20260807_cardano_api_compatible_plutus.yml
```diff
@@ -0,0 +1,8 @@
+project: cardano-api
+pr: 1282
+kind:
+  - feature
+  - breaking
+  - bugfix
+description: |
+  Add plutus spending script support to createCompatibleTx. Breaking: it now takes all transaction content as a single CompatibleTxBodyContent record and fails with the new CompatibleTxError. Export mkSpendingScriptDatum. Fix certificate, proposal and vote script witnesses missing from the witness set, and proposal reference inputs being overwritten. Remove the redundant StakeCredential field from WitTxCert.
```
palas, line 8: The last sentence ("Remove the redundant StakeCredential field from WitTxCert") now belongs to #1288, and is in its own changelog fragment, so we can drop it here.

cardano-api/src/Cardano/Api/Compatible/Tx.hs
```diff
@@ -67,103 +84,309 @@ data AnyVote era where
...
+  , compatibleTxFee :: Lovelace
+  -- ^ Fee.
+  , compatibleTxProtocolUpdate :: AnyProtocolUpdate era
+  -- ^ Era-appropriate protocol update: Shelley-Babbage proposal, Conway-onwards proposal procedure, or none.
+  , compatibleTxVotingProcedures :: AnyVote era
+  -- ^ Governance votes, Conway onwards; 'NoVotes' otherwise.
+  , compatibleTxCertificates :: Exp.TxCertificates (ShelleyLedgerEra era)
+  -- ^ Certificates, witnessed or not.
+  , compatibleTxInsCollateral :: [TxIn]
+  -- ^ Collateral inputs. Meaningful only Alonzo onwards; supply non-empty only when using plutus spending witnesses.
```
palas, line 104:
```suggestion
  -- ^ Collateral inputs. Meaningful only Alonzo onwards; supply non-empty only when using plutus witnesses.
```
Any plutus witness will spend the collateral, not just spending witnesses, also certificates, proposals, and votes

palas (approved):
Nice work! I think it is all correct apart from https://github.com/IntersectMBO/cardano-api/issues/1324. I wrote 3 small things that I found

PR #1286, opened 2026-08-11
Make owner-permission file writes atomic on POSIX

### LINKED ISSUE #1254: Make POSIX owner-permission file writes atomic via temp file + rename
Since #1248, the POSIX implementation of `handleFileForWritingWithOwnerPermission` opens the target file with `O_TRUNC`, so overwriting a file with shorter content no longer leaves stale trailing bytes. However, the write is still not atomic: if it fails midway (crash, full disk, …), the previous file contents are already destroyed and a truncated file is left behind.

The Windows implementation already avoids this by writing to a freshly created temporary file and renaming it over the target. Doing the same on POSIX (creating a temp file with `0600` permissions in the target directory, write the contents, then `rename(2)`) would keep the old file intact until the new one is complete.

The currently documented behaviour that pre-existing files keep their permission bits would be lost, since the replacement file would always end up `0600`. The haddocks (e.g. `writeFileTextEnvelopeWithOwnerPermissions`) will need updating to match.

 Affected functions: `writeFileTextEnvelopeWithOwnerPermissions`, `writeByteStringFileWithOwnerPermissions`, `writeLazyByteStringFileWithOwnerPermissions`, `writeTextFileWithOwnerPermissions`, and `writeSecrets`.

_Originally posted by @carbolymer in https://github.com/IntersectMBO/cardano-api/pull/1248#discussion_r3571791242_

palas:
# Context

Fixes #1254.

Old: the target was opened with O_TRUNC, so a crash or a full disk mid-write destroyed the previous contents and left a truncated file.

New: contents go to a fresh 0600 temp file in the target directory, which is fsynced and renamed over the target - the same approach the Windows implementation already uses. Readers see the old file or the new file, never a partial one.

Behaviour changes on POSIX:
- A pre-existing target no longer keeps its permission bits: the result is always owner-only (0600 filtered by umask).
- A symlink at the target path is replaced by a regular file instead of being written through.
- Overwriting needs a writable directory instead of a writable file, so writeSecrets can now overwrite its own read-only output.

Also: haddocks updated for all five affected functions, and a new test pins the overwrite semantics (contents replaced, permissions re-tightened to owner-only).

# How to trust this PR

  Same temp+rename pattern as the existing `Win32.hs`; temp file is a sibling of the target, so the rename is atomic.
  Added new test that checks the permission changing behaviour works.
  Behaviour changes (always 0600, symlinks replaced) are documented in haddocks and changelog.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/.changes/20260810_120000_cardano-api_palas_atomic_owner_permission_writes.yml
+++ b/.changes/20260810_120000_cardano-api_palas_atomic_owner_permission_writes.yml
@@ -1,0 +1,11 @@
+project: cardano-api
+
+pr: 1286
+
+kind:
+  - bugfix
+
+description: |
+  On POSIX, the `WithOwnerPermissions` family of functions (`writeFileTextEnvelopeWithOwnerPermissions`, `writeByteStringFileWithOwnerPermissions`, `writeLazyByteStringFileWithOwnerPermissions`, `writeTextFileWithOwnerPermissions`) and `writeSecrets` now write atomically: the contents go to a temporary file, are synced to disk, and the temporary file is renamed over the target, as was already the case on Windows.
+  If writing fails midway (crash, full disk, power loss), the previous contents of the target file are kept instead of being destroyed.
+  Behaviour change: a pre-existing target no longer keeps its permission bits (the replacement is always owner-only, `0600` filtered by `umask`), and a symlink at the target path is replaced by a regular file instead of being written through.
--- a/.changes/20260810_120000_cardano-api_palas_atomic_owner_permission_writes.yml
+++ b/.changes/20260810_120000_cardano-api_palas_atomic_owner_permission_writes.yml
@@ -3,6 +3,7 @@ project: cardano-api
 pr: 1286
 
 kind:
+  - breaking
   - bugfix
 
 description: |
--- a/.changes/20260810_120000_cardano-api_palas_atomic_owner_permission_writes.yml
+++ b/.changes/20260810_120000_cardano-api_palas_atomic_owner_permission_writes.yml
@@ -10,3 +10,4 @@ description: |
   On POSIX, the `WithOwnerPermissions` family of functions (`writeFileTextEnvelopeWithOwnerPermissions`, `writeByteStringFileWithOwnerPermissions`, `writeLazyByteStringFileWithOwnerPermissions`, `writeTextFileWithOwnerPermissions`) and `writeSecrets` now write atomically: the contents go to a temporary file, are synced to disk, and the temporary file is renamed over the target, as was already the case on Windows.
   If writing fails midway (crash, full disk, power loss), the previous contents of the target file are kept instead of being destroyed.
   Behaviour change: a pre-existing target no longer keeps its permission bits (the replacement is always owner-only, `0600` filtered by `umask`), and a symlink at the target path is replaced by a regular file instead of being written through.
+  The affected functions (and `writeSecrets`) are also generalised from `IO` to any `MonadIO m`.
--- a/cardano-api/src/Cardano/Api/IO/Internal/Compat.hs
+++ b/cardano-api/src/Cardano/Api/IO/Internal/Compat.hs
@@ -30,7 +30,18 @@ handleFileForWritingWithOwnerPermission = handleFileForWritingWithOwnerPermissio
 -- (the contents go to a temporary file which is then renamed into place) and
 -- ends up readable only by its owner.
 writeSecrets
-  :: HasCallStack => FilePath -> [Char] -> [Char] -> (a -> ByteString) -> [a] -> IO ()
+  :: HasCallStack
+  => FilePath
+  -- ^ Output directory
+  -> [Char]
+  -- ^ Filename prefix
+  -> [Char]
+  -- ^ Filename suffix
+  -> (a -> ByteString)
+  -- ^ Serialisation function for the secrets
+  -> [a]
+  -- ^ Secrets to write, one file each
+  -> IO ()
 writeSecrets = writeSecretsImpl
 
 checkVrfFilePermissions :: File content direction -> ExceptT VRFPrivateKeyFilePermissionError IO ()
--- a/cardano-api/src/Cardano/Api/IO/Internal/Compat.hs
+++ b/cardano-api/src/Cardano/Api/IO/Internal/Compat.hs
@@ -25,6 +25,10 @@ handleFileForWritingWithOwnerPermission
   -> IO (Either (FileError e) ())
 handleFileForWritingWithOwnerPermission = handleFileForWritingWithOwnerPermissionImpl
 
+-- | Write a list of secrets to individual files in the given directory, as
+-- @\<prefix\>.\<3-digit index\>.\<suffix\>@. Each file is written atomically
+-- (the contents go to a temporary file which is then renamed into place) and
+-- ends up readable only by its owner.
 writeSecrets
   :: HasCallStack => FilePath -> [Char] -> [Char] -> (a -> ByteString) -> [a] -> IO ()
 writeSecrets = writeSecretsImpl
--- a/cardano-api/src/Cardano/Api/IO/Internal/Compat.hs
+++ b/cardano-api/src/Cardano/Api/IO/Internal/Compat.hs
@@ -15,15 +15,18 @@ import Cardano.Api.IO.Internal.Compat.Wasm
 import Cardano.Api.IO.Internal.Compat.Win32
 
 import Control.Monad.Except (ExceptT)
+import Control.Monad.IO.Class (MonadIO (..))
 import Data.ByteString (ByteString)
 import GHC.Stack (HasCallStack)
 import System.IO
 
 handleFileForWritingWithOwnerPermission
-  :: FilePath
+  :: MonadIO m
+  => FilePath
   -> (Handle -> IO ())
-  -> IO (Either (FileError e) ())
-handleFileForWritingWithOwnerPermission = handleFileForWritingWithOwnerPermissionImpl
+  -> m (Either (FileError e) ())
+handleFileForWritingWithOwnerPermission path f =
+  liftIO $ handleFileForWritingWithOwnerPermissionImpl path f
 
 -- | Write a list of secrets to individual files in the given directory, as
 -- @\<prefix\>.\<3-digit index\>.\<suffix\>@. Each file is written atomically
@@ -31,6 +34,7 @@ handleFileForWritingWithOwnerPermission = handleFileForWritingWithOwnerPermissio
 -- ends up readable only by its owner.
 writeSecrets
   :: HasCallStack
+  => MonadIO m
   => FilePath
   -- ^ Output directory
   -> [Char]
@@ -41,8 +45,9 @@ writeSecrets
   -- ^ Serialisation function for the secrets
   -> [a]
   -- ^ Secrets to write, one file each
-  -> IO ()
-writeSecrets = writeSecretsImpl
+  -> m ()
+writeSecrets outDir prefix suffix secretOp xs =
+  liftIO $ writeSecretsImpl outDir prefix suffix secretOp xs
 
 checkVrfFilePermissions :: File content direction -> ExceptT VRFPrivateKeyFilePermissionError IO ()
 checkVrfFilePermissions = checkVrfFilePermissionsImpl
--- a/cardano-api/src/Cardano/Api/IO/Internal/Compat/Posix.hs
+++ b/cardano-api/src/Cardano/Api/IO/Internal/Compat/Posix.hs
@@ -33,10 +33,11 @@ import           System.FilePath (splitFileName, (<.>), (</>))
 import qualified System.IO as IO
 import           System.IO (Handle)
 import           System.Posix.Files (fileMode, getFileStatus, groupModes, intersectFileModes,
-                   nullFileMode, otherModes, ownerReadMode, setFileMode)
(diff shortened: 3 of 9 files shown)


cardano-api/src/Cardano/Api/IO/Internal/Compat/Posix.hs
```diff
@@ -23,50 +23,53 @@ import           Cardano.Api.IO.Internal.Base
...
-    Right fd -> do
-      bracket
-        (fdToHandle fd)
-        IO.hClose
-        (runExceptT . handleIOExceptT (FileIOError path) . f)
+        (IO.openTempFile targetDir $ targetFile <.> "tmp")
+        ( \(tmpPath, h) -> do
+            IO.hClose h
+            IO.removeFile tmpPath
+        )
```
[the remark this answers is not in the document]
palas: `hClose` on an already-closed handle is specified to be a no-op, not an error

.changes/20260810_120000_cardano-api_palas_atomic_owner_permission_writes.yml
```diff
@@ -0,0 +1,11 @@
+project: cardano-api
+
+pr: 1286
+
+kind:
+  - bugfix
+
```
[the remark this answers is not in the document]
palas: Addressed in: https://github.com/IntersectMBO/cardano-api/pull/1286/changes/c197bfcf7f9fc45460f823d29c7e0aa06f7b15e8

cardano-api/src/Cardano/Api/IO/Internal/Compat/Posix.hs
```diff
@@ -23,50 +23,53 @@ import           Cardano.Api.IO.Internal.Base
...
+        ( \(tmpPath, h) -> do
+            f h
+            -- 'handleToFd' flushes the handle's buffers and closes it, handing
+            -- us the raw file descriptor. Syncing the descriptor before the
+            -- rename ensures a power failure cannot leave an empty file at the
+            -- target path.
+            bracket (handleToFd h) closeFd fileSynchronise
+            IO.renameFile tmpPath path
```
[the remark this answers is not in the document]
palas: Addressed in: https://github.com/IntersectMBO/cardano-api/pull/1286/changes/c197bfcf7f9fc45460f823d29c7e0aa06f7b15e8

cardano-api/src/Cardano/Api/IO/Internal/Compat/Posix.hs
```diff
@@ -46,27 +45,37 @@ handleFileForWritingWithOwnerPermissionImpl
...
+            -- failure cannot leave an empty file at the target path.
+            bracket
+              (handleToFd h)
+              closeFd
+              (\fd -> setFdOwnerAndGroup fd user (-1) >> fileSynchronise fd)
+            IO.renameFile tmpPath path
+        )
+  case result of
+    Left (err :: IOException) -> pure $ Left $ FileIOError path err
+    Right () -> pure $ Right ()
```
carbolymer: I'd use `fmap (first $ FileIOError path)` before `try`. This case isn't doing anything besides error handling.

cardano-api/test/cardano-api-test/Test/Cardano/Api/IO.hs
```diff
@@ -32,9 +37,54 @@ prop_createVrfFileWithOwnerPermissions =
...
+  H.propertyOnce . H.moduleWorkspace "help" $ \ws -> do
+    file <- H.noteTempFile ws "file"
+
+    -- Create the target with different contents, to check that overwriting
+    -- replaces them.
+    firstResult <- liftIO $ writeByteStringFile (File file) "old contents"
+
+    case firstResult of
+      Left err -> failWith Nothing $ docToString $ prettyError @(FileError ()) err
+      Right () -> return ()
```
carbolymer: Why not `H.leftFail`?

cardano-api/test/cardano-api-test/Test/Cardano/Api/IO.hs
```diff
@@ -32,9 +37,54 @@ prop_createVrfFileWithOwnerPermissions =
...
+      Left _ -> return ()
+      Right () ->
+        failWith Nothing "precondition failed: file is not group/other readable before the overwrite"
+#endif
+
+    result <- liftIO $ writeLazyByteStringFileWithOwnerPermissions (File file) "new contents"
+
+    case result of
+      Left err -> failWith Nothing $ docToString $ prettyError @(FileError ()) err
+      Right () -> return ()
```
carbolymer: Why not H.leftFail?

cardano-api/test/cardano-api-test/Test/Cardano/Api/IO.hs
```diff
@@ -32,9 +37,54 @@ prop_createVrfFileWithOwnerPermissions =
...
+
+    contents <- liftIO $ BS.readFile file
+    contents === "new contents"
+
+    fResult <- liftIO . runExceptT $ checkVrfFilePermissions (File file)
+
+    case fResult of
+      Left err -> failWith Nothing $ show err
+      Right () -> liftIO (removeFile file) >> success
```
carbolymer: Why the file is only being removed in the `Right` case? 

This can also be replaced by `H.leftFail`.

cardano-api/test/cardano-api-test/Test/Cardano/Api/IO.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
```
carbolymer: I'd use runtime checks like `H.isWin32` instead of `CPP` just to have fourmolu working here but this is fine here.

cardano-api/src/Cardano/Api/IO/Internal/Compat/Posix.hs
carbolymer, line 1: Suggestion: use `MonadUnliftIO` / `MonadIO` instead of `IO` for the changed functions

cardano-api/src/Cardano/Api/IO/Internal/Compat.hs
```diff
@@ -25,6 +25,10 @@ handleFileForWritingWithOwnerPermission
   -> IO (Either (FileError e) ())
 handleFileForWritingWithOwnerPermission = handleFileForWritingWithOwnerPermissionImpl
 
+-- | Write a list of secrets to individual files in the given directory, as
+-- @\<prefix\>.\<3-digit index\>.\<suffix\>@. Each file is written atomically
+-- (the contents go to a temporary file which is then renamed into place) and
+-- ends up readable only by its owner.
```
carbolymer, line 34: a haddocks for function parameters would be nice too 😉

cardano-api/test/cardano-api-test/Test/Cardano/Api/IO.hs
```diff
@@ -32,9 +37,54 @@ prop_createVrfFileWithOwnerPermissions =
...
+    case fResult of
+      Left err -> failWith Nothing $ show err
+      Right () -> liftIO (removeFile file) >> success
+
 tests :: TestTree
 tests =
   testGroup
     "Test.Cardano.Api.IO"
     [ testProperty "Create VRF File with Owner Permissions" prop_createVrfFileWithOwnerPermissions
+    , testProperty "Overwrite file with Owner Permissions" prop_overwriteFileWithOwnerPermissions
```
carbolymer, line 100: The PR body claims `writeSecrets` can now overwrite its own read-only output, but no test pins that scenario - and it's the `writeSecrets`-specific payoff of the change (the `0400` files it leaves behind used to block a second run). Something along these lines would keep it from regressing silently:

```haskell
prop_writeSecretsOverwritesItsOwnOutput :: Property
prop_writeSecretsOverwritesItsOwnOutput =
  H.propertyOnce . H.moduleWorkspace "help" $ \ws -> do
    -- first run leaves 0400 owner-read-only files
    liftIO $ writeSecrets ws "secret" "key" id ["old"]
    -- the second run must replace them regardless
    liftIO $ writeSecrets ws "secret" "key" id ["new"]
    contents <- liftIO $ BS.readFile (ws </> "secret.000.key")
    contents === "new"
```

cardano-api/src/Cardano/Api/IO/Internal/Compat/Posix.hs
```diff
@@ -46,27 +45,37 @@ handleFileForWritingWithOwnerPermissionImpl
...
+            -- 'handleToFd' flushes the handle's buffers and closes it, handing
+            -- us the raw file descriptor. Before the rename, we set the file's
+            -- ownership to the real user (which can differ from the effective
+            -- user in setuid programs) and sync the descriptor, so a power
+            -- failure cannot leave an empty file at the target path.
+            bracket
+              (handleToFd h)
+              closeFd
+              (\fd -> setFdOwnerAndGroup fd user (-1) >> fileSynchronise fd)
+            IO.renameFile tmpPath path
```
carbolymer, line 73: Optional, non-blocking: the temp file is fsynced but the directory isn't, so the replacement is atomic but not durable - after a power failure shortly after the call returns, the target may still hold the *old* contents (never a partial file, so the haddock's claim is accurate as written). If we want the rename itself to survive power loss, fsync the directory after `renameFile`:

```haskell
bracket (openFd targetDir ReadOnly defaultFileFlags) closeFd fileSynchronise
```

Two caveats if you take this: `openFd`'s arity differs before unix-2.8 (the CPP this PR just deleted was guarding exactly that), and the cabal file currently has no version bound on `unix` - so it'd need a `unix >= 2.8` bound to stay CPP-free. Fine as a follow-up; for key generation the atomicity is the part that matters, and "power loss rolls back to the previous key" is arguably even acceptable semantics.

carbolymer (approved):

Jimbo4350 (approved):

cardano-api/test/cardano-api-test/Test/Cardano/Api/IO.hs
palas: The idea is to not touch anything after the test fails, so that it is easier to debug. `H.moduleWorkspace` does the same, delete the folder if the test succeeds and keep it if it fails. Which, actually, means we don't even need to remove the file if it succeeds, but well

cardano-api/src/Cardano/Api/IO/Internal/Compat/Posix.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-api/pull/1286/commits/71d5c7e57c9211410a7cdbcf81ee130d673e9e9a

cardano-api/test/cardano-api-test/Test/Cardano/Api/IO.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-api/pull/1286/commits/71d5c7e57c9211410a7cdbcf81ee130d673e9e9a

cardano-api/test/cardano-api-test/Test/Cardano/Api/IO.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-api/pull/1286/commits/71d5c7e57c9211410a7cdbcf81ee130d673e9e9a

cardano-api/test/cardano-api-test/Test/Cardano/Api/IO.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-api/pull/1286/commits/71d5c7e57c9211410a7cdbcf81ee130d673e9e9a

cardano-api/test/cardano-api-test/Test/Cardano/Api/IO.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-api/pull/1286/commits/71d5c7e57c9211410a7cdbcf81ee130d673e9e9a

cardano-api/src/Cardano/Api/IO/Internal/Compat.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-api/pull/1286/commits/71d5c7e57c9211410a7cdbcf81ee130d673e9e9a

cardano-api/test/cardano-api-test/Test/Cardano/Api/IO.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-api/pull/1286/commits/71d5c7e57c9211410a7cdbcf81ee130d673e9e9a

cardano-api/src/Cardano/Api/IO/Internal/Compat/Posix.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-api/pull/1286/commits/cb5ee5b1bd9921d738d7a3cc36bc6c1b579a7b5b

cardano-api/src/Cardano/Api/IO/Internal/Compat/Posix.hs
palas: Addressed in: https://github.com/IntersectMBO/cardano-api/pull/1286/commits/1287b3e7fd5dbbcceb9251e87ddcb9867572fc98

PR #1288, opened 2026-08-11
Fix plutus script redeemer pointer indexing

carbolymer:
Fix plutus script redeemer pointer indexing: proposal pointers now follow the transaction's insertion order and certificate pointers count unwitnessed certificates, in both the experimental and the deprecated transaction builders. Remove the unused StakeCredential field from the WitTxCert constructor. Add property tests checking every redeemer pointer against the ledger's own resolution.

# Context

Redeemer pointers for certificates and proposals were computed against the wrong order. Proposals were sorted by their `Ord` instance before index assignment, but the ledger resolves `Proposing` purposes by the transaction's insertion order (`OSet` -> `StrictSeq` positional lookup). Unwitnessed certificates (e.g. plain stake registrations) were dropped before index assignment, but the ledger counts every certificate's position in the full sequence, so a plutus-witnessed certificate placed after an unwitnessed one received a shifted pointer and failed phase-2 validation. The certificate bug existed in both the experimental builder and the deprecated `createTransactionBody` path (reachable via `makeTransactionBodyAutoBalance`).

Also hardened `getVotes` against gaps in the voter map: it now uses a total lookup with a placeholder witness on a miss, rather than a lookup that silently skips a missing voter and shifts every later voter's redeemer index.

## How this was fixed

- Preserve insertion order in `compareWitnesses` for proposals.
- Keep unwitnessed certificates as index-slot placeholders in both builders.
- Remove the unused `StakeCredential` field from `WitTxCert` (breaking).
- Correct the redeemer-index ordering documentation on `indexTxProposalProcedures`/`indexWitnessedTxProposalProcedures`.

# How to trust this PR

Property tests assert every witnessable category's redeemer pointer against the ledger's own `Indexable`/`redeemerPointer` resolution: container-level oracles for all six categories, old-API bridge twins, and an end-to-end `createTransactionBody` property with per-certificate redeemer binding. Each fix was mutation-verified (reverting the fix makes the corresponding property fail).

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-api/src/Cardano/Api/Experimental/Plutus/Internal/IndexedPlutusScriptWitness.hs
+++ b/cardano-api/src/Cardano/Api/Experimental/Plutus/Internal/IndexedPlutusScriptWitness.hs
@@ -79,7 +79,6 @@ data Witnessable (thing :: WitnessableItem) era where
   WitTxCert
     :: (L.EraTxCert era, L.AlonzoEraScript era)
     => L.TxCert era
-    -> StakeCredential
     -> Witnessable CertItem era
   WitMint
     :: L.AlonzoEraScript era
@@ -111,11 +110,16 @@ compareWitnesses :: Witnessable thing era -> Witnessable thing era -> Ordering
 compareWitnesses a b =
   case (a, b) of
     (WitTxIn txinA, WitTxIn txinB) -> compare txinA txinB
-    (WitTxCert{}, WitTxCert{}) -> LT -- Certificates in the ledger are in an `OSet` therefore we preserve the order.
+    -- Certificates are stored in an `OSet` but resolved positionally, via
+    -- `certsTxBodyL`'s `findIndexL`. `EQ` lets the stable sort in
+    -- `createIndexedPlutusScriptWitnesses` preserve insertion order.
+    (WitTxCert{}, WitTxCert{}) -> EQ
     (WitMint polIdA _, WitMint polIdB _) -> compare polIdA polIdB
     (WitWithdrawal stakeAddrA _, WitWithdrawal stakeAddrB _) -> compare stakeAddrA stakeAddrB
     (WitVote voterA, WitVote voterB) -> compare voterA voterB
-    (WitProposal propA, WitProposal propB) -> compare propA propB
+    -- Proposals are also stored in an `OSet` and resolved positionally
+    -- (`StrictSeq.findIndexL`), same as `WitTxCert` above.
+    (WitProposal{}, WitProposal{}) -> EQ
 
 data WitnessableItem
   = TxInItem
--- a/cardano-api/src/Cardano/Api/Experimental/Tx/Internal/BodyContent/New.hs
+++ b/cardano-api/src/Cardano/Api/Experimental/Tx/Internal/BodyContent/New.hs
@@ -6,7 +6,6 @@
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
-{-# LANGUAGE TupleSections #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
@@ -666,11 +665,12 @@ newtype TxWithdrawals era = TxWithdrawals {unTxWithdrawals :: [(StakeAddress, L.
 
 newtype TxCertificates era
   = TxCertificates
-  {unTxCertificates :: OMap (Exp.Certificate era) (Maybe (StakeCredential, AnyWitness era))}
+  {unTxCertificates :: OMap (Exp.Certificate era) (Maybe (AnyWitness era))}
   deriving (Show, Eq)
 
--- | Create 'TxCertificates'. Note that 'Certificate era' will be deduplicated. Only Certificates with a
--- stake credential will be in the result.
+-- | Create 'TxCertificates'. Note that 'Certificate era' will be deduplicated. Certificates that
+-- require a witness will be stored with 'Just' the caller-supplied witness; those that do not (e.g.
+-- deposit-less stake registration in Conway) will be stored with 'Nothing'.
 --
 -- Note that, when building a transaction in Conway era, a witness is not required for staking credential
 -- registration, but this is only the case during the transitional period of Conway era and only for staking
@@ -686,10 +686,10 @@ mkTxCertificates era certs = TxCertificates . OMap.fromList $ map getStakeCred c
   getStakeCred
     :: (Exp.Certificate (LedgerEra era), AnyWitness (LedgerEra era))
     -> ( Exp.Certificate (LedgerEra era)
-       , Maybe (StakeCredential, AnyWitness (LedgerEra era))
+       , Maybe (AnyWitness (LedgerEra era))
        )
   getStakeCred (c@(Exp.Certificate cert), wit) =
-    (c, (,wit) <$> getTxCertWitness (convert era) (obtainCommonConstraints era cert))
+    (c, wit <$ getTxCertWitness (convert era) (obtainCommonConstraints era cert))
 
 newtype TxMintValue era
   = TxMintValue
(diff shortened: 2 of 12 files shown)


cardano-api/src/Cardano/Api/Experimental/Plutus/Internal/IndexedPlutusScriptWitness.hs
```diff
@@ -111,11 +110,15 @@ compareWitnesses :: Witnessable thing era -> Witnessable thing era -> Ordering
...
+    -- Certificates are stored in an `OSet` but resolved positionally, via
+    -- `certsTxBodyL`'s `findIndexL`. Preserve insertion order.
+    (WitTxCert{}, WitTxCert{}) -> LT
     (WitMint polIdA _, WitMint polIdB _) -> compare polIdA polIdB
     (WitWithdrawal stakeAddrA _, WitWithdrawal stakeAddrB _) -> compare stakeAddrA stakeAddrB
     (WitVote voterA, WitVote voterB) -> compare voterA voterB
-    (WitProposal propA, WitProposal propB) -> compare propA propB
+    -- Proposals are also stored in an `OSet` and resolved positionally
+    -- (`StrictSeq.findIndexL`), same as `WitTxCert` above.
+    (WitProposal{}, WitProposal{}) -> LT
```
Jimbo4350: Ah yes, I missed this.

Jimbo4350 (approved):

cardano-api/src/Cardano/Api/Experimental/Tx/Internal/BodyContent/New.hs
```diff
@@ -867,10 +874,8 @@ extractWitnessableCertificates
 extractWitnessableCertificates txCerts =
   obtainCommonConstraints (useEra @era) $
     List.nub
-      [ ( WitTxCert cert stakeCred
-        , wit
-        )
-      | (Exp.Certificate cert, Just (stakeCred, wit)) <- getCertificates txCerts
+      [ (WitTxCert cert, maybe AnyKeyWitnessPlaceholder snd mCredAndWit)
```
Jimbo4350: So when we transition to `Dijsktra` this is when all certificates need a witness. In `Conway` stake registration certificates don't need any witnesses.

https://github.com/IntersectMBO/cardano-ledger/blob/f8d6ead7c84e87b175efe3259fa838210be6c2d2/eras/conway/impl/src/Cardano/Ledger/Conway/TxCert.hs#L801-L806

Jimbo4350 (changes requested):

cardano-api/src/Cardano/Api/Experimental/Tx/Internal/BodyContent/New.hs
carbolymer: Thanks for checking that. To make the types more precise, without overcomplicating things I think the path forward here is to turn our current definition
```haskell
newtype TxCertificates era
  = TxCertificates
  {unTxCertificates :: OMap (Exp.Certificate era) (Maybe (StakeCredential, AnyWitness era))}
  deriving (Show, Eq)
```

into GADT:

```haskell
data TxCertificates era where
  TxCertificatesConway 
   :: OMap (Exp.Certificate ConwayEra) (Maybe (StakeCredential, AnyWitness ConwayEra))
   -> TxCertificates ConwayEra
  TxCertificatesDijkstra 
   :: OMap (Exp.Certificate DijkstraEra) (StakeCredential, AnyWitness DijkstraEra)
   -> TxCertificates DijkstraEra

```

After hardforking to Dijkstra we could drop the old era constructor and go back to newtype.


`TxCertificates era` is however used in compatible API, so this would mean duplicating the newtype definition there.

carbolymer: Another alternative would be to add a type family:
```haskell
type family CertWitness era where
  CertWitness ShelleyEra = Maybe (StakeCredential, AnyWitness ConwayEra)
  ...
  CertWitness ConwayEra = Maybe (StakeCredential, AnyWitness ConwayEra)
  CertWitness DijkstraEra = (StakeCredential, AnyWitness ConwayEra)
```
But it would pollute experimental API with older eras information, which is exactly what we want to avoid in experimental api.

Jimbo4350: I think we can solve this in a simpler way: `extractWitnessableCertificates` should return `Maybe (AnyWitness era)` and pattern match on `Era era` inside the function. 

`Nothing` means "requires no witness", which only the Conway branch can produce (deposit-less stake registration) and the Dijkstra branch always returns `Just`, falling back to `AnyKeyWitnessPlaceholder`, since from Dijkstra on every certificate is witnessed.

carbolymer: So that would mean we'd have to unwrap the Maybe at every usage site (`getTxScriptWitnessesRequirements`, `createIndexedPlutusScriptWitnesses`)?

I've changed the logic so it's doing the defaulting only for < Dijkstra. For Dijkstra a missing witness now fails loudly instead of being silently defaulted: it can only mean `mkTxCertificates` and the ledger disagree on witness requirements, so it's a bug we want surfaced. I hope this is better now.

Jimbo4350:
```diff
diff --git a/cardano-api/src/Cardano/Api/Experimental/Tx/Internal/BodyContent/New.hs b/cardano-api/src/Cardano/Api/Experimental/Tx/Internal/BodyContent/New.hs
index f6d3e7a18..7d854a51f 100644
--- a/cardano-api/src/Cardano/Api/Experimental/Tx/Internal/BodyContent/New.hs
+++ b/cardano-api/src/Cardano/Api/Experimental/Tx/Internal/BodyContent/New.hs
@@ -6,7 +6,6 @@
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
-{-# LANGUAGE TupleSections #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
@@ -666,11 +665,12 @@ newtype TxWithdrawals era = TxWithdrawals {unTxWithdrawals :: [(StakeAddress, L.
 
 newtype TxCertificates era
   = TxCertificates
-  {unTxCertificates :: OMap (Exp.Certificate era) (Maybe (StakeCredential, AnyWitness era))}
+  {unTxCertificates :: OMap (Exp.Certificate era) (Maybe (AnyWitness era))}
   deriving (Show, Eq)
 
--- | Create 'TxCertificates'. Note that 'Certificate era' will be deduplicated. Only Certificates with a
--- stake credential will be in the result.
+-- | Create 'TxCertificates'. Note that 'Certificate era' will be deduplicated. Certificates that
+-- require a witness will be stored with 'Just' the caller-supplied witness; those that do not
+-- (e.g. deposit-less stake registration in Conway) will be stored with 'Nothing'.
 --
 -- Note that, when building a transaction in Conway era, a witness is not required for staking credential
 -- registration, but this is only the case during the transitional period of Conway era and only for staking
@@ -686,10 +686,10 @@ mkTxCertificates era certs = TxCertificates . OMap.fromList $ map getStakeCred c
   getStakeCred
     :: (Exp.Certificate (LedgerEra era), AnyWitness (LedgerEra era))
     -> ( Exp.Certificate (LedgerEra era)
-       , Maybe (StakeCredential, AnyWitness (LedgerEra era))
+       , Maybe (AnyWitness (LedgerEra era))
        )
   getStakeCred (c@(Exp.Certificate cert), wit) =
-    (c, (,wit) <$> getTxCertWitness (convert era) (obtainCommonConstraints era cert))
+    (c, wit <$ getTxCertWitness (convert era) (obtainCommonConstraints era cert))
 
 newtype TxMintValue era
   = TxMintValue
@@ -874,11 +874,10 @@ extractWitnessableTxIns tIns =
 -- keeps one entry per certificate in insertion order.
 --
 -- In the Conway era only, a certificate may legitimately have no witness
--- (deposit-less stake registration), so a missing witness there defaults to
--- 'AnyKeyWitnessPlaceholder'. From Dijkstra onwards every certificate
--- requires a witness, so a missing one there indicates a bug upstream
--- ('mkTxCertificates' failing to derive it) and is reported with an error
--- rather than silently defaulted or dropped.
+-- (deposit-less stake registration), so a missing witness defaults to
+-- 'AnyKeyWitnessPlaceholder'. From Dijkstra onwards 'mkTxCertificates'
+-- guarantees every entry has a 'Just' witness, so the placeholder is dead
+-- code for those eras.
 extractWitnessableCertificates
   :: forall era
    . IsEra era
@@ -888,21 +887,9 @@ extractWitnessableCertificates (TxCertificates certs) =
   obtainCommonConstraints (useEra @era) $
     List.nub
       [ (WitTxCert cert, wit)
-      | (certificate@(Exp.Certificate cert), mCredAndWit) <- toList certs
-      , let wit :: AnyWitness (LedgerEra era)
-            wit = case useEra @era of
-              ConwayEra -> maybe AnyKeyWitnessPlaceholder snd mCredAndWit
-              DijkstraEra -> maybe (unwitnessedCertificateError certificate) snd mCredAndWit
+      | (Exp.Certificate cert, mWit) <- toList certs
+      , let wit = fromMaybe AnyKeyWitnessPlaceholder mWit
       ]
- where
-  unwitnessedCertificateError :: Exp.Certificate (LedgerEra era) -> AnyWitness (LedgerEra era)
-  unwitnessedCertificateError certificate =
-    error $
-      "extractWitnessableCertificates: certificate without a witness in the Dijkstra era, "
-        <> "but from Dijkstra on every certificate requires a witness. "
-        <> "This is a cardano-api bug (mkTxCertificates should have derived the witness via getTxCertWitness). "
-        <> "Certificate: "
-        <> show certificate
 
 extractWitnessableMints
   :: forall era
diff --git a/cardano-api/src/Cardano/Api/Experimental/Tx/Internal/Certificate/Compatible.hs b/cardano-api/src/Cardano/Api/Experimental/Tx/Internal/Certificate/Compatible.hs
index a51335964..405173351 100644
--- a/cardano-api/src/Cardano/Api/Experimental/Tx/Internal/Certificate/Compatible.hs
+++ b/cardano-api/src/Cardano/Api/Experimental/Tx/Internal/Certificate/Compatible.hs
@@ -205,5 +205,18 @@ getTxCertWitness sbe ledgerCert = shelleyBasedEraConstraints sbe $
   case Ledger.getVKeyWitnessTxCert ledgerCert of
     Just keyHash -> Just $ StakeCredentialByKey $ Api.StakeKeyHash $ Ledger.coerceKeyRole keyHash
     Nothing ->
-      StakeCredentialByScript . fromShelleyScriptHash
-        <$> Ledger.getScriptWitnessTxCert ledgerCert
+      case Ledger.getScriptWitnessTxCert ledgerCert of
+        Just scriptHash -> Just $ StakeCredentialByScript $ fromShelleyScriptHash scriptHash
+        Nothing ->
+          case sbe of
+            ShelleyBasedEraShelley -> Nothing
+            ShelleyBasedEraAllegra -> Nothing
+            ShelleyBasedEraMary -> Nothing
+            ShelleyBasedEraAlonzo -> Nothing
+            ShelleyBasedEraBabbage -> Nothing
+            ShelleyBasedEraConway -> Nothing
+            ShelleyBasedEraDijkstra ->
+              error
+                "getTxCertWitness: certificate has no witness in the Dijkstra era. \
+                \From Dijkstra onwards every certificate requires a witness. \
+                \This indicates a bug in the ledger's EraTxCert instance for this certificate type."
```

So this is the change I had in mind. We throw the error in `getTxCertWitness`.

Jimbo4350 (approved):

PR #1298, opened 2026-08-19
Complete the era eons for Dijkstra

palas:
# Context

Make `DijkstraEra` a first-class era in `cardano-api`: everywhere the API dispatches on eras, Dijkstra now takes a real branch instead of hitting a TODO stub.

This is groundwork: it adds no Dijkstra-specific features, it makes them possible to build. Dijkstra stays hidden from era enumeration (maxBound is still Conway); exposing it there comes in a later PR.

One known gap is left on purpose: simple scripts remain unsupported in Dijkstra: they need the era's new guard construct in `SimpleScript` first, which deserves its own design discussion.

# How to trust this PR

Almost all of this is interpolation of the previous eras. Two exceptions:
1. The eon constraint bundles no longer provide `ShelleyEraTxCert/TxCert ~ ConwayTxCert` (Dijkstra dropped those certificates), so a few call sites now require them explicitly. This is the breaking part.
2. Dijkstra replaced required signer hashes with guards, so the tx-body lens has no Dijkstra arm (like the ledger's) and `createTransactionBody` translates `TxExtraKeyWitnesses` into key-hash guards (`Body.hs`, `Body/Lens.hs`).

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-api/src/Cardano/Api/Era/Internal/Eon/AllegraEraOnwards.hs
+++ b/cardano-api/src/Cardano/Api/Era/Internal/Eon/AllegraEraOnwards.hs
@@ -97,7 +97,10 @@ type AllegraEraOnwardsConstraints era =
   , L.EraTxOut (ShelleyLedgerEra era)
   , L.HashAnnotated (L.TxBody L.TopTx (ShelleyLedgerEra era)) L.EraIndependentTxBody
   , L.AllegraEraTxBody (ShelleyLedgerEra era)
-  , L.ShelleyEraTxCert (ShelleyLedgerEra era)
+  , -- L.ShelleyEraTxCert dropped: gated by AtMostEra "Conway" in the ledger, so
+    -- Dijkstra cannot satisfy it. Callsites needing Shelley-style certs must
+    -- require ShelleyEraTxCert (ShelleyLedgerEra era) explicitly.
+    L.EraTxCert (ShelleyLedgerEra era)
   , FromCBOR (Consensus.ChainDepState (ConsensusProtocol era))
   , FromCBOR (DebugLedgerState era)
   , IsCardanoEra era
@@ -119,7 +122,7 @@ allegraEraOnwardsConstraints = \case
   AllegraEraOnwardsAlonzo -> id
   AllegraEraOnwardsBabbage -> id
   AllegraEraOnwardsConway -> id
-  _ -> const $ error "TODO Dijkstra: allegraEraOnwardsConstraints: era not supported"
+  AllegraEraOnwardsDijkstra -> id
 
 class IsShelleyBasedEra era => IsAllegraBasedEra era where
   allegraBasedEra :: AllegraEraOnwards era
@@ -138,3 +141,6 @@ instance IsAllegraBasedEra BabbageEra where
 
 instance IsAllegraBasedEra ConwayEra where
   allegraBasedEra = AllegraEraOnwardsConway
+
+instance IsAllegraBasedEra DijkstraEra where
+  allegraBasedEra = AllegraEraOnwardsDijkstra
--- a/cardano-api/src/Cardano/Api/Era/Internal/Eon/AllegraEraOnwards.hs
+++ b/cardano-api/src/Cardano/Api/Era/Internal/Eon/AllegraEraOnwards.hs
@@ -97,10 +97,7 @@ type AllegraEraOnwardsConstraints era =
   , L.EraTxOut (ShelleyLedgerEra era)
   , L.HashAnnotated (L.TxBody L.TopTx (ShelleyLedgerEra era)) L.EraIndependentTxBody
   , L.AllegraEraTxBody (ShelleyLedgerEra era)
-  , -- L.ShelleyEraTxCert dropped: gated by AtMostEra "Conway" in the ledger, so
-    -- Dijkstra cannot satisfy it. Callsites needing Shelley-style certs must
-    -- require ShelleyEraTxCert (ShelleyLedgerEra era) explicitly.
-    L.EraTxCert (ShelleyLedgerEra era)
+  , L.EraTxCert (ShelleyLedgerEra era)
   , FromCBOR (Consensus.ChainDepState (ConsensusProtocol era))
   , FromCBOR (DebugLedgerState era)
   , IsCardanoEra era
--- a/cardano-api/src/Cardano/Api/Era/Internal/Eon/BabbageEraOnwards.hs
+++ b/cardano-api/src/Cardano/Api/Era/Internal/Eon/BabbageEraOnwards.hs
@@ -115,7 +115,8 @@ type BabbageEraOnwardsConstraints era =
   , L.MaryEraTxBody (ShelleyLedgerEra era)
   , L.Script (ShelleyLedgerEra era) ~ L.AlonzoScript (ShelleyLedgerEra era)
   , L.ScriptsNeeded (ShelleyLedgerEra era) ~ L.AlonzoScriptsNeeded (ShelleyLedgerEra era)
-  , L.ShelleyEraTxCert (ShelleyLedgerEra era)
+  , -- L.ShelleyEraTxCert dropped: gated by AtMostEra "Conway" in the ledger.
+    L.EraTxCert (ShelleyLedgerEra era)
   , L.TxOut (ShelleyLedgerEra era) ~ L.BabbageTxOut (ShelleyLedgerEra era)
   , L.Value (ShelleyLedgerEra era) ~ L.MaryValue
   , FromCBOR (Consensus.ChainDepState (ConsensusProtocol era))
@@ -136,7 +137,7 @@ babbageEraOnwardsConstraints
 babbageEraOnwardsConstraints = \case
   BabbageEraOnwardsBabbage -> id
   BabbageEraOnwardsConway -> id
-  BabbageEraOnwardsDijkstra -> const $ error "TODO Dijkstra: babbageEraOnwardsConstraints: era not supported"
+  BabbageEraOnwardsDijkstra -> id
 
 class IsAlonzoBasedEra era => IsBabbageBasedEra era where
   babbageBasedEra :: BabbageEraOnwards era
@@ -146,3 +147,6 @@ instance IsBabbageBasedEra BabbageEra where
 
 instance IsBabbageBasedEra ConwayEra where
   babbageBasedEra = BabbageEraOnwardsConway
+
+instance IsBabbageBasedEra DijkstraEra where
+  babbageBasedEra = BabbageEraOnwardsDijkstra
--- a/cardano-api/src/Cardano/Api/Era/Internal/Eon/BabbageEraOnwards.hs
+++ b/cardano-api/src/Cardano/Api/Era/Internal/Eon/BabbageEraOnwards.hs
@@ -115,8 +115,7 @@ type BabbageEraOnwardsConstraints era =
   , L.MaryEraTxBody (ShelleyLedgerEra era)
   , L.Script (ShelleyLedgerEra era) ~ L.AlonzoScript (ShelleyLedgerEra era)
   , L.ScriptsNeeded (ShelleyLedgerEra era) ~ L.AlonzoScriptsNeeded (ShelleyLedgerEra era)
-  , -- L.ShelleyEraTxCert dropped: gated by AtMostEra "Conway" in the ledger.
-    L.EraTxCert (ShelleyLedgerEra era)
+  , L.EraTxCert (ShelleyLedgerEra era)
   , L.TxOut (ShelleyLedgerEra era) ~ L.BabbageTxOut (ShelleyLedgerEra era)
   , L.Value (ShelleyLedgerEra era) ~ L.MaryValue
   , FromCBOR (Consensus.ChainDepState (ConsensusProtocol era))
--- a/cardano-api/src/Cardano/Api/Era/Internal/Eon/MaryEraOnwards.hs
+++ b/cardano-api/src/Cardano/Api/Era/Internal/Eon/MaryEraOnwards.hs
@@ -98,8 +98,9 @@ type MaryEraOnwardsConstraints era =
   , L.EraUTxO (ShelleyLedgerEra era)
   , L.HashAnnotated (L.TxBody L.TopTx (ShelleyLedgerEra era)) L.EraIndependentTxBody
   , L.MaryEraTxBody (ShelleyLedgerEra era)
-  , L.ShelleyEraTxCert (ShelleyLedgerEra era)
-  , L.Value (ShelleyLedgerEra era) ~ L.MaryValue
+  , -- L.ShelleyEraTxCert dropped: Dijkstra cannot satisfy AtMostEra "Conway".
+    -- Callsites that need it must add it explicitly.
+    L.Value (ShelleyLedgerEra era) ~ L.MaryValue
   , FromCBOR (Consensus.ChainDepState (ConsensusProtocol era))
   , FromCBOR (DebugLedgerState era)
   , IsCardanoEra era
@@ -120,7 +121,7 @@ maryEraOnwardsConstraints = \case
   MaryEraOnwardsAlonzo -> id
   MaryEraOnwardsBabbage -> id
   MaryEraOnwardsConway -> id
-  MaryEraOnwardsDijkstra -> const $ error "TODO Dijkstra: maryEraOnwardsConstraints: era not supported"
+  MaryEraOnwardsDijkstra -> id
 
 class IsAllegraBasedEra era => IsMaryBasedEra era where
   maryBasedEra :: MaryEraOnwards era
@@ -136,3 +137,6 @@ instance IsMaryBasedEra BabbageEra where
 
 instance IsMaryBasedEra ConwayEra where
   maryBasedEra = MaryEraOnwardsConway
+
+instance IsMaryBasedEra DijkstraEra where
+  maryBasedEra = MaryEraOnwardsDijkstra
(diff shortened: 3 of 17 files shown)


cardano-api/src/Cardano/Api/Experimental/Tx/Internal/Certificate/Compatible.hs
```diff
@@ -74,7 +74,9 @@ makeStakeAddressDelegationCertificate sCred delegatee =
     ShelleyBasedEraDijkstra -> error "TODO Dijkstra: makeStakeAddressDelegationCertificate: era not supported"
```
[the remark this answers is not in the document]
palas: We'll address that later

cardano-api/src/Cardano/Api/Era/Internal/Eon/AllegraEraOnwards.hs
```diff
@@ -97,7 +97,10 @@ type AllegraEraOnwardsConstraints era =
   , L.EraTxOut (ShelleyLedgerEra era)
   , L.HashAnnotated (L.TxBody L.TopTx (ShelleyLedgerEra era)) L.EraIndependentTxBody
   , L.AllegraEraTxBody (ShelleyLedgerEra era)
-  , L.ShelleyEraTxCert (ShelleyLedgerEra era)
+  , -- L.ShelleyEraTxCert dropped: gated by AtMostEra "Conway" in the ledger, so
+    -- Dijkstra cannot satisfy it. Callsites needing Shelley-style certs must
+    -- require ShelleyEraTxCert (ShelleyLedgerEra era) explicitly.
```
carbolymer: Redundant comment. We don't have comments for other constraints.

cardano-api/src/Cardano/Api/Era/Internal/Eon/BabbageEraOnwards.hs
```diff
@@ -115,7 +115,8 @@ type BabbageEraOnwardsConstraints era =
   , L.MaryEraTxBody (ShelleyLedgerEra era)
   , L.Script (ShelleyLedgerEra era) ~ L.AlonzoScript (ShelleyLedgerEra era)
   , L.ScriptsNeeded (ShelleyLedgerEra era) ~ L.AlonzoScriptsNeeded (ShelleyLedgerEra era)
-  , L.ShelleyEraTxCert (ShelleyLedgerEra era)
+  , -- L.ShelleyEraTxCert dropped: gated by AtMostEra "Conway" in the ledger.
```
carbolymer: Redundant comment.

cardano-api/src/Cardano/Api/Era/Internal/Eon/MaryEraOnwards.hs
```diff
@@ -98,8 +98,9 @@ type MaryEraOnwardsConstraints era =
   , L.EraUTxO (ShelleyLedgerEra era)
   , L.HashAnnotated (L.TxBody L.TopTx (ShelleyLedgerEra era)) L.EraIndependentTxBody
   , L.MaryEraTxBody (ShelleyLedgerEra era)
-  , L.ShelleyEraTxCert (ShelleyLedgerEra era)
-  , L.Value (ShelleyLedgerEra era) ~ L.MaryValue
+  , -- L.ShelleyEraTxCert dropped: Dijkstra cannot satisfy AtMostEra "Conway".
+    -- Callsites that need it must add it explicitly.
```
carbolymer: Redundant comment.

cardano-api/src/Cardano/Api/Era/Internal/Eon/ShelleyBasedEra.hs
```diff
@@ -236,8 +237,11 @@ type ShelleyBasedEraConstraints era =
   , L.EraCertState (ShelleyLedgerEra era)
   , L.EraAccounts (ShelleyLedgerEra era)
   , L.EraGov (ShelleyLedgerEra era)
-  , L.ShelleyEraTxCert (ShelleyLedgerEra era)
-  , FromCBOR (Consensus.ChainDepState (ConsensusProtocol era))
+  , -- L.ShelleyEraTxCert dropped: gated by AtMostEra "Conway" in the ledger, so
+    -- Dijkstra cannot satisfy it. Callsites that construct Shelley-style certs
+    -- must require ShelleyEraTxCert (ShelleyLedgerEra era) explicitly — that
+    -- naturally excludes Dijkstra at the type level.
```
carbolymer: Redundant Comment.

cardano-api/src/Cardano/Api/Experimental/Era.hs
```diff
@@ -160,6 +160,7 @@ instance FromJSON (Some Era) where
 instance Eon Era where
   inEonForEra v f = \case
     Api.ConwayEra -> f ConwayEra
+    Api.DijkstraEra -> f DijkstraEra
```
carbolymer, line 170: `Enum (Some Era)` does not roundtrip. `Bounded (Some Era)` needs updating.

cardano-api/src/Cardano/Api/Tx/Internal/Body/Lens.hs
```diff
@@ -169,7 +171,9 @@ reqSignerHashesTxBodyL
 reqSignerHashesTxBodyL w@AlonzoEraOnwardsAlonzo = alonzoEraOnwardsConstraints w $ txBodyL . L.reqSignerHashesTxBodyL
 reqSignerHashesTxBodyL w@AlonzoEraOnwardsBabbage = alonzoEraOnwardsConstraints w $ txBodyL . L.reqSignerHashesTxBodyL
 reqSignerHashesTxBodyL w@AlonzoEraOnwardsConway = alonzoEraOnwardsConstraints w $ txBodyL . L.reqSignerHashesTxBodyL
-reqSignerHashesTxBodyL AlonzoEraOnwardsDijkstra = error "TODO Dijkstra: reqSignerHashesTxBodyL: era not supported"
+-- Dijkstra replaced required signer hashes with guards; the ledger gates its
+-- lens to @AtMostEra "Conway"@ and stubs the instance with 'L.notSupportedInThisEraL'.
+reqSignerHashesTxBodyL AlonzoEraOnwardsDijkstra = L.notSupportedInThisEraL
```
carbolymer: So this is a runtime error. `AlonzoEraOnwards` cannot be used here.

cardano-api/src/Cardano/Api/Tx/Internal/Body.hs
```diff
@@ -1329,8 +1329,18 @@ createTransactionBody sbe bc =
...
-    setReqSignerHashes <- monoidForEraInEonA era $ \w ->
-      pure $ Endo $ A.reqSignerHashesTxBodyL w .~ convExtraKeyWitnesses apiExtraKeyWitnesses
+    setReqSignerHashes <-
+      let keyWits = convExtraKeyWitnesses apiExtraKeyWitnesses
+       in monoidForEraInEonA era $ \w -> case w of
+            -- Dijkstra replaced required signer hashes with guards, and a key-hash
+            -- guard makes the ledger demand that key's signature: translate, appending
+            -- so any other guards stay intact. (The 'A.reqSignerHashesTxBodyL' arm
+            -- fails for Dijkstra, like the ledger lens.)
+            AlonzoEraOnwardsDijkstra ->
```
carbolymer, line 1341: Usage of eons is a bit pointless if you're pattern-matching on eras.
1. `_` wildcard match will capture newer than dijkstra eras - but this should result in compile error I think
2. It is still an extra maintenance burden when new era comes around.

carbolymer (changes requested):
We need a new eons ending at Conway for that, because there are breaking changes in Dijkstra. But I guess @Jimbo4350 would not like that.

cardano-api/src/Cardano/Api/Era/Internal/Eon/ShelleyBasedEra.hs
```diff
@@ -236,8 +237,11 @@ type ShelleyBasedEraConstraints era =
   , L.EraCertState (ShelleyLedgerEra era)
   , L.EraAccounts (ShelleyLedgerEra era)
   , L.EraGov (ShelleyLedgerEra era)
-  , L.ShelleyEraTxCert (ShelleyLedgerEra era)
-  , FromCBOR (Consensus.ChainDepState (ConsensusProtocol era))
+  , -- L.ShelleyEraTxCert dropped: gated by AtMostEra "Conway" in the ledger, so
```
Jimbo4350: I'm not seeing a Dijkstra `ShelleyEraTxCert` instance in cardano-ledger. Where are you getting the "gated by AtMostEra "Conway" in the ledger" from?  Have you asked the ledger about the implementation of this instance?

cardano-api/src/Cardano/Api/Tx/Internal/Body/Lens.hs
Jimbo4350: Are these lenses even used? @palas I would confirm there is a lens exposed by `cardano-ledger` and deprecate this. All of these lenses need to be deprecated.

cardano-api/src/Cardano/Api/Tx/Internal/Body.hs
Jimbo4350: This function is deprecated.

Jimbo4350 (changes requested):

cardano-api/src/Cardano/Api/Experimental/Era.hs
palas: Right, I'll fix the roundtrip (`toEnum 1 = Some DijkstraEra`). But I'll keep `maxBound` at `ConwayEra` for now.

At this point `makeUnsignedTx` still errors for Dijkstra (see [here](https://github.com/IntersectMBO/cardano-api/blob/857a8b5dcbbddaff333914d897cf109a2b1ba9a1/cardano-api/src/Cardano/Api/Experimental/Tx/Internal/BodyContent/New.hs#L190)), so advertising the era via `maxBound` would expose an era the experimental API can't build transactions for yet. I will address that in a future PR

cardano-api/src/Cardano/Api/Tx/Internal/Body/Lens.hs
palas: > Are these lenses even used? @palas I would confirm there is a lens exposed by cardano-ledger and deprecate this. All of these lenses need to be deprecated.

It is used in `createTransactionBody` only, and it is exported. The ledger exposes the same-named lens (gated to `AtMostEra "Conway"`), plus a total getter `reqSignerHashesTxBodyG` for reads. And both are already re-exported via `Cardano.Api.Ledger`. So I can just deprecate `reqSignerHashesTxBodyL`.

> So this is a runtime error. `AlonzoEraOnwards` cannot be used here.

Right, we still need to decide what to do with the behaviour in Dijkstra of the deprecated function, we basically have 2 options:
- Let it error out, and say in the deprecation message that it doesn't exist in Dijkstra.
- Create an eon like @carbolymer suggested.

We could also make it work by reading/writing key-hash guards in Dijkstra. But that would make it a bad lens, it would lose info, because guards are stored in an `OSet` and this lens returns `Set` (so we would lose ordering). And because there can also be script-hash guards.

I will just let it error out for now, unless you tell me otherwise

cardano-api/src/Cardano/Api/Tx/Internal/Body.hs
palas: Since it is deprecated, I'll just pattern match, and then we can just remove the function when the deprecation period is over

cardano-api/src/Cardano/Api/Era/Internal/Eon/ShelleyBasedEra.hs
palas: I haven't asked, but it is here: https://cardano-ledger.cardano.intersectmbo.org/cardano-ledger-shelley/Cardano-Ledger-Shelley-TxCert.html#t:ShelleyEraTxCert

Documentation also says: `All of the Shelley related certificate functionality that has been fully deprecated in Dijkstra.`

cardano-api/src/Cardano/Api/Era/Internal/Eon/AllegraEraOnwards.hs
palas: Addressed in https://github.com/IntersectMBO/cardano-api/pull/1298/changes/de8aafe6c0019547fb76138c30d913f70cfcc022

cardano-api/src/Cardano/Api/Era/Internal/Eon/MaryEraOnwards.hs
palas: Addressed in https://github.com/IntersectMBO/cardano-api/pull/1298/changes/de8aafe6c0019547fb76138c30d913f70cfcc022

cardano-api/src/Cardano/Api/Era/Internal/Eon/BabbageEraOnwards.hs
palas: Addressed in https://github.com/IntersectMBO/cardano-api/pull/1298/changes/de8aafe6c0019547fb76138c30d913f70cfcc022

cardano-api/src/Cardano/Api/Era/Internal/Eon/ShelleyBasedEra.hs
palas: Addressed in https://github.com/IntersectMBO/cardano-api/pull/1298/changes/de8aafe6c0019547fb76138c30d913f70cfcc022

cardano-api/src/Cardano/Api/Experimental/Era.hs
palas: Addressed in https://github.com/IntersectMBO/cardano-api/pull/1298/changes/de8aafe6c0019547fb76138c30d913f70cfcc022

cardano-api/src/Cardano/Api/Tx/Internal/Body/Lens.hs
palas: Addressed in https://github.com/IntersectMBO/cardano-api/pull/1298/changes/de8aafe6c0019547fb76138c30d913f70cfcc022

palas: Added deprecations here: https://github.com/IntersectMBO/cardano-api/pull/1298/changes/d6b6e520bbfa961342ee3b9d7650cfd101008b65

cardano-api/src/Cardano/Api/Tx/Internal/Body.hs
palas: Addressed in https://github.com/IntersectMBO/cardano-api/pull/1298/changes/de8aafe6c0019547fb76138c30d913f70cfcc022

cardano-api/src/Cardano/Api/Experimental/Era.hs
```diff
@@ -160,6 +161,7 @@ instance FromJSON (Some Era) where
...
     _ -> v
```
Jimbo4350: why is there a wildcard here?

cardano-api/src/Cardano/Api/Ledger/Internal/Reexport.hs
```diff
@@ -142,6 +142,22 @@ module Cardano.Api.Ledger.Internal.Reexport
   -- Babbage
   , CoinPerByte (..)
   , referenceScriptTxOutL
+  , ShelleyEraTxBody
+  , updateTxBodyL
+  , AllegraEraTxBody
```
Jimbo4350, line 151: You have totally ignored the demarcations in the export list.

Jimbo4350 (approved):
LGTM

cardano-api/src/Cardano/Api/Ledger/Internal/Reexport.hs
palas: Oops, completely missed that. I've fixed that now in: https://github.com/IntersectMBO/cardano-api/pull/1298/changes/9caeda9a05a7fd57a8c0ecc83e945e859c3eaff5

cardano-api/src/Cardano/Api/Experimental/Era.hs
palas: It was there already, and it is for eras before Conway, but it is not a good pattern, because when we add a new era it will use the before Conway and it would be better for it to not compile. So I've replaced it with per-era cases in: https://github.com/IntersectMBO/cardano-api/pull/1298/changes/9caeda9a05a7fd57a8c0ecc83e945e859c3eaff5

cardano-api/src/Cardano/Api/Experimental/Era.hs
```diff
@@ -134,6 +134,7 @@ instance Bounded (Some Era) where
 
 instance Enum (Some Era) where
   toEnum 0 = Some ConwayEra
+  toEnum 1 = Some DijkstraEra
```
carbolymer, line 137: `Bounded (Some Era)` needs updating too.

cardano-api/src/Cardano/Api/Tx/Internal/Body/Lens.hs
carbolymer: If we have to use runtime error, I think I would go a tad stricter here, and exclude Dijkstra era e.g.:

```haskell
reqSignerHashesTxBodyL
  :: (era == DijkstraEra ~ False)
  => AlonzoEraOnwards era -> Lens' (LedgerTxBody era) (Set (L.KeyHash L.Guard))
```

but I guess it's fine as it is now

carbolymer (approved):

cardano-api/src/Cardano/Api/Tx/Internal/Body/Lens.hs
palas: Ah, good idea, that avoids the eon but ensures the runtime error. I'll do that

cardano-api/src/Cardano/Api/Experimental/Era.hs
palas: Actually, I think we should hold off a bit on updating `Bounded`, both for `Some Era`, `AnyCardanoEra` and `AnyShelleyBasedEra`, because there are still some things that need to be fixed. I will do that a few PRs from now

cardano-api/src/Cardano/Api/Tx/Internal/Body/Lens.hs
palas: Addressed in https://github.com/IntersectMBO/cardano-api/pull/1298/changes/71fce3a3c8b237bda4ff152c6eea84a9c9812639

PR #1305, opened 2026-08-20
cardano-rpc: fix ReadGenesis initial funds and stop retaining the genesis in memory

### LINKED ISSUE #1314: [BUG] - Retained genesis in cardano-rpc
**Internal**

**Area**

**Summary**
`GenesisBundle` retains the whole read Genesis, which means that on benchmarks that inject +1GB of initial addresses, that is retained for the whole duration of the program. Consensus already solved this in the past by [compacting the genesis](https://github.com/IntersectMBO/ouroboros-consensus/blob/main/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Config.hs#L115-L143).

**Steps to reproduce**
Run the workbench on node 11.1 and compare memory usage with 11.0.

The heap diff:
```
┌─────────────────────────┬───────────┬────────────┬─────────────┬────────┐
│          band           │   11.0    │    11.1    │      Δ      │  ΔMiB  │
├─────────────────────────┼───────────┼────────────┼─────────────┼────────┤
│ PackedBytes28           │ 3,935,501 │ 15,200,503 │ +11,265,002 │ +429.7 │
├─────────────────────────┼───────────┼────────────┼─────────────┼────────┤
│ Ledger.Address:Addr     │ 0         │ 11,300,001 │ +11,300,001 │ +344.8 │
├─────────────────────────┼───────────┼────────────┼─────────────┼────────┤
│ GHC.Types:: (cons)      │ 40,054    │ 12,653,115 │ +12,613,061 │ +288.7 │
├─────────────────────────┼───────────┼────────────┼─────────────┼────────┤
│ GHC.Tuple:(,)           │ 883       │ 12,601,243 │ +12,600,360 │ +288.4 │
├─────────────────────────┼───────────┼────────────┼─────────────┼────────┤
│ Credential:KeyHashObj   │ 4,557,594 │ 15,200,011 │ +10,642,417 │ +162.4 │
├─────────────────────────┼───────────┼────────────┼─────────────┼────────┤
│ STACK                   │ 72        │ 3,749      │ +3,677      │ +91.3  │
├─────────────────────────┼───────────┼────────────┼─────────────┼────────┤
│ Credential:StakeRefBase │ 0         │ 1,300,000  │ +1,300,000  │ +19.8  │
├─────────────────────────┼───────────┼────────────┼─────────────┼────────┤
│ GHC.Num.Integer:IS      │ 1,365     │ 1,301,358  │ +1,299,993  │ +19.8  │
└─────────────────────────┴───────────┴────────────┴─────────────┴────────┘
```

Addr = 11,300,001 exactly, StakeRefBase = 1,300,000 exactly — the genesis initialFunds and stakeCredentials counts, to the entry. And ~12.6 M cons cells alongside ~12.6 M pairs: 11,300,001 + 1,300,000 = 12,600,001. That is the genesis funds and stake datasets retained as a list of pairs. 11.0 holds none of it (Addr 0, (,) 883).

The retainer path:
```
TSO (thread) → STACK → STRef                    src/Cardano/Node/Run.hs:473
  → NodeKernelAccess                            src/Cardano/Rpc/Server/NodeKernelAccess.hs:53
  → GenesisBundle                               NodeKernelAccess.hs:53:87-99
  → ProtocolInfoArgsCardano                     src/Cardano/Node/Protocol/Cardano.hs:156
  → CardanoProtocolParams
  → {Alonzo,Conway,Dijkstra}TransitionConfig     src/Cardano/Ledger/Api/Transition.hs:52-59
  → ShelleyGenesis                              src/Cardano/Ledger/Shelley/Genesis.hs:307-329
  → SJust                                       Genesis.hs:513
  → ShelleyExtraConfig                          Genesis.hs:214-221
  → EmbeddedInjection                           Genesis.hs:802
  → :  and  (,)                                 Data/Aeson/Types/Internal.hs:316, :287
  → Addr
```

### LINKED ISSUE #1315: [BUG] - ReadGenesis doesn't read sgExtraConfig
**Internal**

**Area**

**Summary**
ReadGenesis only looks at `sgInitialFunds` but not at `sgExtraConfig`. That itself is problematic because Ledger decided that we should transition to use `sgExtraConfig` instead. However, I do think that ReadGenesis should read none because of #1314. But if you decide to keep the initial funds, you will have to find the way to efficiently read them from the right place (and not retain them for the whole run).

carbolymer:
# Context

The UTxO RPC `ReadGenesis` response reported an empty `initial_funds` map for any network created with `cardano-cli create-testnet-data`.
`create-testnet-data` provisions wallets by writing their funded addresses into the Shelley genesis' `sgExtraConfig.secInitialFunds`, leaving the legacy `sgInitialFunds` field empty, but cardano-rpc's `ReadGenesis` handler read only `sgInitialFunds`.

Fixing that at `GenesisBundle` construction time would have entrenched #1314: the bundle already retained the whole boot-time `TransitionConfig`, keeping the full parsed genesis in memory for the node's lifetime whether or not RPC was enabled.
This PR fixes both problems together.

Fixes #1314.

- `GenesisBundle` no longer stores anything reachable from `ProtocolInfoArgs`.
  The Byron, Alonzo and Conway genesis values are shared with consensus' own ledger config, and the Shelley genesis is not kept at all, since the only copy the node has is compacted with the initial funds erased.
- The `ReadGenesis` handler reads the Shelley genesis file on demand and verifies its bytes against the hash the node computed at startup; a file that changed since startup fails the request with `FAILED_PRECONDITION`.
  Initial funds are resolved with ledger's own `resolveInjectionSource`/`foldInjectionData`, handling both `EmbeddedInjection` and `InjectionFromFile`, the latter streamed with content-hash verification.
- The resolved genesis is held in a new `TimedCache`: concurrent requests arriving at an empty cache share a single load, every read extends the value's life, and it empties itself after five minutes without a read.
  An idle or unqueried node retains no genesis data and runs no cache thread.
- cardano-api additionally exports `resolveShelleyInitialFunds`, operating on a `ShelleyGenesis`.
- Breaking: `mkNodeKernelAccess` no longer takes `ProtocolInfoArgs` and takes the Shelley genesis file path instead.

The original bug was caught by the `ReadGenesis` integration test added in cardano-node PR [#6655](https://github.com/IntersectMBO/cardano-node/pull/6655).
A stacked cardano-node PR adapting the `mkNodeKernelAccess` call site and re-enabling the `initialFunds` assertion will follow.

# How to trust this PR

- The cardano-rpc unit suite passes (`cabal test cardano-rpc-test`, 98 tests); the Genesis fixture and property tests pin the per-era proto mappings byte-for-byte.
- Consumed via a temporary `source-repository-package` stanza in cardano-node's `cabal.project` pointing at this branch's head commit, the stacked cardano-node PR will demonstrate the initial-funds fix against the real integration test.
- The retention fix has not been heap-profiled yet.
  To verify it, profile a node with a large genesis as in #1314: the `Addr`/`KeyHashObj`/cons-cell bands should return to their 11.0 levels, since nothing in `GenesisBundle` holds the parsed genesis after startup.

# Checklist

- [ ] New tests are added if needed and existing tests are updated. See [Running tests](https://github.com/input-output-hk/cardano-node-wiki/wiki/Running-tests) for more details - unit tests for `TimedCache` and the funds resolution are queued on this branch; integration coverage is the `ReadGenesis` test in the stacked cardano-node PR
- [x] Self-reviewed the diff

### DIFF
--- a/cardano-api/src/Cardano/Api/LedgerState.hs
+++ b/cardano-api/src/Cardano/Api/LedgerState.hs
@@ -77,6 +77,7 @@ module Cardano.Api.LedgerState
   , GenesisConfig (..)
   , readCardanoGenesisConfig
   , mkProtocolInfoCardano
+  , resolveShelleyInitialFunds
 
     -- *** Byron Genesis Config
   , readByronGenesisConfig
@@ -174,6 +175,7 @@ import Cardano.Ledger.Keys qualified as SL
 import Cardano.Ledger.Shelley.API qualified as ShelleyAPI
 import Cardano.Ledger.Shelley.Core qualified as Core
 import Cardano.Ledger.Shelley.Genesis qualified as Ledger
+import Cardano.Ledger.Shelley.Transition qualified as Ledger
 import Cardano.Ledger.Slot qualified as Ledger
 import Cardano.Ledger.State qualified as SL
 import Cardano.Protocol.Crypto qualified as Crypto
@@ -221,8 +223,10 @@ import Ouroboros.Network.Protocol.ChainSync.PipelineDecision
 import Control.Concurrent
 import Control.DeepSeq
 import Control.Error.Util (note)
-import Control.Exception.Safe
+import Control.Exception.Safe hiding (MonadThrow)
 import Control.Monad
+import Control.Monad.Class.MonadST (MonadST)
+import Control.Monad.Class.MonadThrow (MonadThrow)
 import Control.Monad.State.Strict
 import Control.Tracer qualified as Tracer
 import Data.Aeson as Aeson
@@ -272,7 +276,7 @@ import GHC.Stack (HasCallStack)
 import Lens.Micro
 import Network.Mux qualified as Mux
 import Network.TypedProtocol.Core (Nat (..))
-import System.FS.API (SomeHasFS)
+import System.FS.API (SomeHasFS (..))
 import System.FilePath
 
 data InitialLedgerStateError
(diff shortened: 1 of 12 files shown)


cardano-api/src/Cardano/Api/LedgerState.hs
```diff
@@ -1521,6 +1525,42 @@ readCardanoGenesisConfig enc = do
   let transCfg = Ledger.mkLatestTransitionConfig shelleyGenesis alonzoGenesis conwayGenesis dijkstraGenesis
   pure $ GenesisCardano enc byronGenesis shelleyGenesisHash transCfg
 
+-- | Resolve a Shelley genesis' 'Ledger.sgInitialFunds' against its
```
Jimbo4350, line 1528: Can you confirm if this resolution is temporary? I.e can we eventually remove  it? You also need to give come background as to why we have to deal with this.

Jimbo4350 (approved):

cardano-api/src/Cardano/Api/LedgerState.hs
carbolymer: Partially, yes. The reconciliation half exists because `ShelleyGenesis` still carries both the legacy `sgInitialFunds` and `sgExtraConfig.secInitialFunds` (which is what `create-testnet-data` writes), and the ledger reconciles them only while building the initial ledger state (`resolveInjectionSource` in `Cardano.Ledger.Shelley.Transition`), never writing the result back into the `ShelleyGenesis` value. `sgInitialFunds` isn't formally deprecated in the ledger yet, but the migration direction is the injection mechanism, and once the legacy field goes, the reconciliation goes with it. What stays necessary even then is materialising `secInitialFunds` itself, streaming the funds in and hash-verifying from-file sources, because a parsed genesis read outside the ledger's state-construction path never gets that for free. The Haddock on `resolveShelleyInitialFunds` now carries this background.

PR #1326, opened 2026-09-03
Remove the era case combinators in favour of inEonForEra

Jimbo4350:
# Context

Removes all three era case combinators and the `Cardano.Api.Era.Internal.Case` module they lived in. Call sites use `inEonForEra` / `inEonForShelleyBasedEra`, or match on the era constructors where both branches need era constraints.

- `caseShelleyEraOnlyOrAllegraEraOnwards` — internal only, never re-exported. Two of three call sites ignored the `ShelleyEraOnly` witness; the third, `invalidHereAfterTxBodyL`, needs it for `ttlAsInvalidHereAfterTxBodyL` and now matches on constructors.
- `caseByronOrShelleyBasedEra` — no call sites left, and its own comment marked it for deletion once `build-raw --byron-era` was deprecated in cardano-cli. The `Cardano.Api.Network.IPC` haddock example now shows `inEonForEra` instead.
- `caseShelleyToBabbageOrConwayEraOnwards` — 16 of 19 call sites ignored one of the two witnesses. In `toConsensusQueryShelleyBased` the eon is the experimental `Era`, which is what `obtainCommonConstraints` wants, so 11 `convert` calls go away too. The last three are certificate generators in `Test.Gen.Cardano.Api.Typed` needing `ShelleyEraTxCert` pre-Conway and `ConwayEraTxCert` after; they match on constructors.

Note that `inEonForShelleyBasedEra` hands the callback a witness but *no* constraints, unlike the case combinators' rank-2 arguments — so a few sites now call `shelleyToBabbageEraConstraints` / `obtainCommonConstraints` explicitly.

**Downstream:** cardano-cli has 5 uses of `caseShelleyToBabbageOrConwayEraOnwards` (`Compatible/Transaction/Run.hs`, `Compatible/Transaction/Option.hs`, `Compatible/Governance/Option.hs`) that will need the same treatment. All but `pRefScriptFp` discard a witness, so they map straight onto `inEonForShelleyBasedEra`.

# How to trust this PR

No behaviour change: each branch of the removed combinators maps one-to-one onto the new code, and the era partitions are unchanged.

```bash
cabal build all -j4 --enable-tests
cabal test cardano-api-test cardano-api-golden -j4
```

Locally: clean build under `-Werror`, 259/259 and 125/125 tests pass. `cardano-rpc-test` fails to build on `master` already (`Not in scope: T.show`, `Test/Cardano/Rpc/Pagination.hs:89`) — untouched here.

# Checklist

- [x] Self-reviewed the diff


cardano-api/src/Cardano/Api/Query/Internal/Type/QueryInMode.hs
```diff
@@ -758,25 +744,23 @@ toConsensusQueryShelleyBased sbe = \case
...
-      ( const $
-          error
-            "toConsensusQueryShelleyBased: GetDRepDelegations is only available from the Conway era onwards"
+    inEonForShelleyBasedEra
+      ( error
+          "toConsensusQueryShelleyBased: GetDRepDelegations is only available from the Conway era onwards"
       )
       ( \w ->
-          obtainCommonConstraints (convert w) $
+          obtainCommonConstraints w $
```
palas: Same problem as [here](https://github.com/IntersectMBO/cardano-cli/pull/1438#discussion_r3928757412) in all this file:

> Because of the use of inEonForShelleyBasedEra, when Conway goes out of Era, then it will lose the conway related commands (like drep), and we won't get any compilation error that warns about it here.
> So maybe we should do something like postConway below.

palas (approved):
All good except for the shrinking `Era` issue

PR #1332, opened 2026-09-08
(Re)move keys and credential decoding layer

palas:
# Context

In https://github.com/IntersectMBO/cardano-keys/pull/4, we move cardano key machinery to a separate repo with fewer dependencies than `cardano-api`, so that they can be used by testing projects in the consensus repos, among others, without having to depend in the full stack that `cardano-api` requires.

This PR simply removes things that were moved, and links existing code so that it uses them from `cardano-keys`.

Initially, this PR has an SRP to ensure that it works before we release `cardano-keys`. After `cardano-keys` is released we will remove that SRP and depend directly through CHaP.

# How to trust this PR

The thing to check would be that it all keeps working, and that everything was moved as it was, and nothing was deleted here that was not added in https://github.com/IntersectMBO/cardano-keys/pull/4, other than the refactorings and modifications that were necessary. CI passing would be a pretty good sign that both this PR and https://github.com/IntersectMBO/cardano-keys/pull/4 are correct.

Additionally https://github.com/IntersectMBO/cardano-cli/pull/1444 builds on this PR to ensure `cardano-cli` will build and tests pass.

You can use [this report](https://html-preview.github.io/?url=https://raw.githubusercontent.com/IntersectMBO/cardano-keys/8b2a3a8fc80ace37c6f2662453fac4dfa07de063/keys-migration-report.html) to check what was moved, copied, deleted, or written from scratch in `cardano-keys`.

# Checklist

- [x] Self-reviewed the diff


cabal.project
```diff
@@ -51,6 +51,18 @@ semaphore: True
 constraints:
   -- haskell.nix patch does not work for 1.6.8
   , any.crypton-x509-system < 1.6.8
+  -- cardano-crypto-class (in cardano-api and in the cardano-keys dependency)
+  -- requires crypton < 1.1; without this the GHC 9.14 solver greedily picks
+  -- crypton 1.1.4 and then cannot resolve cardano-crypto-class.
+  , crypton < 1.1
```
carbolymer, line 57: this should be behind GHC version check

Jimbo4350 (approved):

cabal.project
palas: Addressed here: https://github.com/IntersectMBO/cardano-api/pull/1332/changes/9c1efdd9dc4f63ac6848b6649012da68aef74740

